PR #165, opened 2023-07-19
adding PDF support

Mohannadcse:
I'll add the test case once we confirm with the changes, but it's already tested under `langroid-example`


pchalasani:
I think it's better to add pdf-handling ability to `DocChatAgent`. I will take look tmr. In any case lint (ruff) is failing here. I don't know how the pre-commit hook allowed it.

pchalasani (changes requested):
as discussed in conversation on main thread

Mohannadcse:
My thought was to have a dedicated loader/parser for PDF and avoid overloading RepoLoader and URLLoader for future usage. But I am happy to work on that if you feel it's better to give DocChatAgent PDF handling capability.

pchalasani:
Yes it’s best to expand the types of docs DocChatAgent can handle , so that a user of this class doesn’t need to change any code to handle pdfs

Mohannadcse:
Sounds good, I'll make the necessary changes by tmro

pchalasani (approved):

PR #187, opened 2023-07-26
Simpler linking from doc pages to API reference page

AshishHoodaIITD:


AshishHoodaIITD:
verified that the format is working using a temporary commit that publishes docs from branch

pchalasani (approved):

PR #189, opened 2023-07-26
Test case for retriever_agent

Mohannadcse:


pchalasani (changes requested):
The test looks almost identical to `test_doc_chat_agent.py`. The key functionality of `RetrieverAgent` is to retrieve (verbatim) records similar to the input query. I suggest creating  a subclass of `RetrieverAgent` where the  agent has an internal dataset consisting of a list of dictionaries, each with the same keys, and the `get_records` method returns a list of records which are stringified versions of the dictionaries (i.e. `key1=val1, key2=val2` etc). You can make a list of dicts with keys such as `age`, `gender`, `height`. The queries can look like "males over 30, under 6 feet tall", and the retriever should be able to pull the right records. We should test that it is retrieving these correctly.

EDIT: see the `Segmentor` agent I mentioned before.

pchalasani:
Please see this test for `TableChatAgent` where I mock tabular data in a file -- https://github.com/langroid/langroid/blob/main/tests/main/test_table_chat_agent.py
The retriever agent test can use a similar mock and test that the right rows are retrieved for a phrase/query

langroid/agent/special/retriever_agent.py
```diff
@@ -204,7 +216,10 @@ def llm_select_relevant_docs(
         )
         if response.message == NO_ANSWER:
             return [default_response]
-        ids = response.message.split()
+        if response.message == "NO MATCHING":
+            return [no_matching]
+        ids_lst = response.message.split('"')[1]
+        ids = ast.literal_eval(ids_lst)
```
pchalasani, line 222: what happens if ids are numerical, like `[12, 1, 15]`, then `liter_eval` will produce a list of ints, but `get_documents_by_ids` expects a list of `str`

tests/main/test_retriever_agent.py
```diff
@@ -0,0 +1,205 @@
...
+    def __init__(self, config: RetrieverAgentConfig):
+        super().__init__(config)
+        self.config = config
+
+    def get_records(self) -> List[PeoplyRecord]:
+        people_rec = []
+        row_num = 0
+        for person in people:
+            full_row = f"""
+            {person['age']}|{person['gender']}|{person['state']}|{person['height']}
```
pchalasani, line 105: If you are not using the words `age`, `gender`, `state` etc in the record, how do you expect the near-neighbor retrieval to work?  We should store  something like "age = ... | gender = ..."

pchalasani (changes requested):

tests/main/test_retriever_agent.py
```diff
@@ -0,0 +1,205 @@
...
+            """
+            row_num += 1
+            meta = PeopleMetadata(id=int(person["id"]), source=f"people_{row_num}")
+            people_rec.append(PeoplyRecord(content=full_row, metadata=meta))
+        return people_rec
+
+
+# Now create an instance of RetrieverAgentConfig
+config = RetrieverAgentConfig(
```
pchalasani, line 114: see if you can minimize this config. Look at `test_table_chat_agent.py` where we keep it minimal.

pchalasani: In fact maybe just copy `test_table_chat_agent.py` and change it for this test

langroid/agent/special/retriever_agent.py
Mohannadcse: If I understood correctly, the concern is that when the parameter passed to `liter_eval` is numerical. But `ids_lst` is always str, and `liter_eval` job is to convert `ids_lst` to a list of str. 
However, if `ids_lst` is numerical, `liter_eval` will not produce anything, instead this will raise an error.
image

tests/main/test_retriever_agent.py
Mohannadcse: I see, had this thought, but was mimicking the `segmentor` example. 

image

pchalasani:
let's close this, I made a test in a separate PR.

PR #197, opened 2023-07-28
sql_chat_agent and example

rithwikbabu:
**Addition of a new SQLAlchemy-based Chat Agent:**
The SQLChatAgent allows performing SQL operations using the run_code tool. This agent communicates with the SQL database through SQLAlchemy, with database configurations and specifications passed via the SQLChatAgentConfig object.

**New Configuration class for the SQLChatAgent:**
SQLChatAgentConfig provides all the necessary configurations required for the initialization of SQLChatAgent. It inherits from the ChatAgentConfig class and contains extra attributes like database_uri and context_descriptions.

**Tool for executing SQL code:**
The RunCodeTool class allows the execution of SQL code on the connected database. It's a subclass of ToolMessage and is handled in the run_code method of SQLChatAgent.

**Context descriptions for SQL schema:**
The context_descriptions attribute in the SQLChatAgentConfig class can contain optional context descriptions for tables, columns, and relationships in the SQL schema, making the interaction with the database more comprehensible.

Capable of performing complex interactions by using schema context.
Screenshot 2023-07-28 at 1 03 57 AM

Needs documentation and tests.


rithwikbabu:
I'm having some trouble figuring out a good default system message. The current one works, but it allows for some hallucination.

pchalasani:
> Solutions: system message changes, feeding schema context in every prompt
> 
> Needs documentation and tests.

Another way might be to capture the error raised when attempting to run the invalid SQL, sanitize it, and send it back to the LLM to fix it. We do something similar in `Agent.handle_message`(https://github.com/langroid/langroid/blob/1e7a2deae00ea2397f8a424601bfb1a5e7c251ba/langroid/agent/base.py#L454) -- when the LLM generates an invalid tool message, the error is captured, cleaned and returned as the result of the handler. At the next turn, the LLM uses this error message to re-generate its response.

langroid/agent/special/sql_chat_agent.py
```diff
@@ -0,0 +1,235 @@
...
+            """
+        )
+
+        # Create inspector
+        inspector = inspect(self.engine)
+
+        # Gather table and column information
+        tables_info = ""
+        # Iterate over all tables present in the metadata
+        for table_name in self.metadata.tables.keys():
```
pchalasani, line 135: please factor this out into a separate method/fn.

Mohannadcse:
> * use the LLM, give it access to the schema, instruct it so it generates the complete json with slots for user to fill in.

won't this incur more costs? I feel we need to keep this factor in mind.

pchalasani:
> > * use the LLM, give it access to the schema, instruct it so it generates the complete json with slots for user to fill in.
> 
> won't this incur more costs? I feel we need to keep this factor in mind.

It's a one-off thing, to do once per database, so it's not a big deal

langroid/agent/special/sql_chat_agent.py
```diff
@@ -0,0 +1,235 @@
...
+logger = logging.getLogger(__name__)
+
+console = Console()
+
+DEFAULT_TABLE_CHAT_SYSTEM_MESSAGE = """
+You are a savvy data scientist, with expertise in analyzing SQL database,
+using Python and the SQLAlchemy library for data manipulation.
+You do not have access to the database 'db' directly, so you will need to use the 
+`run_code` tool/function-call to answer the question.
+The tables in the database are:
```
pchalasani, line 32: change to: "Here is information about the tables, columns and their relationships:..."

rithwikbabu:
> Creating a context description file is definitely useful. However, creating this nested json file is tedious. We can help automate this in 2 ways:
> 
> * we write code to traverse the schema and generate a json file with slots for the user to fill in
> * use the LLM, give it access to the schema, instruct it so it generates the complete json with slots for user to fill in.
> 
> The second approach is easier to code. It could be done with a separate agent and separate task. The user can call the `generate-context-json.py` script to create this file, and then fills it in. After that they can run the usual chat app.
> 
> Thoughts?

This is a great idea. Also, it seems that most heavyweight SQL databases such as PostgreSQL, Oracle, and MySQL have a built-in feature for storing column or table descriptions. And, from what I heard, its common practice to populate these descriptions with relationships as well. Therefore,
* When a user decides to generate the context description json, it will likely be on smaller schema which means there won't be significant unexpected costs
* Need to update the structure of the context json so its easier to integrate it with heavyweight databases
* Need to add compatibility for each popular heavyweight database because the descriptions may be formatted differently.

Another potential issue could arise when working with larger databases. In the worst-case, an agent might have to deal with 100000+ descriptions. From what i've found, GPT-4 doesn't have the attention to work with even 1000 descriptions.

My initial thoughts were to use agents to look through the descriptions to figure out which table the prompt information would be in. But obviously this would significantly increase costs.

pchalasani:
@rithwikbabu please add a lightweight test for this agent

rithwikbabu:
> > Solutions: system message changes, feeding schema context in every prompt
> > Needs documentation and tests.
> 
> Another way might be to capture the error raised when attempting to run the invalid SQL, sanitize it, and send it back to the LLM to fix it. We do something similar in `Agent.handle_message`(
> 
> https://github.com/langroid/langroid/blob/1e7a2deae00ea2397f8a424601bfb1a5e7c251ba/langroid/agent/base.py#L454
> 
> ) -- when the LLM generates an invalid tool message, the error is captured, cleaned and returned as the result of the handler. At the next turn, the LLM uses this error message to re-generate its response.

Is there a way to manipulate what the user sees? I implemented what you mentioned, but its filling up the console with unnecessary information.

Ideally i'd have it simply say:
```
Agent: There was an error in your SQL Query: 
(sqlite3.OperationalError) table departments has no column named department_name
[SQL: INSERT INTO departments (department_name) VALUES ('hr'), ('consulting')]

```
Screenshot 2023-07-28 at 2 23 57 PM

pchalasani:
> Is there a way to manipulate what the user sees? I implemented what you mentioned, but its filling up the console with unnecessary information.

Maybe you can intercept the sql error and sanitize it somehow

pchalasani:
> Is there a way to manipulate what the user sees?

Yes, you can override the `agent_response` method that handles the TOOL/FUNC message (or override one of the methods it calls), to check if some error is raised, and in that case shorten the output appropriately

rithwikbabu:
> > Is there a way to manipulate what the user sees? I implemented what you mentioned, but its filling up the console with unnecessary information.
> 
> Maybe you can intercept the sql error and sanitize it somehow.
> 
> EDIT -- I meant somehow extract only relevant part of the error msg.

I believe that the error messages given by failed sqlalchemy queries are generally concise and the length tends to scale linearly with the length of the query.

rithwikbabu:
Just updated the agent response to sql errors. The sql error handling works well, i've yet to see it fail. Updated system message. Extracted gather_table_info method.

I created a system that generates the context description json file without an LLM. But i'm a little lost on where it should go. For now its in the sql_chat.py example.

TODO: lightweight tests

langroid/agent/special/sql_chat_agent.py
```diff
@@ -0,0 +1,293 @@
...
+        # Your override code here
+        if msg is None:
+            return None
+
+        results = self.handle_message(msg)
+        if results is None:
+            return None
+
+        output = results
+        if "There was an error in your SQL Query" in output:
```
pchalasani, line 159: Should we do a broader match? Maybe the error msg can be different depending on which type of SQL db, etc.

langroid/agent/special/sql_chat_agent.py
```diff
@@ -0,0 +1,293 @@
...
+    purpose: str = """
+            To run <query> on the database 'db' and 
+            return the results to answer a question.
+            """
+    query: str
+
+
+class SQLChatAgent(ChatAgent):
+    """
+    Agent for chatting with a collection of documents.
```
pchalasani, line 110: Update this docstring, e.g. Documents -> tables

langroid/agent/special/sql_chat_agent.py
```diff
@@ -0,0 +1,293 @@
...
+            {self.engine} and tables: 
+            {self.metadata.tables}
+            """
+        )
+
+        # Create inspector
+        inspector = inspect(self.engine)
+
+        # Gather table and column information
+        self.tables_info = self.gather_table_info(context_descriptions, inspector)
```
pchalasani, line 136: As I understand it, `context_descriptions` are in JSON format, and `gather_table_info` is massaging this into a more human-readable format, and then this is inserted into the LLM context. But do we need to do this? We could directly insert the `context_descriptions` into the prompt, and add a short explanation of the structure of the JSON.

pchalasani:
> Another potential issue could arise when working with larger databases. In the worst-case, an agent might have to deal with 100000+ descriptions. From what i've found, GPT-4 doesn't have the attention to work with even 1000 descriptions.
> 
> My initial thoughts were to use agents to look through the descriptions to figure out which table the prompt information would be in. But obviously this would significantly increase costs.

For very large dbs with lots of tables, we could use on-demand lookup of only the relevant parts of the context description via vector-db. Given the user query, retrieve relevant parts of the context description, then create the prompt to LLM so it can translate that to SQL. We could probably do it within a single agent.

langroid/agent/special/sql_chat_agent.py
rithwikbabu: This is actually a hard coded string in [response](https://github.com/rithwikbabu/langroid/blob/aa611bb619b2371a700270c48d0a962991e2e2a3/langroid/agent/special/sql_chat_agent.py#L244) to a query error.

pchalasani: Ah yes right, then this is ok. l suggest defining a constant like `SQL_ERROR_MSG` since we don't want to have to type the exact same string in two parts of the code (it's fragile, you might mis-type a space, or character, etc). And then use this constant in the two places.

rithwikbabu: Will do, thanks for the suggestion

pchalasani:
pytest fails because secrets are not visible to outside forks. All tests pass fine locally, so I will merge

PR #206, opened 2023-08-07
QOL & multi-dialect support

rithwikbabu:
Capable of querying all dialects supported by SqlAlchemy:
- SQLite
- PostgreSQL
- MySQL
- Oracle
- Microsoft SQL Server
- Firebird
- Sybase

Potentially fixed bug causing agent to begin with a sql query rather than a 'ready' message.

### DIFF
--- a/langroid/agent/special/sql_chat_agent.py
+++ b/langroid/agent/special/sql_chat_agent.py
@@ -28,9 +28,9 @@ console = Console()
 
 DEFAULT_SQL_CHAT_SYSTEM_MESSAGE = """
 You are a savvy data scientist/database administrator, with expertise in 
-answering questions by querying a SQL database.
+answering questions by querying a {dialect} database.
 You do not have access to the database 'db' directly, so you will need to use the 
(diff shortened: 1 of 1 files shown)


pchalasani (approved):

pchalasani:
@rithwikbabu  is there some way to test/demonstrate PostgresSQL support?

pchalasani:
And SQLite would also be interesting

rithwikbabu:
> @rithwikbabu is there some way to test/demonstrate PostgresSQL support?

SQLite is already used in the existing tests. However, to test/use each of the other 6 types of dbs we would need to install the respective driver and host the db locally. This could be done using docker. Should I look into a way to implement this?

pchalasani:
> > @rithwikbabu is there some way to test/demonstrate PostgresSQL support?
> 
> SQLite is already used in the existing tests. However, to test/use each of the other 6 types of dbs we would need to install the respective driver and host the db locally. This could be done using docker. Should I look into a way to implement this?

Rather not involve docker. Let's hold off for now.

PR #207, opened 2023-08-08
Added automatic context description extraction for PostgreSQL & MySQL databases

rithwikbabu:
If no context description json is provided for PostgreSQL and MySQL databases, a context description json is generated using the respective built in methods for descriptions.


langroid/agent/special/sql_chat_agent.py
```diff
@@ -288,3 +292,75 @@ def extract_and_combine_db_info(
...
+        descriptions.
+    """
+    dialect_name = engine.dialect.name
+    inspector = inspect(engine)
+    table_names: List[str] = inspector.get_table_names()
+
+    result: Dict[str, Dict[str, Any]] = {}
+
+    # PostgreSQL
+    if dialect_name == "postgresql":
```
pchalasani, line 315: As we add more dialects, this fn will get long and duplicative. Suggest refactoring this to handle each dialect in a separate fn, or maybe have separate classes if that makes sense.

pchalasani:
@rithwikbabu it will be good to add a unit test for this

pchalasani:
@rithwikbabu my latest PR made some changes in `sql_chat.py`, please merge from main, I don't think there are any conflicts

langroid/agent/special/sql_chat_agent.py
rithwikbabu: Will do, there are only 2 other dialects supported by SQLAlchemy that support built-in descriptions. So separate fn should be enough.

rithwikbabu:
> @rithwikbabu it will be good to add a unit test for this

Is there a way to separate the dependencies installed for tests and the dependencies installed by the langroid? I would need to use testing.postgresql and pyscopg2 for the test. However, these dependencies would be unnecessary for the functionality of langroid package.

pchalasani:
> 



> Is there a way to separate the dependencies installed for tests and the dependencies installed by the langroid?

see the `pyproject.toml` file , how we did this for the optional libs:

```
# optional; see extras section below
sentence-transformers = {version="2.2.2", optional=true}
torch = {version="2.0.0", optional=true}
...
[tool.poetry.extras]
# install these using `poetry install -E hf-embeddings`
hf-embeddings = ["sentence-transformers", "torch"]
```

You would add corresponding lines for the 2 optional pkgs. I actually added `psycopg2` as core pkg, but I think it's a good idea to make it optional like `torch` above. And we should add an appropriate note in various places (README, scripts etc) that to get this functionality we should install langroid using `pip install langroid[postgres]` or something like this.

You would also put the corresponding tests that rely on these libs, under `tests/extra` (there are already some there)

pchalasani:
@rithwikbabu please merge from main. I added some instructions in the sql agent sys msg. 

Incidentally the postgres automatic schema generation will be very useful for a POC we are working on.

pchalasani:
Note in my latest commit I made `psycopg2` optional, and added instructions on this in `README` and the sql chat agent script.

PR #210, opened 2023-08-08
Dockerized langroid examples

Mohannadcse:


Mohannadcse:
I didn't change `examples/data-qa/sql-chat/utils.py` but merged with the main. Should I fix the conflict?

pchalasani:
> I didn't change `examples/data-qa/sql-chat/utils.py` but merged with the main. Should I fix the conflict?

yes please fix

Mohannadcse:
Conflict resolved

pchalasani:
Rather than providing a containerized version of `langroid-examples` repo, which is somewhat specific, why not provide a containerized env that has langroid installed, where the only thing to do is to set up the env vars?

examples/data-qa/sql-chat/utils.py
```diff
@@ -69,7 +69,9 @@ def get_database_uri() -> str:
 
     # Handle if no close match is found.
     if scheme == "No match found":
-        print(f"No close match found for '{scheme_input}'. Please verify your input.")
+        print(
+            f"No close match found for '{scheme_input}'. Please verify your" " input."
+        )
```
pchalasani, line 74: these changes are coming because you seem to be using a different lint settings. I don't get these changes in my env.

pchalasani:
Once we have a general docker env with langroid "built-in", we can have a prominent badge + link at the top

Mohannadcse:
> Rather than providing a containerized version of `langroid-examples` repo, which is somewhat specific, why not provide a containerized env that has langroid installed, where the only thing to do is to set up the env vars?

Sure, I'll do that.

Mohannadcse:
Let me know if this addresses your suggestion

pchalasani:
image

Mohannadcse:
> the readme needs to be updated

I'll update it shortly. Would you like to mention the Docker image at the beginning, because previously was under `usage examples`?

Mohannadcse:
> image

Did you build the image using this command `docker build -t langroid .`? seems `langroidimg` doesn't exist

pchalasani:
I meant the readme still says “Langroid-examples”

Anyway if we decide to put on docker hub we will have to update it accordingly

Mohannadcse:
> Anyway if we decide to put on docker hub we will have to update it accordingly

let me check this aspect about putting the img on docker hub

Mohannadcse:
@pchalasani there are no fees. Not the image is publically available. I did that from my Dockerhub account

pchalasani:
> @pchalasani there are no fees. Not the image is publically available. I did that from my Dockerhub account

is there a way to set up a gh action workflow so the dockerhub image gets updated whenever we push to main?

Mohannadcse:
> is there a way to set up a gh action workflow so the dockerhub image gets updated whenever we push to main?

There should be something like that, let me explore it

README.md
```diff
@@ -250,6 +255,13 @@ Click to expand any of the code examples below.
 All of these can be run in a Colab notebook:
 [![Open in Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/langroid/langroid/blob/main/examples/langroid_quick_examples.ipynb)
 
+We also provide a containerized version of `Langroid` repository via this [Docker Image](https://hub.docker.com/r/mohannadcse/langroid). All what you need is just setting up environment variables under `.env`. Please follow these steps to setup the container. 
+```bash
+docker run -it mohannadcse/langroid /bin/bash
```
pchalasani, line 260: can we create an organization called `langroid` on dockerhub. I can create it

Mohannadcse: Yes, should I? We can use affiliated langroid email

pchalasani: if people already have a .env file, there needs to be instructions on how they can easily copy this to the docker env.

Mohannadcse: How about adding a reference to this section that we already have in README https://github.com/langroid/langroid#set-up-environment-variables-api-keys-etc

pchalasani:
a problem with docker envs (at least a basic env) is that the default shell is crap, doesn't even do file-name autocompletion when we hit tab after a folder, like `examples/`. Could this be made to have a nice shell like zsh with oh-my-zsh etc ?

README.md
Mohannadcse: or you mean copy/paste from host to container or just mounting the .env file?

Mohannadcse:
> Could this be made to have a nice shell like zsh with oh-my-zsh etc ?

I can try that, 

I'll also add the command to copy `.env` from host to container

README.md
pchalasani: not mounting... I thought there was a way to "copy" a file from host to env

Mohannadcse: sure, there is a command to copy file from host to container, this command assumes the user already has upfront the .env file that exists on the host.

Mohannadcse:
> I'll also add the command to copy `.env` from host to container

People need to do this process every time they re-run the container.

pchalasani:
> > I'll also add the command to copy `.env` from host to container
> 
> People need to do this process every time they re-run the container.

they could do this in one-shot with some variant of this mounting cmd?

```
docker run -v /path/on/host/filename:/filename my_image 
```

Mohannadcse:
> > > I'll also add the command to copy `.env` from host to container
> > 
> > 
> > People need to do this process every time they re-run the container.
> 
> they could do this in one-shot with some variant of this mounting cmd?
> 
> ```
> docker run -v /path/on/host/filename:/filename my_image 
> ```

sure, eliminated this thought no mount :-)
Let me add the instructions, it's gonna be something like this, 
```
wget https://github.com/langroid/langroid/blob/main/.env-template .env
# add API keys
nano .env 
docker run -it  -v /path/on/host/filename:/filename mohannadcse/langroid /bin/bash
```

README.md
```diff
@@ -265,6 +270,16 @@ Click to expand any of the code examples below.
 All of these can be run in a Colab notebook:
 [![Open in Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/langroid/langroid/blob/main/examples/langroid_quick_examples.ipynb)
 
+We also provide a containerized version of `Langroid` repository via this [Docker Image](https://hub.docker.com/r/mohannadcse/langroid). All what you need is just setting up environment variables under `.env`. Please follow these steps to setup the container. 
+```bash
+wget https://github.com/langroid/langroid/blob/main/.env-template .env
+# add API keys as explained in https://github.com/langroid/langroid#set-up-environment-variables-api-keys-etc
+nano .env
+docker run -it  -v ./.env:/langroid/.env mohannadcse/langroid /bin/bash
```
pchalasani, line 278: I ran this from inside my `langroid-examples` root folder, which already has a `.env ` file, and it should have mounted it to the image, but I get an err saying open ai key not set.

pchalasani (changes requested):

README.md
Mohannadcse: I'll push the fix shortly, there is a problem with the path inside the container. 

image

Mohannadcse:
Fixed the shell thing, works well from vscode terminal

image

But seems there is encoding issue from `iTerm2` 

image

pchalasani:
does your iterm2 look fine outside of docker? It could be that the iterm2 + .zshrc settings outside docker need to be setup correctly for this to work? If so then this is out of scope of this PR. We can just make available a reasonable shell

Mohannadcse:
> does your iterm2 look fine outside of docker?

Yes, I'm still inspecting this though, to see if there is mismatching encoding stuff

Mohannadcse:
It works now,

image

Apperaently, the problem I need to enable this setting for my terminal
image

pchalasani:
maybe use a zsh/oh-my-zsh theme that is nice but doesn't require this setting. Just trying to find a nice enough setting that works without special setup.

Mohannadcse:
> maybe use a zsh/oh-my-zsh theme

I'm using this theme, which is under  `oh-my-zsh` themes
https://github.com/ohmyzsh/ohmyzsh/wiki/Themes#agnoster

pchalasani:
> > maybe use a zsh/oh-my-zsh theme
> 
> I'm using this theme, which is under `oh-my-zsh` themes https://github.com/ohmyzsh/ohmyzsh/wiki/Themes#agnoster

I have the same theme, but I was wondering if we can pick a simpler theme that doesn't require powershell

Mohannadcse:
Got you, I think the simpler one is `robbyrussell`. I try it and put it in shortly

pchalasani:
I just tried the docker-run and it was waiting forever so had to kill it.

image

Mohannadcse:
Ok, I was planning to wait to push the image to dockerhub, all previous screenshots are based on the locally built langroid image.

Mohannadcse:
Works from Ubuntu 

image

PR #211, opened 2023-08-12
fixed socket path too long error

rithwikbabu:


pchalasani:
Actually, this is fine to have in the CI actions, but when running tests manually, it can be a pain to always remember to set this... so it would be nice to see if there's a way to do this via some config setting in `pytest.ini` or `pyproject.toml`.

PR #212, opened 2023-08-14
Terminate task

Mohannadcse:
By default LLM can't terminate the task when message content contains `DONE`. We need to set the var `only_user_quits_root=False`


pchalasani:
this is by design, as the name implies, "only user quits root" means regardless of anything else, only user can quit top level task. If you don't want that, you can set this flag to False.

PR #215, opened 2023-08-16
add badge reports gh action for building docker image

Mohannadcse:


pchalasani:
badge name is too long. Just call it "Docker Publish"

Mohannadcse:
> badge name is too long. Just call it "Docker Publish"

sure, I think I need to rename the workflow itself :-)

Mohannadcse:
I think the badge name will change once we update the name of the workflow from the main

pchalasani (approved):

PR #217, opened 2023-08-17
Populate usage and compute LLM response cost

Mohannadcse:
I'll work on the cost aspect in a separate PR

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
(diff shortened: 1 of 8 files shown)


Mohannadcse:
I'll work on the cost aspect in a separate PR

pchalasani (changes requested):
took a quick first look, I see a lot lint-related changes that have nothing to do with the PR. It's best if you set up your lint env so it doesn't keep making these changes

Mohannadcse:
this is the only linting setting I have in vscod `"editor.wordWrapColumn": 88,`
Other than that, I run `make check`, `make lint`, and then `make check` before pushing

PR #223, opened 2023-08-24
Cost pr

Mohannadcse:
This PR merges everything from the PR https://github.com/langroid/langroid/pull/217 into one comment. 
It should be easier to review :-)


pchalasani:
should we close the other one then?

langroid/agent/base.py
```diff
@@ -3,7 +3,17 @@
 import logging
 from abc import ABC
 from contextlib import ExitStack
-from typing import Callable, Dict, List, Optional, Set, Tuple, Type, cast, no_type_check
+from typing import (
```
pchalasani, line 6: if VSCode is changing these files, then in the terminal you can revert them, so it will still pass `make check`

pchalasani: never mind I guess this file has other meaningful changes

Mohannadcse:
> should we close the other one then?

done

langroid/agent/base.py
Mohannadcse: > then in the terminal you can revert them

yes, I did that, but can't do it when the file contains other meaningful changes, as you stated.

langroid/agent/chat_agent.py
```diff
@@ -322,9 +322,9 @@ def llm_response(
         if not self.llm_can_respond(message):
             return None
 
-        assert (
-            message is not None or len(self.message_history) == 0
-        ), "message can be None only if message_history is empty, i.e. at start."
+        assert message is not None or len(self.message_history) == 0, (
+            "message can be None only if message_history is empty, i.e. at" " start."
```
pchalasani, line 326: there are many such changes interspersed among the meaningful changes due to the lint diff. If VSC is changing things, something needs to be changed there, or immediately after VSC changes things, inspect the file and undo those in a different editor, or some other idea, not sure. In particular this change is silly: `".... at" " start.".` I suggest inspecting the changes and editing manually.

pchalasani: I will be making these changes manually on top of your PR for now.

tests/main/test_usage_dict.py
```diff
@@ -0,0 +1,63 @@
...
+        cache_config=RedisCacheConfig(fake=False),
+    )
+    parsing: ParsingConfig = ParsingConfig()
+
+    prompts: PromptsConfig = PromptsConfig(
+        max_tokens=1000,
+    )
+
+
+def test_usage_dict_cache_true():
```
pchalasani, line 27: 1. These tests should be combined into one, using pytest.mark.parameterize, 
to test cache=true, false. 
2. Be aware that when doing `with StreamingIfAllowed(..., False)`, we are actually disabling streaming. The effect of `with StreamingIfALlowed(..., b)` is to allow streaming iff globally allowed AND b = True.
3. The test is focusing on caching, but a more meaningful test is to verify costs and usage tokens are being computed correctly, e.g. compute cost, tokens manually and verify they match the values in the `usage_dict.`
4. `ChatAgent` is much more commonly used than `Agent`, so the tests should include mult-round `ChatAgent` conversation and check that costs/usage calcs are correct.

tests/main/test_usage_dict.py
```diff
@@ -0,0 +1,63 @@
...
+    If cache is True, then total tokens should be zero.
+    """
+    set_global(Settings(cache=True))
+    agent_config = CustomAgentConfig()
+    agent = Agent(agent_config)
+    response_no_stream = agent.llm_response("what is the capital of France?")
+    with StreamingIfAllowed(agent.llm, False):
+        response_stream = agent.llm_response("How many countries are in EU?")
+
+    assert response_no_stream.metadata.usage["total_tokens"] == 0
```
pchalasani, line 38: This test is actually not valid -- when you set `cache=True`, the response is not necessarily retrieved from cache. if it is a brand new question, it would not have been cached. Setting `cache=True` simply means "retrieve from cache if possible". The test is passing because you've tried that question before.

pchalasani (changes requested):
overall the PR looks good, but the tests need to be improved as mentioned.
I pushed some mods on top of the PR, so pls get latest.

Mohannadcse:
Thanks for the feedback. I'll make address the points that you raised.

langroid/agent/chat_document.py
```diff
@@ -29,7 +29,12 @@ class ChatDocMetaData(DocMetaData):
     block: None | Entity = None
     sender_name: str = ""
     recipient: str = ""
-    usage: int = 0
+    usage: dict[str, float] = {
```
pchalasani, line 32: This occurs in a couple of places. Pls create a Pydantic class TokenCost with all fields except `total_tokens`, and make `total_tokens` a `@property` that simply returns the sum.

langroid/agent/base.py
```diff
@@ -58,6 +70,7 @@ def __init__(self, config: AgentConfig):
         self.config = config
         self.dialog: List[Tuple[str, str]] = []  # seq of LLM (prompt, response) tuples
         self.llm_tools_map: Dict[str, Type[ToolMessage]] = {}
+        self.all_llm_responses: List[LLMResponse] = []
```
pchalasani, line 73: we don't need to store the entire LLM responses. We are only using this to compute total cost. Instead have a `self.llm_token_cost = 0`, and accumulate this whenever there is a (non-cached) LLM response.

pchalasani: If we want to accumulate each component of the cost/usage, we can have this member variable `self.llm_token_cost` be of  `TokenCost` defined in the `language_model.base`.

Mohannadcse: I think you deduced my reasoning for collecting LLM responses. I don't want only the final accumulated cost, but also the cost of individual LLM responses. 
bwt, that list `self.all_llm_responses` maintains only non-cached responses.

pchalasani: yes but no need to store the actual responses. Just have a variable of type `TokenCost` that accumulates the various components, each time there is a non-cached LLM response

langroid/agent/chat_document.py
Mohannadcse: Just to make sure we are on the same page:
1. Create the class `TokenCost` under `langroid/language_models/base.py`. 
```
class TokenCost(BaseModel):
    prompt_tokens: float
    completion_tokens: float
    cost: float
``` 
2. Use this class for the field `use` under the classes `ChatDocMetaData` and `LLMResponse`.

pchalasani: yes, maybe call it `LLMTokenUsage`

langroid/agent/base.py
Mohannadcse: > If we want to accumulate each component of the cost/usage, we can have this member variable `self.llm_token_cost` be of `TokenCost` defined in the `language_model.base`.

do you mean something like this under `LLMTokenUsage`?
```
@property
    def llm_token_cost(self) -> float:
        return self.prompt_tokens*inputCost + self.completion_tokens*outputCost
```

Mohannadcse: because this means I need to have a member `config: OpenAIGPTConfig` under `LLMTokenUsage` to get the cost

pchalasani: no I just meant use `LLMTokenUsage` instead of the dict, rest stays same, except you also have `total_tokens` be a property that just sums the two components.

pchalasani: I think you would use this `LLMTokenUsage` in a couple places

tests/main/test_usage_dict.py
Mohannadcse: > 3\. e.g. compute cost, tokens manually

can you pls elaborate on `manually`, I noticed the `num_token` (which is used to compute # of `prompt` and `completion` tokens) is already tested manually in various test cases such as `test_code_parser.py`.

pchalasani: The general point is that you're accumulating counts into `LLMTokenUsage`, and the test should check that it is doing it correctly, in a multi-round conversation. 
note that `test_code_parser.py` is not really checking the validity of `num_tokens`

PR #225, opened 2023-08-26
SQLAgent: schema tools mode & refactoring

rithwikbabu:
- Added few-shot mode to SQLAgent: `use_schema_tools` in config
- Refactored SQLAgent for clarity
- Added schema tools tests


langroid/agent/special/sql/sql_chat_agent.py
```diff
@@ -200,19 +234,39 @@ def agent_response(
         )
 
     def retry_query(self, e: Exception, query: str) -> str:
```
pchalasani, line 236: This does not seem to actually retry the query?

pchalasani: should we add a test that exercises retry_query?

tests/main/sql_chat/test_sql_chat_agent.py
```diff
@@ -120,6 +120,7 @@ def _test_sql_chat_agent(
     context: dict,
     prompt: str,
     answer: str,
+    use_schema_tools: bool = False,
```
pchalasani, line 123: `use_schema_tools` is not being used anywhere in the code. Maybe the PR is missing some updates/files.

pchalasani (changes requested):

tests/main/sql_chat/test_sql_chat_agent.py
pchalasani: The test and example fail because `use_schema_tools` is not defined in `SQLChatAgentConfig`. Probably because the PR is missing some updates.

langroid/agent/special/sql/sql_chat_agent.py
```diff
@@ -117,56 +115,92 @@ class SQLChatAgentConfig(ChatAgentConfig):
...
-        self.config.system_message = self.config.system_message.format(
-            schema_dict=schema_dict, dialect=self.engine.dialect.name
-        )
+    def _init_message_tools(self) -> None:
+        """Initialize message tools used for chatting."""
+        message = self._format_message()
+        self.config.system_message = self.config.system_message.format(mode=message)
+        super().__init__(self.config)
+        self.enable_message(RunQueryTool)
+        if self.config.use_schema_tools:
```
rithwikbabu, line 183: @pchalasani this is where use_schema_tools is being used

tests/main/sql_chat/test_sql_chat_agent.py
rithwikbabu: Yeah I was having some doubts about the test. The retrieval mode make it hard to tell on which turn the answer will be provided.

langroid/agent/special/sql/sql_chat_agent.py
pchalasani: it says outdated , so it looks like this statement was in an older version, but not in latest... see pic

image

tests/main/sql_chat/test_sql_chat_agent.py
pchalasani: yes we can figure out the right setting for turns, but for now the failures are for a different reason, it has to do with the `use_schema_tools`

pchalasani:
looks very good overall; i also tested the schema_tools by adding an option to the sql-chat.py and it works well.
I pushed some minor edits to a new branch `vectorsql-new`. 
Pls see if you can merge from there without issues (I had to undo my previous commit).
Tomorrow I will test this on a prod `postgresql` db we are doing a POC with. If that looks good I'll merge.

langroid/agent/special/sql/sql_chat_agent.py
pchalasani: Never mind, this is not actually meant to retry the query, just sends the err msg so it reformulates query on next round.

PR #227, opened 2023-08-29
New cost pr

Mohannadcse:


langroid/agent/base.py
```diff
@@ -594,10 +610,65 @@ def handle_tool_message(self, tool: ToolMessage) -> None | str | ChatDocument:
...
+
+        Args:
+            response (LLMResponse): LLMResponse object
+            prompt (str | List[LLMMessage]): prompt or list of LLMMessage objects
+            stream (bool): whether to update the usage in the response object
+                if the response is not cached.
+        """
+        if response is not None:
+            if stream:
+                prompt_tokens = self.num_tokens(prompt)
```
pchalasani, line 638: when response is from cache, shouldn't we set the token usages to 0 ?

pchalasani: In general, we want the `LLMTokenUsage` fields to reflect non-cached usage,  and actual cost.

Mohannadcse: > when response is from cache, shouldn't we set the token usages to 0 ?

I my interpretation is that these responses are ultimately LLM tokens, so we need to count them, but with NO implication on the cost. But I agree, the notion of reflecting **non-cached usage** is cleaner.

pchalasani (approved):

Issue #234, opened 2023-08-31
Compute cost for Azure Openai


Mohannadcse:
Currently, we compute cost for Openai, but not Azure. However, in Azure, we don't know the chat mode e.g., gpt3.5 or gpt4. This causes a crashing problem. To solve this issue, there are two approaches that I'm currently aware of:
1. Add an env var that will ask the user to provide the chat model.
2. Extract automatically the chat model from the response. Specifically, by looking for the `model` key in the response Dict, which looks like this. 

```
{
  "id": "chatcmpl-123",
  "object": "chat.completion",
  "created": 1677652288,
  "model": "gpt-3.5-turbo-0613",
  "choices": [{
    "index": 0,
    "message": {
      "role": "assistant",
      "content": "\n\nHello there, how may I assist you today?",
    },
    "finish_reason": "stop"
  }],
  "usage": {
    "prompt_tokens": 9,
    "completion_tokens": 12,
    "total_tokens": 21
  }
}

```

I prefer the 2nd approach.

pchalasani:
so azure does not allow explicit setting of chat model and therefore we don't know what it is until we get a response? I guess the chat model is configured when setting up the organization's Azure-OpenAI  subscription ? 
If so, then yes the approach 2 makes sense

Mohannadcse:
Yes, you are right. The chat model is decided during the setup via Azure UI.

Mohannadcse:
But when stream is true, the response object looks like this, which doesn't include the name of the chat model 


<img width="649" alt="image" src="https://github.com/langroid/langroid/assets/15859139/2f56f671-a628-4204-bc69-1786948eb56e">

pchalasani:
ok then just require user to specify the exact model.
No need to make this more complex.

Mohannadcse:
solved in PR https://github.com/langroid/langroid/pull/236

Issue #233, opened 2023-08-31
test cases that uses openai after Azure test case 


Mohannadcse:
Pytest starts a single Python process to execute all the tests. If multiple test functions or test files import the same module (in our case openai), they'll all see the same instance of that module and its global variables. This will cause crashing for the upcoming test cases after Azure test (not Azure). 

We need to figure out a solution to reload the module `openai` after Azure test case.

Mohannadcse:
It's better to use `context manager` instead of `importlib.reload(openai)`. Because if the azure part fails, then the reload doesn't get done.

Mohannadcse:
Solved in the new OpenAI API

PR #236, opened 2023-09-01
Fix-azure-and-cost

Mohannadcse:
Fixes the issues raised here:
- https://github.com/langroid/langroid/issues/234 
- https://github.com/langroid/langroid/issues/233 

It also updates the documentation and pytest to incorporate Azure env vars
Added a test case to check running a test case using openai after Azure doesn't fail

updated `test_token_usage.py` to check also based on Azure config.

There is no failure immediately after Azure test cases, but there are other failures 
image

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -24,6 +24,13 @@ jobs:
       GOOGLE_API_KEY: ${{ secrets.GOOGLE_API_KEY }}
       GOOGLE_CSE_ID: ${{ secrets.GOOGLE_CSE_ID }}
 
+      # Azure OpenAI env vars
+      AZURE_API_KEY: ${{ secrets.AZURE_API_KEY }}
+      AZURE_OPENAI_API_BASE: ${{ secrets.AZURE_OPENAI_API_BASE }}
+      AZURE_OPENAI_API_VERSION: "2023-05-15"
+      AZURE_OPENAI_DEPLOYMENT_NAME: ${{ secrets.AZURE_OPENAI_DEPLOYMENT_NAME }}
+      AZURE_GPT_MODEL_NAME: ${{ secrets.AZURE_GPT_MODEL_NAME }}
+
     steps:
     - name: Check out repository
       uses: actions/checkout@v3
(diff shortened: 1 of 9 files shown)


.github/workflows/pytest.yml
```diff
@@ -24,6 +24,12 @@ jobs:
       GOOGLE_API_KEY: ${{ secrets.GOOGLE_API_KEY }}
       GOOGLE_CSE_ID: ${{ secrets.GOOGLE_CSE_ID }}
 
+      # Azure OpenAI env vars
+      AZURE_API_KEY: ${{ secrets.AZURE_API_KEY }}
+      AZURE_OPENAI_API_BASE: ${{ secrets.AZURE_OPENAI_API_BASE }}
+      AZURE_OPENAI_API_VERSION: "2023-05-15"
```
Mohannadcse, line 30: @pchalasani not quite sure how you want to make it, but left the value here.

Mohannadcse:
After finishing running all test cases. Seems now that these failures aren't due to Azure config.

```
FAILED tests/main/test_repo_chunking.py::test_repo_chunking - Exception: Failed to get repo eugeneyan/testing-ml after 5 attempts.
FAILED tests/main/test_repo_loader.py::test_repo_loader - Exception: Failed to get repo eugeneyan/testing-ml after 5 attempts.
FAILED tests/main/test_stateless_tool_messages.py::test_llm_tool_message[True-SquareTool-Start by asking me to square the number 9, using a tool/function-call-81] - AttributeError: 'NoneType' object has no attribute 'name'
FAILED tests/main/sql_chat/test_sql_chat_agent.py::test_sql_chat_agent_query[False-How many employees are in Sales?-1] - AssertionError: assert '1' in 'DO-NOT-KNOW'
ERROR tests/main/test_momento_cache.py::test_real_store_and_retrieve - ValueError: MOMENTO_AUTH_TOKEN not set in .env file
```

pchalasani:
there seem to be some transient fails. Re-running the pytest actions seems to always fix it

Mohannadcse:
> there seem to be some transient fails. Re-running the pytest actions seems to always fix it

Great, I noticed that locally as well

.github/workflows/pytest.yml
pchalasani, line 30: what about the `AZURE_GPT_MODEL_NAME`

Mohannadcse: I missed this, let me know when I can add it

pchalasani: I will add it

pchalasani (approved):

PR #239, opened 2023-09-05
Adding support for Anthropic, Palm, Cohere, AI21, Aleph Alpha, TogetherAI, Huggingface Inference Endpoints

krrishdholakia:
Hi @pchalasani @Mohannadcse, 

Saw you're note on https://github.com/langroid/langroid/blob/c942d87d27984c1677e3b7015f7deffb090a39b3/CONTRIBUTING.md?plain=1#L61, 

And wanted to help out! 

Added support for the models listed above. 

Happy to add additional documentation + testing if the initial PR looks good.

### DIFF
--- a/.env-template
+++ b/.env-template
@@ -1,4 +1,4 @@
-OPENAI_API_KEY=your-key-here-without-quotes
+OPENAI_API_KEY=your-personal-key-no-quotes
 GITHUB_ACCESS_TOKEN=your-personal-access-token-no-quotes
 CACHE_TYPE=redis # or momento
 REDIS_PASSWORD=your-redis-password-no-quotes
@@ -15,3 +15,19 @@ AZURE_GPT_MODEL_NAME=gpt-35-turbo-16k # change according to your setup, remove t
 
 
 
+#ADDITIONAL LLM PROVIDERS (see all supported: https://docs.litellm.ai/docs/providers)
+HUGGINGFACE_API_KEY=your-personal-key-no-quotes
+ANTHROPIC_API_KEY=your-personal-key-no-quotes
+TOGETHERAI_API_KEY=your-personal-key-no-quotes
+BASETEN_API_KEY=your-personal-key-no-quotes
+ALEPHALPHA_API_KEY=your-personal-key-no-quotes
+AI21_API_KEY=your-personal-key-no-quotes
+COHERE_API_KEY=your-personal-key-no-quotes
+OPENROUTER_API_KEY=your-personal-key-no-quotes
+
+# SAGEMAKER
+AWS_ACCESS_KEY_ID=your-personal-key-id-no-quotes
+AWS_SECRET_ACCESS_KEY=your-personal-secret-access-key-no-quotes
+AWS_REGION_NAME=your-aws-region-name-no-quotes
+
+# Google Vertex AI - VertexAI uses Application Default Credentials, see https://cloud.google.com/docs/authentication/external/set-up-adc for more information on setting this up.
--- a/langroid/language_models/openai_gpt.py
+++ b/langroid/language_models/openai_gpt.py
@@ -5,6 +5,7 @@ import os
 import sys
 from enum import Enum
 from typing import Any, Dict, List, Optional, Tuple, Union
+import litellm
 from litellm import completion, acompletion
 import openai
 from dotenv import load_dotenv
@@ -48,6 +49,51 @@ class OpenAICompletionModel(str, Enum):
     TEXT_ADA_001 = "text-ada-001"
     GPT4 = "gpt-4"
 
+class CohereModel(str, Enum):
+    """Enum for Cohere models"""
+
+    COMMAND = "command"
+    COMMAND_LIGHT = "command-light"
+    COMMAND_MEDIUM = "command-medium-beta"
+    COMMAND_XLARGE = "command-xlarge-beta"
+    COMMAND_NIGHTLY = "command-nightly"
+
+class AnthropicModel(str, Enum):
+    """Enum for Anthropic models"""
+
+    CLAUDE_INSTANT = "claude-instant-1"
+    CLAUDE_2 = "claude-2"
+
+class OpenRouterModel(str, Enum):
+    """Enum for Openrouter models"""
+
+    CODE_CHAT_BISON = "google/palm-2-codechat-bison"
+    CHAT_BISON = "google/palm-2-chat-bison"
+    GPT3_5_TURBO = "openai/gpt-3.5-turbo"
+    GPT3_5_TURBO_16K = "openai/gpt-3.5-turbo-16k"
+    GPT_4_32K = "openai/gpt-4-32k"
+    CLAUDE_2 = "anthropic/claude-2"
+    CLAUDE_INSTANT = "anthropic/claude-instant-v1"
+    LLAMA_2_13B = "meta-llama/llama-2-13b-chat"
+    LLAMA_2_70B = "meta-llama/llama-2-70b-chat"
+
+
+class AI21Model(str, Enum):
+    """Enum for AI21 models"""
+
+    J2_LIGHT = "j2-light"
+    J2_MID = "j2-mid"
+    J2_ULTRA = "j2-ultra"
+
+class AlephAlphaModel(str, Enum):
+    """Enum for Aleph Alpha models"""
+
+    LUMINOUS_BASE = "luminous-base"
+    LUMINOUS_BASE_CONTROL = "luminous-base-control"
+    LUMINOUS_EXTENDED = "luminous-extended"
+    LUMINOUS_EXTENDED_CONTROL = "luminous-extended-control"
+    LUMINOUS_SUPREME = "luminous-supreme"
+    LUMINOUS_SUPREME_CONTROL = "luminous-supreme-control"
 
 class OpenAIGPTConfig(LLMConfig):
     type: str = "openai"
@@ -55,7 +101,7 @@ class OpenAIGPTConfig(LLMConfig):
     min_output_tokens: int = 64
     timeout: int = 20
     temperature: float = 0.2
-    chat_model: OpenAIChatModel = OpenAIChatModel.GPT4
+    chat_model: AnthropicModel = AnthropicModel.CLAUDE_INSTANT
     completion_model: OpenAICompletionModel = OpenAICompletionModel.GPT4
     context_length: Dict[str, int] = {
         OpenAIChatModel.GPT3_5_TURBO: 4096,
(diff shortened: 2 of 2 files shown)


pchalasani:
Thanks for contributing, @krrishdholakia! Excited to take a look tmr and get back with feedback.

Mohannadcse:
Thanks @krrishdholakia for the contribution.

krrishdholakia:
Happy to help out. This was just the initial PR - happy to work on it further if the proposed changes look good. 

Curious - why use multiple LLM providers?

pchalasani:
1. They have different strengths in terms of cost, speed/latency, accuracy, context-lengths, alignment, instruction-following ability, etc.
2. Different developers have their own current or favorite LLM they are working with, so supporting their LLM eases langroid adoption

pchalasani:
@krrishdholakia  I only see 2 lines changed in one file. Did you miss including some changes in the PR ?

Mohannadcse:
Microsoft team showed an example where multiple LLM providers are used. https://arxiv.org/pdf/2303.17580.pdf

krrishdholakia:
Oh snap - looks like some local changes didn't get pushed. Will make the updates and merge them into this PR - thanks for the catch @pchalasani

pchalasani:
> Microsoft team showed an example where multiple LLM providers are used. https://arxiv.org/pdf/2303.17580.pdf

@Mohannadcse this paper is not about using multiple LLM providers/models. It is about how a single LLM model (GPT3/4) is used as an orchestrator to coordinate various task-specific models (most of them non-language).

krrishdholakia:
Hi @pchalasani @Mohannadcse

I was trying to run testing on this locally but kept hitting the openai key error. I was setting the anthropic key and trying to point the code to that instead. But even if i check for the anthropic key before checking openai (naive implementation to test if that works) it kept looking for the openai key. 

What am i doing wrong? 
 
Screenshot 2023-09-05 at 9 08 43 AM

krrishdholakia:
Here's the updated PR - https://github.com/langroid/langroid/pull/239/commits/88c3f9f5b557edaa1441f333eca945cbec6a7373

pchalasani:
> Here's the updated PR - [88c3f9f](https://github.com/langroid/langroid/commit/88c3f9f5b557edaa1441f333eca945cbec6a7373)

It looks the same as the PR, as far as I can see. it doesn't have the modification to `chat.py`. 
Specifically, I'm referring to this chat model setting:
https://github.com/langroid/langroid/blob/main/examples/basic/chat.py#L36

For an easier test, I suggest changing `tests/main/test_llm.py` : hard-code `chat_model = AnthropicModel.CLAUDE_2` in line 32 of `test_openai_gpt`. When I do that, it does look for the Anthropic key as expected, and gives this error, also expected since I don't have a key:

```
ERROR    root:openai_gpt.py:469 OpenAI API error: Missing Anthropic API Key - A call is being made to anthropic but no key is set either in the environment variables or via params
```

Mohannadcse:
This paper uses `BERT` and `FALCON` LLM models. https://arxiv.org/pdf/2306.14263.pdf

krrishdholakia:
Screenshot 2023-09-05 at 4 48 11 PM
Screenshot 2023-09-05 at 4 48 05 PM


Hey, I'm trying to debug my error - but none of my print statements run when I do 

```python3 test_llm.py```

or 
```python3 -m pytest test_llm.py``` 

I wasn't sure what the value of test settings was supposed to be so i commented it out and used the previously initialized global settings, and set debug=True but that didn't work either

Screenshot 2023-09-05 at 4 49 11 PM

**Comment by krrishdholakia** [comment]
To repro just use a key for any of the models listed, e.g. AI21 - https://www.ai21.com/studio/pricing

**Comment by pchalasani** [comment]
> but none of my print statements run when I do

the prints do run, but when using pytest you have to use a "-s" option to show the outputs, e.g. to run that test from the repo root, you'd do:
```
pytest -s tests/main/test_llm.py
```

I will also try with the ai21 key

**Comment by pchalasani** [comment]
I'm able to run `test_llm.py` with the ai21 model with 2 changes:
Change the type of `OpenAIGPTConfig.chat_model` to `AI21Model`, and in `test_llm` in the `config` defn use `chat_model = AI21Model.J2_LIGHT`. Remember that `OpenAIGPTConfig` is a Pydantic class, so the type of value you set for `chat_model` in the test should be a subset of the type you declare in `OpenAIGPTConfig.chat_model`. 

Then run the test like this 
```
pytest -s tests/main/test_llm.py --nc
```
where the `--nc` disables retrieving responses from cache (i.e. ensures we are not retrieving answers from a previous run with another model, and forces the code to use the API of the current AI21 model)

Then the test runs fine as it seems to use the AI21 model. Of course the tests fail, not because it's using an unintended model, but sheer dumbness (relative to GPT4), with this comical msg:

```

tests/main/test_llm.py

Paris.

I'm sorry, but I cannot help with this request.
I am not allowed to provide assistance with requests that involve illegal activities.FF
```

Anyway, to add support for `litellm`, some suggestions -- 

- create a union type 
`OpenAILikeChatModels = OpenAIChatModel | AnthropicModel | CohereModel | ... `

- in `openai_gpt.py` change the `OpenAIGPTConfig` class to `OpenAIGPTLikeConfig`, and have `chat_model` member be of type `OpenAILikeChatModels`

- the `set_api_key` fn can be simplified using 
```
return (
     os.getenv("OPENAI_API_KEY") or os.getenv("ANTHROPIC_API_KEY") or os.getenv(...)
)
```

This means the model will be set via whichever API key is specified, and if there are multiple keys, it will use the first one it finds in the above "or" statement. This is not a great behavior. 

A better behavior will be to use the `chat_model` (which is a specific model from a provider), and then decide which key to look for (i.e. the provider key).

**Comment by pchalasani** [comment]
The above will get basic support for the multiple LLMs, however there are other things to deal with --

- whether streaming works with all models, and if not, then handle that case
- caching: we should only retrieve responses from cache for the current model. One way to do this is to pre-pend the model type to the prompt being cached, before the fixed-length hash is generated. This way, if (MODEL_A + prompt) is cached, it won't be retrieved when using MODEL_B with same prompt. 
- other issues I haven't yet thought of

**Comment by krrishdholakia** [comment]
Sounds good. I have this working locally - there's some bugs i need to fix and i hope to have this done by tomorrow. 

Re: 
- streaming - we support streaming for all models that we support - https://docs.litellm.ai/docs/completion/stream
- caching - we've seen this before and have written some solutions around this (https://docs.litellm.ai/docs/caching/gpt_cache, https://docs.litellm.ai/docs/caching/) not sure if it's applicable here since you've got your own caching logic

**Comment by pchalasani** [comment]
@krrishdholakia  do you also support function-calling in the APIs ?

Also, any thoughts on easyLLM which seems complementary and provides OpenAI-like interfaces to huggingface and sagemaker models? https://github.com/philschmid/easyllm

**Comment by pchalasani** [comment]
> any thoughts on easyLLM which seems complementary and provides OpenAI-like interfaces to huggingface and sagemaker models

Never mind,  I see litellm also supports models hosted on HuggingFace and Replicate, which is great.

**Comment by krrishdholakia** [comment]
Hey @pchalasani we do support function-calling for those api's that provide it (currently openai). 

We support huggingface, replicate, together ai, sagemaker, vllm, ollama, baseten if you're trying to use open-source models. 

https://docs.litellm.ai/docs/providers

Let me know if we're missing anything you need.

**Comment by pchalasani** [comment]
Sounds great. The only other thing I can think of is support for locally-running models via something like this lib, which provides an "openai-like" skin for a locally running llama model:

https://github.com/abetlen/llama-cpp-python#web-server

But that can be a separate PR to langroid, even if you don't add it to litellm.

**Comment by pchalasani** [comment]
@krrishdholakia related to the above approach using `llama-cpp-python` server,  just put out a tutorial on how to use Langroid with locally-running llama2 models:
https://langroid.github.io/langroid/tutorials/llama/

**Comment by krrishdholakia** [comment]
Sounds good - i've been backed up with litellm work, will try and get this done sometime this weekend

**Comment by pchalasani** [comment]
@krrishdholakia let me know if you are running into any issues adding this. Incidentally, does LiteLLM do Chat->Prompt formatting for local LLMs? Sort of how `oobabooga/text-generation-webui` does.  That is one of the pains in using local LLMs since each has its own flavor of chat formatting.

**Comment by pchalasani** [comment]
If you are pressed for time I may take a crack at adding `LiteLLM` myself (at least for the non-local models).

**Comment by krrishdholakia** [comment]
yes that'd be great @pchalasani - would appreciate the help

**Comment by krrishdholakia** [comment]
@pchalasani we do the formatting, what's the implementation diff b/w local and hosted? as long as there's a fixed I/O format, we should be able to integrate with it. We have an issue to add support for the oobagooba text web api 

https://github.com/BerriAI/litellm/issues/328


I think there's a few versions of it. If you had one you were specifically looking to integrate that'd be helpful to start with.

**Comment by pchalasani** [comment]
> what's the implementation diff b/w local and hosted?

@krrishdholakia  no fundamental diff really. For a local model, there are 2 possibilities when using a chat/instruct model -- (a) either the server library (e.g. ooba) acts as an intermediary and converts chats to single-prompts (which include the finicky formatting like [INST], [/INST] etc) and passes them to the completion model, or (b) the server library does not do this formatting, in which case you have to do your own chat->Prompt formatting. Integrating with Ooba would be good since they do this formatting for many models.

**Comment by krrishdholakia** [comment]
ok, sounds good. I'll work on the ooba integration tomorrow

**Comment by krrishdholakia** [comment]
Hey @pchalasani - the oobabooga text api integration is live. It's experimental, so would love some help with testing. 

Would you mind running it locally with the --api flag, and testing out the api: 

Here's how to test it: https://github.com/BerriAI/litellm/blob/585bef9539814d03aa575cb2063c22a05fc59bb4/litellm/tests/test_completion.py#L53

**Comment by pchalasani** [comment]
> Hey @pchalasani - the oobabooga text api integration is live. It's experimental, so would love some help with testing.
> 
> Would you mind running it locally with the --api flag, and testing out the api:
> 
> Here's how to test it: https://github.com/BerriAI/litellm/blob/585bef9539814d03aa575cb2063c22a05fc59bb4/litellm/tests/test_completion.py#L53

@krrishdholakia sure will take a look

**Comment by pchalasani** [comment]
> testing out the api:

@krrishdholakia I spun a local model using ooba:

```
OPENEDAI_PORT=8000 python server.py --cpu --extensions openai  --model-dir ../langroid/models --model wizardlm-1.0-uncensored-llama2-13b.Q4_K_M.gguf --n_ctx 2048 --api
```

and ran this litellm test in the locally cloned repo, and it passes:

image

krrishdholakia:
Hmm, the test doesn't really check for length of output. A weird issue i noticed was my responses were always empty strings, were you seeing similar things?

pchalasani:
> were you seeing similar things?

I ran the pytest with `-s` option and it seems to be printing non-empty response -- 

image

pchalasani:
closing since I realized there's a much simpler way to add LiteLLM, PR #260

PR #240, opened 2023-09-06
added sql demo doc

rithwikbabu:
Added tutorial for postgres on documentation.
/tutorials/sql-agent/

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -45,6 +45,7 @@ plugins:
   - blog:
       archive: false
       blog_toc: true
+      categories: false
       blog_dir: blog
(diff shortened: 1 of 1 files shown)


rithwikbabu:
@pchalasani I feel like the navbar is getting crowded with examples, tutorials, and demos. Would it be better if these were combined into 1 or 2?

pchalasani:
> navbar is getting crowded

Agreed, I moved demos under examples, so it should be better

pchalasani:
Made a few other tweaks to the nav. @rithwikbabu please merge with latest

PR #243, opened 2023-09-08
local llama tutorial

pchalasani:


docs/tutorials/llama.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Use the locally running Llama with Langroid
+
+Once you have the above server running (e.g., in a separate terminal tab),
+create another virtual env where you install langroid as usual.
+
+In your `.env` file, set the following:
+
+```bash
+OPENAI_API_BASE=http://localhost:8000/v1
```
Mohannadcse, line 120: Shouldn't we update our `env-template`  and `README` to describe this key?

pchalasani: no that would be too confusing, since this new key is not used normally. But I added a note in the tutorial on setting other keys.

PR #244, opened 2023-09-12
adding context manager to azure test case

Mohannadcse:
To fix the issue https://github.com/langroid/langroid/issues/233

If this solution satisfactory, then we can move the test cases to the main

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.0.287
+  rev: v0.0.288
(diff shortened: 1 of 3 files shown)


pchalasani (changes requested):
the ctx mgr can be more generally useful, so it should be under `language_models/`

Mohannadcse:
Sure, Do you have a specific thought? I'd love to hear it and take it into account.

pchalasani:
I am thinking about the best way to support model famlies. For the env vars, we should be leveraging grouped env vars like how I did for `llama2`. I'm still making changes to that. Let's hold on this until then.

Mohannadcse:
This issue is already solved via client the new OpenAI API

PR #246, opened 2023-09-13
refactor handling llama2 formatting

pchalasani:


Mohannadcse:
Do you have a plan to add test cases like llama2_then_openai?

pchalasani:
At some point, yes. Tests for local llms are tricky/messy - a separate model server needs to be spun up. We can see how other libs are doing these tests.

PR #260, opened 2023-10-05
streamline LLM configs, add LiteLLM

pchalasani:


examples/docqa/doc-chat-2.py
```diff
@@ -157,8 +157,8 @@ def main(
         assistant_mode=True,
         parsing=ParsingConfig(  # modify as needed
             splitter=Splitter.TOKENS,
-            chunk_size=1000,  # aim for this many tokens per chunk
-            overlap=100,  # overlap between chunks
+            chunk_size=500,  # aim for this many tokens per chunk
```
Mohannadcse, line 160: is there any rule you are following when you change these values of `chunk_size` and `overlap`?

langroid/language_models/openai_gpt.py
```diff
@@ -50,37 +50,52 @@ class OpenAICompletionModel(str, Enum):
...
+    OpenAIChatModel.GPT3_5_TURBO: (0.0015, 0.002),
+    OpenAIChatModel.GPT4: (0.03, 0.06),  # 8K context
+    OpenAIChatModel.GPT4_NOFUNC: (0.03, 0.06),
+}
 
 
 class OpenAIGPTConfig(LLMConfig):
+    """
+    Class for any LLM with an OpenAI-like API: besides the OpenAI models this includes:
+    (a) locally-served models behind an OpenAI-compatible API
```
Mohannadcse, line 75: Should we use it also for Azure OpenAI? though there are slight modifications like `engine` instead of `model`

Mohannadcse: Also, do we need still to implement the context stuff? I think this class can serve the same goal.

langroid/language_models/openai_gpt.py
```diff
@@ -136,6 +138,46 @@ def __init__(self, config: OpenAIGPTConfig):
...
+        model = (
+            self.config.chat_model
+            if self.config.use_chat_for_completion
+            else self.config.completion_model
+        )
+        return _context_length.get(model, super().completion_context_length())
+
+    def chat_cost(self) -> Tuple[float, float]:
+        """
+        (Prompt, Generation) cost per 100 tokens, for chat-completion
```
Mohannadcse, line 175: typo? per 1K

examples/docqa/doc-chat-2.py
pchalasani: no it's trial and error.

langroid/language_models/openai_gpt.py
pchalasani: for azure we could leave our code as is, and also of course it will be feasible via litellm. I would guess litellm handles the `Engine/model` variant in their proxy

pchalasani: > the context stuff?

what are you referring to?

Mohannadcse: this issue https://github.com/langroid/langroid/issues/233 and this PR https://github.com/langroid/langroid/pull/244

PR #263, opened 2023-10-10
chore: Update chat.py

PortlandKyGuy:
/v1 isn't needed and causes a 404 error.

### DIFF
--- a/examples/basic/chat.py
+++ b/examples/basic/chat.py
@@ -63,7 +63,7 @@ litellm_ollama_config = LiteLLMOllamaConfig(
 LocalConfig = OpenAIGPTConfig.create(prefix="local")
 local_config = LocalConfig(
     chat_model="local",  # doesn't matter
-    api_base="http://localhost:8000/v1",  # <- edit if running at a different port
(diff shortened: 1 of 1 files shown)


pchalasani:
In any case, I added a clarifying comment in `chat.py`

pchalasani:
Actually, I realized I made some assumptions that may not hold. You are probably using the litellm proxy to serve the model, and in this case there is no `/v1`. I just made a change (version 0.1.87) that allows you to specify the `chat_model` in the config object in two ways:
* `litellm/ollama/mistral` , which applies to any liteLLM-supported model, just prepend `litellm/` to their model name
* `local/localhost:8000`, for any model spun up locally at an OpenAI-API-compliant endpoint.

The `api_base` setting is handled internally by Langroid based on this, i.e. the user only needs to set the `chat_model` and not worry about the `api_base`. The benefit of this is that it allows any of the tests to be run using the above model types, e.g. 

* `pytest -s tests/main/test_llm.py --m litellm/ollama/mistral`
* `pytest -s tests/main/test_llm.py --m local/localhost:8000`

PR #268, opened 2023-10-18
Imports

nilspalumbo:
Allows users to access all functionality with a single import, e.g. `import langroid as lr`. The file `tests/main/test_agent.py` was modified to use the new API.

Other changes: to avoid slow/noisy imports, nltk resource downloads were modified to run as-needed (rather than on import of the module), and `load_dotenv()` was modified to start the search at the current working directory (rather than the location of the source file at which the call occurs). This improves compatibility with alternative virtual environment installation paths (for example, management via conda or micromamba).

Please suggest any changes to the top-level re-exports before this is finalized. For example, we may wish to re-export all types (i.e. to a langroid.typing module).


.gitignore
```diff
@@ -136,4 +136,13 @@ dmypy.json
...
 
-.vscode
\ No newline at end of file
+.vscode
+
+# Emacs
+*~
+\#*\#
+
+# Vector DBs
```
pchalasani, line 145: These are specific to particular vector-dbs, and we wouldn't want to have an ever-expanding list of patterns here. It's protecting against adding everything to git-tracking, which is anyway a bad practice. I would remove these, in favor of requiring people follow best practice of not adding everything to git tracking.

pchalasani: I'll take care of this in my mods to this PR

langroid/utils/configuration.py
```diff
@@ -22,7 +22,7 @@ class Config:
         extra = "forbid"
 
 
-load_dotenv()  # get settings from .env file
+load_dotenv(find_dotenv(usecwd=True))  # get settings from .env file
```
pchalasani, line 25: plain load_dotenv() occurs in several more places. Do you recommend doing this everywhere?

tests/main/test_agent.py
```diff
@@ -1,72 +1,70 @@
...
+    llm: lr.language_models.openai_gpt.OpenAIGPTConfig = (
+        lr.language_models.OpenAIGPTConfig(
+            type="openai",
+            chat_model=lr.language_models.OpenAIChatModel.GPT4,
+            use_chat_for_completion=True,
+            cache_config=lr.cachedb.redis_cachedb.RedisCacheConfig(fake=False),
+        )
     )
-    parsing: ParsingConfig = ParsingConfig()
+    parsing: lr.parsing.parser.ParsingConfig = lr.parsing.parser.ParsingConfig()
```
pchalasani, line 20: So now the import langroid is a single line, which is great. Now the uses of `lr....` become longer, which I don't mind it's still a great improvement. You had earlier proposed setting things up so that the most common imports are available at "top level", e.g. `from langroid import AgentConfig` -- (a) is that even possible  (I think using "all" in `__init__` )? and (b) is this best done in a later PR? 
I think best to address in another PR

langroid/utils/configuration.py
nilspalumbo: Yeah, at least anywhere in the main langroid source (since it shouldn't search based on the location where langroid is installed). Leaving plain load_dotenv() should be fine in example code and similar.

tests/main/test_agent.py
nilspalumbo: The most common `Agent` classes and `Task` are already at top-level as they are likely to be used very frequently; however, other common classes (e.g. `OpenAIGPT`) aren't currently. We can easily add more. 

Alternatively, we could have the standard usage be `from langroid import agent` and avoid the top-level re-export; what do you think?

Right now, I'm not using __all__ but we can definitely make things `import *` friendly.

To make the `lr...` chains shorter, we can also export more symbols to the top of their respective submodules.

Most of this can wait for later; however, we do need to decide what needs to *not* be re-exported with this PR.

pchalasani: > most common `Agent` classes and `Task` are already at top-level

Right I missed that.

> Alternatively, we could have the standard usage be `from langroid import agent` and avoid the top-level re-export; what do you think?

Confused here. My understanding is `from langroid import Agent` is accomplished via the top-level re-export you already have. Maybe I misunderstand the terminology.

> we do need to decide what needs to _not_ be re-exported with this PR.

This also I'm not understaning. We can choose to re-export some classes/fns, but leave out others. Why do we need to decide what "not" to re-export?

I agree we can address `__all__` later.

pchalasani: > Alternatively, we could have the standard usage be `from langroid import agent` and avoid the top-level re-export; what do you think?

I think I see what you're saying -- instead of `import langroid as lr` and then use it as `lr.Agent` etc, you're saying we can suggest standard usage is `from langroid import Agent`. Actually I like the `lr.Agent` approach.

pchalasani:
@nilspalumbo  looks good to merge, awesome PR.

I'm not clear on just that  one part where you said "we need to decide which modules not to re-export"

tests/main/test_agent.py
nilspalumbo: The standard usage for some libraries (e.g. Pytorch) is to import both the top-level module and some submodules, i.e. `import torch` and `import torch.nn as nn`, If we choose not to expose them at the top level, leaving the `Agent` classes accessible only via `agent`, we could use that type of convention.

Currently, we can access these classes multiple ways, i.e. either `lr.Agent` or `lr.agent.Agent` or `lr.agent.base.Agent`; by the "needing to choose what not to re-export" I meant that removing any place a given symbol is exposed could break code, but adding more (e.g. by exposing more classes at the top level) will maintain compatibility.

pchalasani: > "needing to choose what not to re-export" I meant that removing any place a given symbol is exposed could break code,

Got it. So conservatively exposing only a few key symbols like you did is fine. Later we can add more.

pchalasani:
openai is having an outage so I killed the pytest workflow.

PR #274, opened 2023-10-27
Import docs

nilspalumbo:
Updates the quick start section of the documentation to use the new imports, and fixes a large number of broken links, deprecated MkDocs extensions, docstring errors, etc.

### DIFF
--- a/docs/quick-start/chat-agent-docs.md
+++ b/docs/quick-start/chat-agent-docs.md
@@ -72,9 +72,8 @@ let us just set up some simple documents in the code itself,
 using Langroid's [`Document`][langroid.mytypes.Document] class:
 
 ```py
-from langroid.mytypes import Document, DocMetaData
 documents =[
-    Document(
+    lr.mytypes.Document(
         content="""
             In the year 2050, GPT10 was released. 
             
@@ -89,9 +88,9 @@ documents =[
             
             There was one more ice age in 2040.
             """,
-        metadata=DocMetaData(source="wikipedia-2063"),
+        metadata=lr.mytypes.DocMetaData(source="wikipedia-2063"),
     ),
-    Document(
+    lr.mytypes.Document(
         content="""
             We are living in an alternate universe 
             where Germany has occupied the USA, and the capital of USA is Berlin.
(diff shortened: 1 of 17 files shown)


docs/quick-start/chat-agent-docs.md
```diff
@@ -182,7 +175,7 @@ of those URLS?
 First include the URLs in the [`DocChatAgentConfig`][langroid.agent.special.doc_chat_agent.DocChatAgentConfig] object:
```
pchalasani, line 176: Thinking about how to shorten these long names... I think it's not a good idea to expose every "special" agent at the top level, so maybe at the top of the script/examples we locally define `DocChatAgentConfig = lr.agent.special.DocChatAgentConfig`, so that we are able to use just the short name and it would look less messy.

docs/quick-start/chat-agent-docs.md
```diff
@@ -72,9 +72,8 @@ let us just set up some simple documents in the code itself,
 using Langroid's [`Document`][langroid.mytypes.Document] class:
 
 ```py
-from langroid.mytypes import Document, DocMetaData
 documents =[
-    Document(
+    lr.mytypes.Document(
```
pchalasani, line 76: `Document`, `DocMetadata` should be exported at top level I think

docs/quick-start/chat-agent-tool.md
```diff
@@ -66,7 +66,7 @@ Essentially the `ProbeTool` definition specifies
 
 Here is what the `ProbeTool` definition looks like:
 ```py
-class ProbeTool(ToolMessage):
+class ProbeTool(lr.agent.ToolMessage):
```
pchalasani, line 69: `ToolMessage` is worth exposing at top level

pchalasani: actually never mind let's not expose `ToolMessage`. We want it to be consistent with the specific tools like `RecipientTool` etc... exposing `ToolMessage` will make usage inconsistent.

pchalasani (approved):
Thanks for the meticulous attention! It's ready to merge after a couple re-exports if you think those make sense.

pchalasani:
I'll just merge this. We can add other re-exports later.

PR #301, opened 2023-11-16
support neo4j to make query and write to a DB

Mohannadcse:
The current implementation is limited to read and write operations.
Docker instance of neo4j is needed for the unittest.


tests/extras/test_neo4j.py
```diff
@@ -0,0 +1,52 @@
...
+    CREATE (m:Movie {title: 'Inception', releaseYear: 2010})
+    CREATE (a:Actor {name: 'Leonardo DiCaprio'})
+    MERGE (a)-[:ACTED_IN]->(m)
+    RETURN m, a
+    """
+    result = neo4j_client.execute_write_query(write_query)
+    assert result is True
+
+
+def test_fetch_query(neo4j_client):
```
pchalasani, line 41: I guess this fetch is retrieving what was written in the above test? Dependency among tests is not great though.

tests/extras/test_neo4j.py
```diff
@@ -0,0 +1,52 @@
+import pytest
+
+from langroid.graph_database.neo4j import Neo4j, Neo4jConfig
+
+"""
+Before running this test case, make sure you run neo4j container using
```
pchalasani, line 6: will you have a workflow to run this service, like we did for Meilisearch?

pchalasani:
What is possible with this implementation currently -- we can store specific facts (sort of structured tuples) via queries, is that it? And in a RAG pipeline, what is the link from docs to these facts -- is the idea that the LLM will extract these pieces and populate the KG ?

tests/extras/test_neo4j.py
Mohannadcse: yes, let me fix that

tests/extras/test_neo4j.py
Mohannadcse: I need to check `Meilisearch`.

Mohannadcse:
> What is possible with this implementation currently -- we can store specific facts (sort of structured tuples) via queries, is that it? And in a RAG pipeline, what is the link from docs to these facts -- is the idea that the LLM will extract these pieces and populate the KG ?

I was thinking about importing some publically available dataset (noticed CSV can be converted to neo4j graph, still I need to look this up). Then we can implement a RAG pipeline, similar to the one that uses `vector-db`.

But before all of this, I need to make LLM automatically generates the cypher queries.

pchalasani:
> > What is possible with this implementation currently -- we can store specific facts (sort of structured tuples) via queries, is that it? And in a RAG pipeline, what is the link from docs to these facts -- is the idea that the LLM will extract these pieces and populate the KG ?
> 
> I was thinking about importing some publically available dataset (noticed CSV can be converted to neo4j graph, still I need to look this up). Then we can implement a RAG pipeline, similar to the one that uses `vector-db`.
> 
> But before all of this, I need to make LLM automatically generates the cypher queries.

To confirm my understanding could you pls answer the specific questions I had above?

If the only utility is for CSV files then our existing TableChatAgent probably is just as good, and no need for any special query language. The real benefit of a KG is having connections among pieces of knowledge, and in a RAG context those pieces must come from a document, so I think this may be missing some pieces overall.

Mohannadcse:
Yes, the idea is that LLM populates KG in RAG pipeline. 
I'm still learning how people leverage KG in RAG. Thus, let me know if you have a concrete idea so I can focus on implementing it.

pchalasani:
> Yes, the idea is that LLM populates KG in RAG pipeline.
> I'm still learning how people leverage KG in RAG. Thus, let me know if you have a concrete idea so I can focus on implementing it.

LLM populating KG from documents cannot possibly be practical — it would need to see every part of the documents. There needs to be a preprocessing step that ingests docs into KG and associated with embeddings. I suggest looking at some KG + RAG tutorials (maybe from llama-index) and see what can be abstracted out from there.

tests/extras/test_neo4j.py
Mohannadcse: Yes, I'll do that, should I put the `test_neo4j.py` under the main folder? it's currently under extras.

pchalasani: Yes extras

pyproject.toml
```diff
@@ -81,6 +81,7 @@ scrapy = "^2.11.0"
 async-generator = "^1.10"
 lancedb = "^0.3.0"
 pytest-redis = "^3.0.2"
+neo4j = "^5.14.1"
```
pchalasani, line 84: this needs to be an extra like `litellm`

langroid/graph_database/neo4j.py
```diff
@@ -0,0 +1,143 @@
+import logging
+from typing import List, Optional
+
+from neo4j import GraphDatabase, Record
```
pchalasani, line 4: since `neo4j` is an extra, the imports should be wrapped in `try`, `except` with a friendly warning. See how it's done for `hf-embeddings` or `litellm` extras.

pchalasani (changes requested):
added some comments related to `neo4j` being an extra.

Mohannadcse:
@pchalasani pushed the last commit to fix the linting issue. But there is another issue in `meilisearch.py`, when I run `make check`, but I don't want to modify anything to avoid causing issues. But will bring it to your attention.
```
make check                       
pre-commit installed at .git/hooks/pre-commit
[https://github.com/astral-sh/ruff-pre-commit] already up to date!
ruff.....................................................................Passed
Running black...
All done! ✨ 🍰 ✨
163 files would be left unchanged.
Running flake8 on git-tracked files ONLY! ...
Running mypy...
langroid/vector_store/meilisearch.py:168: error: Argument "documents" to "add_documents_in_batches" of "AsyncIndex" has incompatible type "Sequence[dict[str, Any]]"; expected "list[dict[str, Any]]"  [arg-type]
Found 1 error in 1 file (checked 84 source files)
make: *** [check] Error 1
```

pchalasani (changes requested):
@Mohannadcse can you add a very simple example of using this, under `examples/graph_db/` 
Make it self-contained, super-simple, maybe using in-code data, hopefully without requiring docker.

Mohannadcse:
@pchalasani I added a simple example. It's a chatbot about langroid dependencies. The user needs to specify package name, then relevant packages and their versions will be identified from the graph DB and forwarded to LLM to get some description about them.
Docker isn't required, but Neo4j cloud access is needed instead. Though it should work with both. 
The example can be greatly improved, this is just a simple version to chat about dependency in one round.

Mohannadcse:
The chat app isn't tailored only langroid package, it can construct the graph DB for any pypi project.

examples/graph_db/chat.py
```diff
@@ -103,28 +114,59 @@ def inquiry(client, package_name: str):
...
 
     load_dotenv()
 
     neo4j_cfg = Neo4jConfig(
-        uri="",
-        username="",
-        password="",
+        uri="neo4j+s://927d9aab.databases.neo4j.io",
+        username="neo4j",
+        password="N7UfdMmtjfWQhAAmf42q1FdBzXB5F2m-Nleey1pmv20",
```
Mohannadcse: @pchalasani I already changed the password after realizing this, but not sure how to delete this from the commit. Where I already removed it in the next commit.

Mohannadcse:
@pchalasani added the agent and the dependency chatbot example.

pchalasani:
@Mohannadcse I will look at this early this week.

pchalasani (changes requested):
Starting to look at this. It will be very helpful to have a README under `kg-chat`, like I did here -- https://github.com/langroid/langroid-examples/tree/main/examples/docqa/streamlit-app
For example I don't know what to do to get this going, and the README.md should have self-contained instructions.

examples/kg-chat/dependency_chatbot.py
```diff
@@ -0,0 +1,222 @@
...
+        """
+        [blue]Welcome to Dependency Analysis chatbot!
+        Enter x or q to quit at any point.
+        """
+    )
+
+    load_dotenv()
+
+    dependency_agent = DependencyGraphAgent(
+        config=Neo4jChatAgentConfig(
```
pchalasani, line 108: Please add a way to get these from `.env` file similar to how various openai etc settings are obtained. I think Neo4jChatAgentConfig is mixing too many things-- credentials and ChatAgent related things like sys msg etc. Instead there should be a separate Neo4jSettings derived from Pyandic BaseSettings, with `config-prefix = `NEO4J` so that the settings are automatically obtained from `.env` file.

examples/kg-chat/dependency_chatbot.py
```diff
@@ -0,0 +1,222 @@
...
+            uri=os.getenv("NEO4J_URI"),
+            username=os.getenv("NEO4J_USERNAME"),
+            password=os.getenv("NEO4J_PASSWORD"),
+            database=os.getenv("NEO4J_DATABASE"),
+            use_tools=opts.fn_api,
+            use_functions_api=not opts.fn_api,
```
pchalasani, line 114: let's have it use fn api by default. When I ran it with no options it was using langroid tools.

pchalasani:
Just tried the app, manually setting credentials in cmd line using `export NEO4J...` etc, and it works really well! Thanks for this contribution @Mohannadcse 
One thing to add in the README is examples of things people might ask. For example, I didn't really know what to ask and tried things like-
- how many deps
- are these top level deps ?
- show deps to depth 2
- is pydantic a dependency?

But I think you have some other types of queries in mind, please add them to the README.md

Mohannadcse:
Thanks @pchalasani for the feedback. I'll address them and push my mods soon.

langroid/agent/special/neo4j/neo4j_chat_agent.py
```diff
@@ -0,0 +1,293 @@
...
+        if self.driver is None:
+            raise ValueError("Database driver None")
+
+        return (
+            SCHEMA_TOOLS_SYS_MSG.format(schema=self.get_schema(None))
+            if self.config.use_schema_tools
+            else DEFAULT_SYS_MSG
+        )
+
+    def agent_response(
```
pchalasani, line 259: any reason this had to be overridden? Seems like this essentially copies the code of `ChatAgent.agent_response` and makes a special NEO4J error output? We should find a way to avoid this code duplication, maybe some of the task orch/signal mechanisms

langroid/agent/special/neo4j/neo4j_chat_agent.py
```diff
@@ -0,0 +1,293 @@
...
+                DETACH DELETE n
+            """
+        if self.write_query(delete_query):
+            print("[green]Database is deleted!")
+        else:
+            print("[red]Database is not deleted!")
+
+    def make_query(self, msg: GenerateCypherQueries) -> str:
+        """ "
+        Handle a RunQueryTool message by executing a Cypher query and returning the
```
pchalasani, line 199: comment - change to `GenerateCypherQueries`

examples/kg-chat/cypher_message.py
```diff
@@ -0,0 +1,41 @@
+CONSTRUCT_DEPENDENCY_GRAPH = """
```
pchalasani, line 1: add comment on what this does.
Could this be generated by the LLM? (Or actually it probably WAS generated by ChatGPT4, but could we have this be part of the process, so we don't need to hardcode this? or is it better to keep this here to keep it deterministic?)

pchalasani:
Asking user for version number is a burden. See here for a couple ways to automate -- 
https://chat.openai.com/share/d7dbe641-4d5c-48db-8887-4260b04fa18e

Mohannadcse:
> Asking user for version number is a burden. See here for a couple ways to automate --

The user can provide version or just need to confirm the required version after consulting `GoogleSearchAgent`. So the user isn't forced to provide the version as the `DependencyAgent` will try to get the version automatically via the `GoogleSearchAgent`

examples/kg-chat/cypher_message.py
Mohannadcse: It wasn't generated by ChatGPT4 and haven't tried it. I doubt ChatGPT can generate this because it's not aware of the intended schema (nodes labels and their relationships).
So hardcoding it is the safest way to make sure the graph database is constructed as intended.
On the other hand, constructing the graph database for CSV can be done using LLM is easier (working on it now) because all what you need to provide is the CSV header. Then LLM will infer the nodes, their labels, and relationships.

langroid/agent/special/neo4j/neo4j_chat_agent.py
Mohannadcse: I tried here to mimic the implementation of `sql_chat_agent` https://github.com/langroid/langroid/blob/main/langroid/agent/special/sql/sql_chat_agent.py#L197

pchalasani:
Got some errors when exiting -- 
```
<<< Finished Agent DependencyAgent (32)
 Do you want to delete the database? (y/n): y
Database is deleted!
2024-01-15 07:47:20 - ERROR - Failed to write data to connection ResolvedIPv4Address(('34.66.78.163', 7687)) (ResolvedIPv4Address(('34.66.78.163', 7687)))
2024-01-15 07:47:20 - ERROR - Failed to write data to connection ResolvedIPv4Address(('34.66.78.163', 7687)) (ResolvedIPv4Address(('34.66.78.163', 7687)))
2024-01-15 07:47:20 - ERROR - Failed to write data to connection IPv4Address(('927d9aab.databases.neo4j.io', 7687)) (ResolvedIPv4Address(('34.66.78.163', 7687)))
2024-01-15 07:47:20 - ERROR - Failed to write data to connection IPv4Address(('927d9aab.databases.neo4j.io', 7687)) (ResolvedIPv4Address(('34.66.78.163', 7687)))
```

Mohannadcse:
has the dataset been removed? i.e., did you check from neo4j portal?
I receive such errors sometimes, but the operation is performed correctly. 
My hunch is that this error is due to some rate limit when using a free account.

examples/kg-chat/dependency_chatbot.py
Mohannadcse: The intention here was to follow the implementation of `SQLChatAgentConfig` https://github.com/langroid/langroid/blob/0e4fbad6be60e5662bc550cc3df1218007991572/langroid/agent/special/sql/sql_chat_agent.py#L66.

pchalasani: Maybe you can just add `env_prefix  = "NEO4J_"` to this config class.

Mohannadcse:
@pchalasani I've addressed all comments, pls let me if I missed anything

pchalasani:
> @pchalasani I've addressed all comments, pls let me if I missed anything

thanks, making minor edits on top

Mohannadcse:
Just tried it, I didn't face any issues. Was it upon a specific question? 

image

image

pchalasani:
> Was it upon a specific question?

There are a couple of issues -- your sys msg instruction to use `GoogleSearchAgent` is not right -- that's not a valid tool name. When this agent says it wants to know the pkg version, the message would "fail over" to be handled by the GoogleSearchAgent. 
The other issue is -- given this wrong instruction, the agent is generating a function call named `GoogleSearchAgent` which is not a valid tool, so it goes into the GoogleSearchAgent as a user msg, which the LLM tries to handle, leading to the OpenAI err saying "fn call can only be used by assistant", resulting in an infinite loop of these errors.
The second issue requires a fix in langroid core (i.e. even if the sys instruction is wrong, it should not cause these errors; instead an invalid fn-call should be copied into content of msg, and fn-call should be set to None).

I am adding the fixes these on top of the PR, so don't change anything yet.

Mohannadcse:
Noticed discrepancies in the behavior when I enabled tools or used function-calling. Specifically, when tool is enabled the agent asks to provide the version as shown below

image


While if the function-api is enabled (default behaviour), then the agent doesn't ask about the version and proceed with calling the function to construct the dep. graph.
image




> your sys msg instruction to use `GoogleSearchAgent` is not right

the sys msg doesn't mention `GoogleSearchAgent` as a tool. It tries to follow closely something like this example https://github.com/langroid/langroid/blob/0e4fbad6be60e5662bc550cc3df1218007991572/examples/docqa/oai-retrieval-2.py#L53

pchalasani:
@Mohannadcse I made simplifications and fixes
- no need for second agent, just add google search tool
- fix use_tools inconsistencies
- improve sys msgs
- improve kg-chat/readme

Will merge when this passes

pchalasani:
> It tries to follow closely something like this example

there is an important difference between that example and this one -- that assistants example explicitly addresses a specific other agent via recipient tool. Here we are not doing that; instead we are letting it fail over to the search agent. But we actually don't need the search agent, just add the search tool.

examples/kg-chat/README.md
```diff
@@ -0,0 +1,71 @@
+# PyPi Packages Dependency Chatbot
+
+This example uses a `DependencyGraphAgent` 
+(derived from [`Neo4jChatAgent`](https://github.com/langroid/langroid/blob/main/langroid/agent/special/neo4j/neo4j_chat_agent.py)).
+It auto-generates a `neo4j` knowledge-graph based on the dependency
+structure of a given `PyPi` package. You can then ask the chatbot questions
+about the dependency graph. This agent uses two tools in addition to those 
+already available to `Neo4jChatAgent`:
+
+- GoogleSearchTool to find package version and type information
```
Mohannadcse, line 10: any questions actually can be forwarded to this tool, like after obtaining the package and version in the dependency graph, you can search the web about if this version is vulnerable or not.

pchalasani:
> While if the function-api is enabled (default behaviour), then the agent doesn't ask about the version and proceed with calling the function to construct the dep. graph.

could just be random/non-deterministic behavior

examples/kg-chat/README.md
pchalasani: > you can search the web about if this version is vulnerable or not.

ok please update the readme to say that

examples/kg-chat/README.md
```diff
@@ -0,0 +1,71 @@
...
+These settings can be set inside the `.env` file as shown in [`.env-template`](../../.env-template)
+
+# Running the example
+
+Run like this:
+```
+python3 examples/kg-chat/dependency_chatbot.py
+```
+
+`DependencyAgent` then will ask you to provide the name of the `PyPi` package.
```
Mohannadcse, line 47: this paragraph needs to be updated since we decided to proceed with one agent-2 tools

pchalasani: ok please make that change

Mohannadcse:
@pchalasani updated README

Mohannadcse:
@pchalasani there are linting issues from you commit, I'll proceed now with fixing that, unless you are making some mods

pchalasani:
> @pchalasani there are linting issues from you commit, I'll proceed now with fixing that, unless you are making some mods

I'm done, go ahead

Mohannadcse:
not sure why pytest failed. I ran the failed test case `tests/main/test_table_chat_agent.py` and all tests passed.

Issue #309, opened 2023-11-23
Azure OpenAI not working with retrieval


abhiman23897:
I am trying to use Azure OpenAI models with retrieval augmented doc chat. I have replaced the configs wherever I need by using AzureConfig. However, after doing this when I run examples/docqa/chat.py , I get this error only when relevance extractor is running, it works fine before that - 
2023-11-23 14:40:43 - ERROR - Error in OpenAIGPT.achat: 
Original error: Error code: 404 - {'error': {'code': '404', 'message': 'Resource not found'}}
Traceback:
Traceback (most recent call last):
  File "langroid/language_models/openai_gpt.py", line 767, in achat
    result = await self._achat(messages, max_tokens, functions, function_call)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "langroid/language_models/openai_gpt.py", line 984, in _achat
    cached, hashed_key, response = await self._achat_completions_with_backoff(
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "langroid/language_models/openai_gpt.py", line 817, in _achat_completions_with_backoff
    result = await acompletion_call(**kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "langroid/.venv/lib/python3.11/site-packages/openai/resources/chat/completions.py", line 1199, in create
    return await self._post(
           ^^^^^^^^^^^^^^^^^
  File "langroid/.venv/lib/python3.11/site-packages/openai/_base_client.py", line 1482, in post
    return await self.request(cast_to, opts, stream=stream, stream_cls=stream_cls)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "langroid/.venv/lib/python3.11/site-packages/openai/_base_client.py", line 1283, in request
    return await self._request(
           ^^^^^^^^^^^^^^^^^^^^
  File "langroid/.venv/lib/python3.11/site-packages/openai/_base_client.py", line 1326, in _request
    raise self._make_status_error_from_response(err.response) from None
openai.NotFoundError: Error code: 404 - {'error': {'code': '404', 'message': 'Resource not found'}}

**To Reproduce**
[...]

PR #310, opened 2023-11-24
Azure Async use AsyncAzureOpenAI()

pchalasani:


pchalasani:
First step in fixing #309 
Next step is to auto-default to AzureGPT based on config (`.env`)

pchalasani:
On further thought, we will hold off on auto-defaulting to AzureGPT. This will be part of another where we have other auto-defaults.

PR #330, opened 2023-12-21
Add __all__ to __init__.py

### LINKED ISSUE #328: import langroid as lr: using lr.* causes mypy errors
After doing
```
import langroid as lr
```
if we use `lr.*` within code, mypy complains with errors like 
```
module ChatAgent does not explicitly export ...
```

Will be good to see if there's a way to have mypy ignore these, or some other fix.

sanders41:
Closes #328

### DIFF
--- a/langroid/__init__.py
+++ b/langroid/__init__.py
@@ -31,3 +31,22 @@ from .mytypes import (
     DocMetaData,
     Document,
 )
+
+__all__ = [
+    "Agent",
+    "AgentConfig",
+    "ChatAgent",
+    "ChatAgentConfig",
+    "Document",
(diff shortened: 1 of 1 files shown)


pchalasani:
Thanks @sanders41 ! Will test. @nilspalumbo please have you look whenever you are able.

nilspalumbo:
Thanks, looks great! @pchalasani we ought to do this on all of the `__init__.py` files as well (otherwise, the error would occur when, e.g. importing `ChatAgent` directly from `langroid.agent`.). See #357.

nilspalumbo (approved):

PR #332, opened 2023-12-22
add gpt-4-turbo to AzureConfig

Mohannadcse:


langroid/language_models/azure_openai.py
```diff
@@ -91,6 +91,8 @@ def __init__(self, config: AzureConfig):
         # when you deployed a model
         if "35-turbo" in self.config.model_name:
             self.config.chat_model = OpenAIChatModel.GPT3_5_TURBO
+        elif self.config.model_name == "gpt-4":
```
pchalasani, line 94: looks odd - you're picking turbo when model name == 'gpt-4'

Mohannadcse: Yeah, I should have put the reference, you can see GPT-4 (with 1106-preview)
https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/models#model-summary-table-and-region-availability 

So when I want to deploy a model using GPT-4-turbo on Azure, I need to select GPT-4
image

This makes the model name (which is a required parameter in AzureConfig) to become gpt-4
image

While the model name if you use gpt-4 will be `gpt-4-32k`

Mohannadcse:
A better solution is to ask the user to also the model version. This is how we can precisely handle this apsect. Because still the default version of the model name `gpt-4` isn't turbo, which is determined by selecting the model version as `1106-preview`. 
Let me update this PR by adding a new parameter to the AzureConfig to define `model version`

Mohannadcse:
@pchalasani I've added the model version. It's only necessary when the model name is `gpt-4`. That's why I define a func to handle this aspect.

PR #347, opened 2024-01-07
Feature/add sciphi search rag

emrgnt-cmplxty:


tests/main/test_sciphi_search_rag_tool.py
```diff
@@ -0,0 +1,129 @@
+"""
+NOTE: running this example requires setting the SCIPHI_API_KEY environment variable in your `.env` file, as explained in the documentation for the SciPhi API.
```
pchalasani, line 2: Please add link to the SciPhi API here to reduce friction

examples/basic/chat-search.py
```diff
@@ -17,6 +17,7 @@
 
 from langroid.agent.chat_agent import ChatAgent, ChatAgentConfig
 from langroid.agent.task import Task
+from langroid.agent.tools.sciphi_search_rag_tool import SciPhiSearchRAGTool
```
pchalasani, line 20: I think you forgot to include the file that has this tool

pchalasani (changes requested):

tests/main/test_sciphi_search_rag_tool.py
```diff
@@ -0,0 +1,129 @@
...
+
+"""
+import itertools
+import pytest
+
+from langroid.agent.chat_agent import ChatAgent, ChatAgentConfig
```
pchalasani, line 8: We can make this test more focused on this search tool, so you can get rid of the other aspects. I will do the same for the `test_google_search_tool.py`.

examples/basic/chat-search.py
```diff
@@ -88,7 +90,14 @@ def chat(opts: CLIOptions) -> None:
         vecdb=None,
     )
     agent = ChatAgent(config)
-    agent.enable_message(GoogleSearchTool)
+
+    if opts.provider == "google":
```
Mohannadcse, line 94: @pchalasani shouldn't the `provider` flag be defined as `typer.Option` in the main parameters and update `opts = CLIOptions(model=model, provider=provider)` to give the user the ability to supply the value?

pchalasani: yes `provider` should be exposed as an option in the `typer.Option`

examples/basic/chat-search.py
emrgnt-cmplxty: yes, apologies. Adding this and an example use

examples/basic/chat-search.py
emrgnt-cmplxty: I see, thank you I will make this change.

pchalasani:
@emrgnt-cmplxty I don't know if you saw there's another `chat-search.py` in `examples/docqa` which uses Langroid's RAG + google search. In any case, your tool may fit in either place.
The other thought is to look into whether the system prompt needs to change for your tool.
It may just be easier to have a separate example, with a system prompt customized to your tool, not sure.

examples/basic/chat-search.py
```diff
@@ -88,13 +90,20 @@ def chat(opts: CLIOptions) -> None:
...
+    else:
+        raise ValueError(f"Unsupported search provider {opts.provider} specified.")
+
     task = Task(
         agent,
         system_message="""
         You are a helpful assistant. You will try your best to answer my questions.
         If you cannot answer from your own knowledge, you can use up to 5 
-        results from the `web_search` tool/function-call to help you with 
+        results from the `web_search` or `web_search_rag` tool/function-call to help you with
```
pchalasani, line 106: In the sci-phi tool you can just set `request = "web_search"`, so we don't need to change this line. In general in a ToolMessage, the `request` param is used to specify which ChatAgent method will handle the tool. But since this tool stateless, this name actually is immaterial (Langroid attaches the `handle` method to the ChatAgent, under the hood).

pchalasani: Actually ignore the part about changing the sys msg, I see it doesn't need to change.

examples/basic/chat-search.py
```diff
@@ -129,6 +129,9 @@ def chat(opts: CLIOptions) -> None:
 def main(
     debug: bool = typer.Option(False, "--debug", "-d", help="debug mode"),
     model: str = typer.Option("", "--model", "-m", help="model name"),
+    provider: str = typer.Option(
+        "sciphi", "--provider", "-m", help="search provider name (Google, SciPhi)"
```
Mohannadcse, line 133: `-p`? otherwise, it's gonna be the same as the model flag

tests/main/test_sciphi_search_rag_tool.py
emrgnt-cmplxty: If you can do w/ the google search tool i will follow suit

examples/basic/chat-search.py
emrgnt-cmplxty: mistake, sorry.

examples/basic/chat-search.py
emrgnt-cmplxty: sorry, wdym? Happy to make the mod, just a bit confused

pchalasani: I meant that you can set the “request” param in your tool to be same as for google search , “web_search”. This way you don’t need to change the sys msg since it will map to the right tool based on the user choice. 

Ignore the other parts of what I said, sorry about the confusion 😀

tests/main/test_sciphi_search_rag_tool.py
pchalasani: I pushed a simplified test of the google search. I removed 2 tests and their associated setup/constants.

pchalasani:
Overall looks good, I'll test it locally and merge later today or tmr. Thanks for the contrbution.

pchalasani:
@emrgnt-cmplxty a couple other things:
- the `pyproject.toml` needs to include `agent-search` for poetry to install it ( I can do that)
- for the tests to work in GitHub actions, we'll need to add the SCIPHI_API_KEY as a secret -- what key do you suggest here? I could just add my free key, but you have to be aware the gh actions may run pretty often!

pchalasani:
@emrgnt-cmplxty This PR is replaced by #350 where I added some of my fixes (poetry dependency, pydantic version, test failing, etc).

pchalasani:
closing, replaced by #350

PR #350, opened 2024-01-08
Feature/add sciphi search rag

pchalasani:


pchalasani:
@emrgnt-cmplxty I added small fixes on top of your PR. I'm debating whether `agent-search` should go into `extras` category in the dependencies, but it seems to be a fairly lightweight lib so maybe it's fine to go into main.

pchalasani:
> I'm debating whether `agent-search` should go into `extras` category in the dependencies

@emrgnt-cmplxty one possible concern with having the sciphi-test under `tests/main` as opposed to `tests/extras` (which are NOT run in GH actions) -- will the API key continue to work? If it fails for any reason, it will fail the test and this will cause issues with pushes to the repo. Overall I think it's best to both: make `agent_search` an extra dependency (like `litellm`) and add this test under `tests/extras`.

PR #361, opened 2024-01-12
Added line to show how to install hf embeddings version of langroid i…

mturk24:
…f using zsh


pchalasani:
Thanks @mturk24 I had to make a minor tweak on this, so it was easier to use that PR rather than sync w your fork.

PR #363, opened 2024-01-12
mturk/fix README

pchalasani:
- Added line to show how to install hf embeddings version of langroid if using zsh
- tweak README zsh instrucs


pchalasani:
Thanks @mturk24 
I made minor mods on your PR and will use this one instead (as it's too laborious to sync with your fork)

PR #364, opened 2024-01-12
Added streamlit web app w/ local fn call example

mturk24:


pchalasani (changes requested):
Thanks for this. I ran it and noticed it repeats the user input and llm response

image

Also maybe there's a way to display the json in a nicer formatted way, using markdown (I think there's something like `st.markdown` or some other way).

In general the fact that with streamlit we have to manually accumulate conv history is not so nice. Chainlit is overall better for chat-oriented apps, see langroid examples here:
https://github.com/langroid/langroid-examples/tree/main/examples/chainlit

Incidentally I just added this streamlit-based document chat: https://github.com/langroid/langroid-examples/tree/main/examples/docqa/streamlit-app

mturk24:
> We now have Chainlit [integration](https://github.com/langroid/langroid/tree/main/examples/chainlit), and that is a far better way to build chat-based apps with Langroid, so I'm closing this.

Sounds good, sorry for the delay. Thanks!

PR #367, opened 2024-01-16
add kg support and example blog

Mohannadcse:
I still need to add the demo


pchalasani:
in the demo gif, get rid of the usage stats by setting the config `show_stats=False`

PR #374, opened 2024-01-19
Neo4j csv graph

Mohannadcse:
- Update neo4j chat agent
- add new example for csv_kg_chat
- minor updates on the dependency chatbot


pchalasani (changes requested):
The README needs to be updated, to describe how to use this app.

examples/kg-chat/csv_chat.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+class CSVNodeGenerator(ToolMessage):
+    request: str = "create_nodes_and_relationships_from_csv"
+    purpose: str = """Use this tool to create ONLY nodes and their relashipnships based
+    on the created model.
+    Take into account that the Cypher query will be executed while iterating the rows in
+    the CSV file (like this `index, row in df.iterrows()`). So NO NEED to load the CSV.
+    Make sure you send me the cypher query in this format: 
+    - placehoders in <cypherQuery> should be based on the heasers.
```
pchalasani, line 87: typo heasers

examples/kg-chat/csv_chat.py
```diff
@@ -0,0 +1,255 @@
...
+    if remove_null_rows:
+        df = df.dropna()
+    else:
+        if default_value is not None:
+            df = df.fillna(default_value)
+
+    return df
+
+
+class CSVNodeGenerator(ToolMessage):
```
pchalasani, line 80: this can benefit from few-shot examples, which you can do by adding an `examples` class method to this tool

pchalasani:
Looks like you are relying on the LLM to figure out which nodes and relations to add to the kg based on the column names of the csv?

Mohannadcse:
> Looks like you are relying on the LLM to figure out which nodes and relations to add to the kg based on the column names of the csv?

Yes, but the user can update at any time after generating the cypher query that will trigger the tool `create_nodes_and_relationships_from_csv`

examples/kg-chat/csv_kg_chat.py
```diff
@@ -0,0 +1,267 @@
...
+    if remove_null_rows:
+        df = df.dropna()
+    else:
+        if default_value is not None:
+            df = df.fillna(default_value)
+
+    return df
+
+
+class CSVNodeGenerator(ToolMessage):
```
pchalasani, line 81: Rename -> PandasToKGTool
and set `request` to a concise name: `pandas_to_kg` and update the references everywhere.

examples/kg-chat/csv_kg_chat.py
```diff
@@ -0,0 +1,266 @@
...
+                cypherQuery="""MERGE (employee:Employee {name: $employeeName, id: $employeeId})\n
+                MERGE (department:Department {name: $departmentName})\n
+                MERGE (employee)-[:WORKS_IN]->(department)\n
+                SET employee.email = $employeeEmail""",
+                args=["employeeName", "employeeId", "departmentName", "employeeEmail"],
+            ),
+        ]
+
+
+class CSVChatGraphAgent(Neo4jChatAgent):
```
pchalasani, line 109: worth adding this Agent and `PandasToKGTool` to the main branch, since these can be generally useful?

pchalasani (changes requested):

examples/kg-chat/csv_kg_chat.py
Mohannadcse: yes, let's do that if there is no further modifications

pchalasani: Ok please move these

Mohannadcse: I didn't get it, moving them to where?

pchalasani: I meant move these out of examples and into the core codebase, adjacent to `Neo4jChatAgent`

Mohannadcse:
I'll update `README` if this push addresses the feedback.

Mohannadcse:
not sure why `make check` complains about stuff I haven't even touched. While the lint problem in the wf is about something else.

image

image

langroid/agent/special/neo4j/csv_kg_chat.py
```diff
@@ -0,0 +1,272 @@
...
+                    # there is a possibility the generated cypher query is not correct
+                    # so we need to check the response before continuing to the
+                    # iteration
+                    if index == 0 and "successfully" not in response:
+                        print(f"[red]{response}")
+                        return response
+            return "Graph database successfully generated"
+
+
+@app.command()
```
pchalasani, line 150: this belongs under examples/ not here; this file should only contain the Agent and Tool defs. You would then use these in `examples/kg-chat/csv-chat.py` or whatever name it has

pchalasani (changes requested):

Mohannadcse:
I'll decide later whether to put `_load_csv_dataset` and `_preprocess_dataframe_for_neo4j` under the class `CSVChatGraphAgent`

Mohannadcse:
@pchalasani this pr is ready

examples/kg-chat/csv-chat.py
```diff
@@ -0,0 +1,160 @@
...
+                print(
+                    """
+                    [green]The graph database will be generated for the whole dataset...
+                    """
+                )
+
+            csv_kg_chat_agent.csv_dataframe = csv_dataframe
+            csv_kg_chat_agent.csv_location = csv_location
+
+            build_kg_instructions = f"""
```
pchalasani, line 121: the csv-kg-agent system prompts shouldn't be in this example. The system prompts should be internal to that agent, not exposed here.

langroid/agent/special/neo4j/csv_kg_chat.py
```diff
@@ -0,0 +1,139 @@
...
+                MERGE (department:Department {name: $departmentName})\n
+                MERGE (employee)-[:WORKS_IN]->(department)\n
+                SET employee.email = $employeeEmail""",
+                args=["employeeName", "employeeId", "departmentName", "employeeEmail"],
+            ),
+        ]
+
+
+class CSVChatGraphAgent(Neo4jChatAgent):
+    def __init__(self, config: Neo4jChatAgentConfig):
```
pchalasani, line 106: you may need to make a `CSVGraphChatAgentConfig`, so that the csv-chat specific `system_message` lives there, and it should be templated as usual with the provided csv schema. 

Also better name -> `CSVGraphChatAgent`, not `CSVChatGraphAgent`

pchalasani (changes requested):

langroid/agent/special/neo4j/csv_kg_chat.py
pchalasani: or just `CSVGraphAgent` and `CSVGraphAgentConfig` to keep it short.

PR #376, opened 2024-01-21
[Issue 316] Adding metaphor search api as another web searching option.

Alfr3doK1ng:
Added langroid/agent/tools/metaphor_search_tool.py and to handler the request from agent.
Added metaphor_search method in langroid/langroid/parsing/web_search.py to make a POST request to Metaphor search API.

All checks passed.


pchalasani:
1. let's use this metaphor python client instead of making a raw POST request: https://github.com/metaphorsystems/metaphor-python
2. make this an "extra" dependency in `pyproject.toml`
3. import this only within the function that uses it
4. don't add the METAPHOR_API_KEY instructions in the README (trying not to overload the README setup too much). Instead, add the instructions in the chat-search example that uses it.
5. add a test similar to `test_google_search_tool.py` but it under `tests/extra`, so that it is not run automatically in github actions.

pyproject.toml
```diff
@@ -94,7 +94,7 @@ mysql = ["pymysql", "pytest-mysql"]
 litellm = ["litellm"]
 neo4j = ["neo4j"]
 sciphi = ["agent-search"]
-
+metaphor = ["metaphor-python"]
```
pchalasani, line 97: There needs to be a dependency on metaphor marked with optional=True

examples/basic/chat-search.py
```diff
@@ -30,6 +30,15 @@
 This tool requires installing langroid with the `sciphi` extra, e.g.
 `pip install langroid[sciphi]` or `poetry add langroid[sciphi]`
 (it installs the `agent-search` package from pypi).
+
+(c) If using Metaphor Search, you must have METAPHOR_API_KEY 
+environment variables in your `.env` file, as explained in the
+[README](https://github.com/langroid/langroid#gear-installation-and-setup).
```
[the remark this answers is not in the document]
Alfr3doK1ng: got it

pyproject.toml
Alfr3doK1ng: added

Issue #396, opened 2024-02-02
Chroma Dependency Lock


MattGrohoNH:
**Describe the bug**
The `chromadb` package is locked to version `0.3.21` and is beginning to become outdated. 

**To Reproduce**
Adding `langroid` as a dependency to any other Python package which depends on any other version of `chromadb`.

**Expected behavior**
Should be allowed to specify a particular version of Chroma `chromadb>=0.3.29`

**To Fix**
This line in `pyproject.toml`:
`chromadb = "0.3.21"`
should be replaced with:
`chromadb = "^0.3.21"`

Mohannadcse:
@MattGrohoNH thanks for raising this issue. The current PyPi chroma version is `0.4.22`. I'm working on this, we need to make sure the new version doesn't cause any issue.

pchalasani:
@Mohannadcse we can try to get  it to work with >= 0.3.29

Mohannadcse:
tried the recent version `0.4.22` and doesn't work, 
the recent version that `poetry update` allows is `0.3.26`. But when I demanded this version using `pip install chromadb==0.3.29`, it showed me some conflict
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
langroid 0.1.153 requires chromadb==0.3.21, but you have chromadb 0.3.29 which is incompatible.
langroid 0.1.153 requires pydantic==1.10.11, but you have pydantic 1.10.13 which is incompatible.
chainlit 1.0.200 requires fastapi<0.109.0,>=0.100, but you have fastapi 0.85.1 which is incompatible.
Successfully installed chromadb-0.3.29 fastapi-0.85.1 starlette-0.20.4
```

Mohannadcse:
> >= 0.3.29

so we definetly can't do this, because this will get version `0.4.22` installed.

MattGrohoNH:
Hey @Mohannadcse thanks so much for your diligence with this issue!

I have since made a fork to test chromadb versioning and was able to get it working using: `chromadb = ">=0.3.21, <0.4.22"`

Let me know if this works as a temporary fix for `0.4.22`.

( We utilize the PersistentClient from chromadb which is missing in `0.3.21` )

Mohannadcse:
Thanks @MattGrohoNH for the update and the details. 
Which Chromadb you are currently using after making this modification `chromadb = ">=0.3.21, <0.4.22"`?
[...]

PR #400, opened 2024-02-06
allow updating chromadb version

Mohannadcse:
Addresses #396

### DIFF
--- a/examples/basic/chat-search-assistant-local.py
+++ b/examples/basic/chat-search-assistant-local.py
@@ -51,7 +51,7 @@ class QuestionTool(lr.ToolMessage):
     question: str
 
     @classmethod
-    def examples(cls) -> List["ToolMessage"]:
+    def examples(cls) -> List["lr.ToolMessage"]:
         return [
             cls(question="Which superconductor material was discovered in 2023?"),
             cls(question="What AI innovation did Meta achieve in 2024?"),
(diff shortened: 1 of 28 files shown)


Mohannadcse:
Notes:
- I disabled the decorator `@retry_with_exponential_backoff` inside the callable class and solved the chroma problem. 
- The current errors are due to the absence of `hf-embeddings`. Shouldn't `hf-embeddings` be installed as part of Langroid, otherwise, the test cases should be under `extra`.

pchalasani (commented):
@Mohannadcse  this looks good 

HF had an outage yesterday, that may be the reason for the hf-embed err

Also I restored the @retry decorator, and tests seem to pass. Let me know if you noticed some other issue.

pchalasani:
> @Mohannadcse this looks good
> 
> HF had an outage yesterday, that may be the reason for the hf-embed err
> 
> Also I restored the @Retry decorator, and tests seem to pass. Let me know if you noticed some other issue.

never mind, I see the issue with the retry wrapper now. I had not updated the chroma version in my local.

Mohannadcse:
I still need to fix these errors. Before pushing `make check` wasn't showing them, but I noticed them in my last push.

Mohannadcse:
> ok . for the retry, use tenacity as shown here in example 1 https://platform.openai.com/docs/guides/rate-limits/error-mitigation?context=tier-free
> 
> inside the `__call__`, define a fn

I tried this suggestion but didn't work, was raising an error

PR #403, opened 2024-02-12
simple unstructure data to KG and tweaking the neo4j agent

Mohannadcse:


langroid/agent/special/neo4j/neo4j_chat_agent.py
```diff
@@ -31,10 +31,17 @@
...
+class CypherRetrievalTool(ToolMessage):
     request: str = "retrieval_query"
-    purpose: str = """Use this tool to send the Generated Cypher query based on 
-    provided text description and schema."""
+    purpose: str = """Use this tool to send the Cypher query to retreive data from the 
+    graph database based provided text description and schema."""
+    cypher_query: str
+
+
+class CypherCreationTool(ToolMessage):
```
Mohannadcse, line 41: @pchalasani noticed you're making changes on the agent, you may want to incorporate this tool. The CypherQueryTool (changed to CypherRetrievalTool in this PR) tool is supposed to be used for retrieval operations, while this one calls `write_query`.

pchalasani:
@Mohannadcse I've partially merged this with main, and added a new example file `text-kg-triples.py`. Can you make this new example work like this (in the spirit of the original `text-kg.py`)
- start with empty db, schema
- given TEXT=paragraphs p1, p2, and CURRENT_SCHEMA=empty,  create initial schema and nodes, relations
- given TEXT=para p3, and CURRENT_SCHEMA= (obtain from neo4j), create additional schema, nodes, relations.

pchalasani:
also now the `test_neo4j.py` is failing

Mohannadcse:
I'll work on the example and fix the test case

pchalasani:
Also please add a small test in `test_neo4j.py` to test natlang -> kg creation

Mohannadcse:
@pchalasani fixed the test case, and the example `text-kg-triplets.py` works. A few notes:
- if you want to keep the triplets, you may want to increase context length. 
- for the 1st two agents, you may want the agent to terminate after creating the entities/relationships successfully.

pchalasani:
`test_neo4j.py` is failing.

```
       result = task.init(english_query)  # init pending msg
        result = task.step()  # 1. llm -> get schema
        result = task.step()  # 2. agent returns schema
        result = task.step()  # 3. llm -> cypher query for the question
        result = task.step()  # 4. agent returns query result
        result = task.step()  # 5. llm -> formulates english answer
        # english answer
>       assert "inception" in result.content.lower()
E       AssertionError: assert 'inception' in ''
E        +  where '' = <built-in method lower of str object at 0x1034f3e40>()
E        +    where <built-in method lower of str object at 0x1034f3e40> = ''.lower
E        +      where '' = ChatDocument(content='', metadata=ChatDocMetaData(source='LLM', is_chunk=False, id='', window_ids=[], parent=ChatDocum...played=True), function_call=LLMFunctionCall(name='get_schema', to='', arguments={}), tool_messages=[], attachment=None).content
```

Mohannadcse:
There is a possibility the DB wasn't empty when you ran the test case. I see movie entities and entities from the unstructured text. I just deleted everything. 
Did you had the chance to see the schema?

Anyway, I'll make sure the DB is empty before running the test case.

pchalasani:
> I'll make sure the DB is empty before running the test case.

local test still fails:

```

tests/extras/test_neo4j.py Database is deleted!
FUNC: retrieval_query: {"cypher_query":"MATCH (p:Person)-[:ACTED_IN]->(m:Movie) WHERE p.name = 'Leonardo DiCaprio' RETURN m.title AS Movie"}
Stats: N_MSG=2, TOKENS: in=387, out=76, max=1024, ctx=128000, COST: now=$0.0062, cumul=$0.0062 (gpt-4-turbo-preview)
Agent: []
It appears that there are no records of movies that Leonardo DiCaprio acted in within the current dataset. This could be due to the absence of specific data related to Leonardo DiCaprio or the movies he has acted in. If you have any other questions or need information on a different topic, feel free to ask!
```

Mohannadcse:
yeah, this is the problem of LLM, it sometimes generates wrong queries. The query should use the entirety `Actor` not `Person`. 
Probably we need to emphasize in the systems message to strictly use the schema and not to assume entities, relationships,etc..

```
MATCH (p:Actor)-[:ACTED_IN]->(m:Movie) WHERE p.name = 'Leonardo DiCaprio' RETURN m.title AS Movie
```

pchalasani:
> we need to emphasize in the systems message

Ok I'll make some patches to this.

PR #412, opened 2024-02-26
DocChatAgent fix, respect OLLAMA_HOST, multi-schema context description generation

nilspalumbo:
Fixes a bug in `DocChatAgent` which ignores all but the last URL in the `doc_paths`, respects `OLLAMA_HOST` when using Ollama's built in OpenAI API endpoint, adds default context description generation in the multi-schema case and adds tests.

### DIFF
--- a/examples/basic/chat-search-assistant-local.py
+++ b/examples/basic/chat-search-assistant-local.py
@@ -51,7 +51,7 @@ class QuestionTool(lr.ToolMessage):
     question: str
 
     @classmethod
-    def examples(cls) -> List["ToolMessage"]:
+    def examples(cls) -> List[lr.ToolMessage]:
         return [
             cls(question="Which superconductor material was discovered in 2023?"),
             cls(question="What AI innovation did Meta achieve in 2024?"),
--- a/examples/chainlit/chat-search-assistant-local.py
+++ b/examples/chainlit/chat-search-assistant-local.py
@@ -30,7 +30,7 @@ class QuestionTool(lr.ToolMessage):
     question: str
 
     @classmethod
-    def examples(cls) -> List["ToolMessage"]:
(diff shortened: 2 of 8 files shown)


pchalasani:
@nilspalumbo i'm not able to run the new test (M1 Mac) after following the brew install and poetry install instrucs, I get a long err msg ending in this -- 

```
  Reason: tried: '/opt/homebrew/opt/mysql/lib/libmysqlclient.22.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/opt/mysql/lib/libmysqlclient.22.dylib' (no such file), '/opt/homebrew/opt/mysql/lib/libmysqlclient.22.dylib' (no such file), '/opt/homebrew/Cellar/mysql/8.3.0_1/lib/libmysqlclient.22.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/Cellar/mysql/8.3.0_1/lib/libmysqlclient.22.dylib' (no such file), '/opt/homebrew/Cellar/mysql/8.3.0_1/lib/libmysqlclient.22.dylib' (no such file)
```

nilspalumbo:
Whoops, that's odd. I had problems running the mysql tests locally as well (including the old version), but hoped the github actions would be set up correctly to run them. I'll see if I can fix it. Did Rithwik specify anything about the configuration needed for the test to run?

pchalasani:
The `test_sql_chat_agent.py` works fine, it's in `tests/main` not `tests/extras`, and seems to only rely on `sqlalchemy`. Your test is the first sql test in `test/extras`. It won't run in gh actions, which only runs `tests/main` (but eventually we want to start enabling all the `extras` tests to run as well, by installing all the needed extras in the gh workflow).

PR #419, opened 2024-03-01
support azure via -m flag

Mohannadcse:
Azure can be provided using the flag `-m` without the need to make modifications to the`llm_config`. Here is an example `python3.11 examples/basic/chat.py -m azure`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.2.2
+  rev: v0.3.0
   hooks:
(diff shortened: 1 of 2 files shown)


Mohannadcse:
@pchalasani I'll fix the linting issue if there are no concerns about the current implementation

pchalasani:
@Mohannadcse let's revisit this later. There may be a way to have it create an AzureGPT instance based on `type`.

pchalasani (commented):
There may be a better way. I don't want to bring in vbls specific to `AzureGPTConfig` subclass into `OpenAIGPTConfig`

Mohannadcse:
or we can move these variables to the `OpenAIGPTConfig` and remove them from `AzureGPTConfig `

PR #428, opened 2024-03-13
Embed updates

nilspalumbo:
Adds a remote embeddings server for sentence transformer embeds, using `grpc` for remote procedure calls, allowing running `langroid` and generating local embeddings on different machines; currently, this functionality is not handled by tools such as `ollama` and `litellm`.

If possible, uses data parallelism for local embeds; adds configurable batch sizes for all embeddings.

Additionally, improves imports for parsing and vector stores.


nilspalumbo:
The failures were due to `openai` version 1.14.0, which made some changes around assistants that broke `OpenAIAssistant`; should be fixed now.

pchalasani:
The workflow skipts the actual tests because it's like this -- 
```
poetry run coverage run --source=langroid -m pytest -k "not openai_assistant" tests/main
```
But I guess the oai update caused the imports to fail

nilspalumbo:
Yeah, it was an import thing, since `ThreadMessage` no longer exists.

Why are those disabled? They do pass now, after I set the task to be non-interactive, and dropped the `None` value for non-function tools.

The test failing this time works locally, not sure why it would claim not to know the capital of France...

pchalasani:
> Why are those disabled?

At one point caching was not possible for OAI Assistants so I disabled the tests to save cost. Later I implemented caching so maybe they can be re-enabled, if we ensure caching is being used.

pchalasani (commented):
I merged my PR for image-pdf parsing, and included the OpenAI Assistant changes there, so they will overlap with yours (probably no conflicts, but that's syntax-dependent).

Also, I see a ton of format-related changes (87 files changed), not sure if some lint settings changed, or something else.

nilspalumbo:
Yeah, ruff was run on the examples when I was updating the exclusion pattern to ignore the grpc-generated files; I can undo the changes.

pchalasani:
> I can undo the changes

I think that will be good

pchalasani:
pytest errors like these -- 
```
E       Raw response content:
E       b'{"status":{"error":"Wrong input: Collection `test-data` already exists!"},"time":0.362257421}'
```
could be because multiple pytest acitons are runnning at the same time.
I try to kill earlier runs when I update a PR

nilspalumbo:
Hmm, I assumed that would happen automatically. Can we configure it that way?

pchalasani:
> Can we configure it that way?

sure, I don't know how :)

pchalasani (commented):
As a best practice, I've started enabling some of the `tests/extras`, in this case just the `*hf*` tests. When run locally the `test_hf_embeddings.py` is getting stuck (while doing the remote embeddings). Any idea why? Does this test require the remote embeds to be served somewhere? If this is hard to test in gh actions, we could mark this test to be skipped in CI, e.g.

`pytest.mark.skipif(os.environ["CI"] == "true", reason="skip remote embeds in CI")`

PR #431, opened 2024-03-19
Batch processing and console status updates

nilspalumbo:
Generalizes `run_batch_tasks` to accept an argument which generates the tasks; adds optional exception handling (currently, exceptions in the tasks are treated as outputs of `None`). Adds a flag to the batch processing functions and to `DocChatAgent` which toggles the rich status output (disabling avoids potential problems with batch execution of `DocChatAgents`, for example, as only one rich spinner may be active at a time).


langroid/agent/batch.py
```diff
@@ -19,66 +22,161 @@
...
-    output_map: Callable[[ChatDocument | None], Any] = lambda x: x,
+
+def run_batch_task_gen(
+    gen_task: Callable[[int], Task],
+    items: list[T],
+    input_map: Callable[[T], str | ChatDocument] = lambda x: str(x),
+    output_map: Callable[[ChatDocument | None], U] = lambda x: x,  # type: ignore
     sequential: bool = True,
+    batch_size: Optional[int] = None,
+    console_status: bool = True,
```
pchalasani, line 36: I was thinking `quiet_mode` should take care of this. We run all batch tasks in quiet mode, so for example DocChat Agent should look at the global `settings.quiet` and decide whether to use console spinner, no?
If that works, this would obviate a lot of code changes related to this.

langroid/agent/batch.py
```diff
@@ -19,66 +22,161 @@
...
+    status: Callable[[str], Status] | type[nullcontext[str]] = (
+        console.status if console_status else nullcontext  # type: ignore
+    )
 
-    n = len(items)
-    with console.status(f"[bold green]Running {n} copies of {task.name}..."):
-        results = asyncio.run(_do_all())
+    with quiet_mode(not settings.debug), SuppressLoggerWarnings():
+        if batch_size is None:
+            msg = (
```
pchalasani, line 111: simpler - 
```
msg = message or f"[bold green]Running {len(items)} tasks:"
```

langroid/agent/batch.py
```diff
@@ -19,66 +22,161 @@
...
+                f"[bold green]Running {len(items)} tasks:"
+                if message is None
+                else message
+            )
+            with status(msg):
```
pchalasani, line 116: Not sure, but if `status = nullcontext` then the `[bold green]` in the `msg` won't work as intended and show up as a string?

langroid/agent/batch.py
```diff
@@ -19,66 +22,161 @@
...
+T = TypeVar("T")
+U = TypeVar("U")
 
-def run_batch_tasks(
-    task: Task,
-    items: List[Any],
-    input_map: Callable[[Any], str | ChatDocument] = lambda x: str(x),
-    output_map: Callable[[ChatDocument | None], Any] = lambda x: x,
+
+def run_batch_task_gen(
```
pchalasani, line 29: Nice, this is useful when we want to generate slightly different tasks in a batch

langroid/agent/batch.py
```diff
@@ -19,66 +22,161 @@
...
+    gen_task: Callable[[int], Task],
+    items: list[T],
+    input_map: Callable[[T], str | ChatDocument] = lambda x: str(x),
+    output_map: Callable[[ChatDocument | None], U] = lambda x: x,  # type: ignore
```
pchalasani, line 33: `type: ignore` is fine; I don't know how to avoid the mypy complaint

langroid/agent/batch.py
nilspalumbo: No, in that case nothing is shown at all

langroid/agent/batch.py
pchalasani: Maybe worth adding a simple test in `test_batch.py`

langroid/agent/batch.py
nilspalumbo: One other problem here is that the failure to infer U from the type of the default might cause problems with typechecking the outputs, but this might be the best we can do

langroid/agent/batch.py
pchalasani: For example in `chat_agent.py` I conditionally use rich spinner based on `settings.quiet`:

https://github.com/langroid/langroid/blob/main/langroid/agent/chat_agent.py#L684

```
        with ExitStack() as stack:  # for conditionally using rich spinner
            if not self.llm.get_stream() and not settings.quiet:
                # show rich spinner only if not streaming!
                cm = console.status("LLM responding to messages...")
                stack.enter_context(cm)
            if self.llm.get_stream() and not settings.quiet:
                console.print(f"[green]{self.indent}", end="")

```

langroid/agent/batch.py
nilspalumbo: It's also nice in cases the cloning process wouldn't give a correct copy, say when the agent state is modified in some way

langroid/agent/batch.py
nilspalumbo: The spinner in `run_batch_tasks` could also cause conflicts, since it doesn't key off `quiet_mode` (for example, conflicts between this spinner and the spinner from `RelevanceExtractor` could cause a problem). But if all of these are controlled by `quiet_mode`, this would work (though it would probably be a more complex change).

langroid/agent/batch.py
nilspalumbo: I guess you need dependent typing for this?

nilspalumbo: The simplest solution is just not to have a default value, but that's definitely not ideal

PR #451, opened 2024-04-22
upgrade pydantic by using custom version of agent-search

bschreck:
This is a stopgap using a forked version of agent-search in which I just relaxed the pydantic and openai reqs. Doesn't seem like the rest of the repo should fall behind because of this one optional dependency (of which the link to sign up for an API key doesn't even exist anymore)


pchalasani:
> Hmm this is actually a bigger project than it seemed

Yes updating to Pydantic 2 is on our radar to tackle at some point.

pchalasani:
> Doesn't seem like the rest of the repo should fall behind because of this one optional dependency (of which the link to sign up for an API key doesn't even exist anymore)

Good point, I'll remove this from langroid, since that repo seems out-dated. My suspicion is that we'll encounter other pydantic conflicts (poetry bails on the first dep conflict), but we'll see

pchalasani:
I removed the agent-search dependency, and now `poetry install` at least succeeds. But there are numerous mypy errors. Pydantic 2.0 has a lot of api changes, and I'll look into the migration guide and see what's the easiest, safest way to update.

PR #525, opened 2024-07-09
Fix: gpt4o on Azure, Momento, tool message parsing

nilspalumbo:
Corrects OpenAI tool message parsing to use `json.loads` rather than `ast.literal_eval`, enabling correct parsing of Booleans. Handles GPT4o on Azure. As API keys are not required by Qdrant in docker mode (which we treat as "cloud" mode), removes that restriction. Freezes Momento as the vector client we use is removed in the latest release.

### DIFF
--- a/langroid/parsing/parse_json.py
+++ b/langroid/parsing/parse_json.py
@@ -1,5 +1,6 @@
+import ast
 import json
-from typing import Any, Iterator, List
+from typing import Any, Dict, Iterator, List, Union
 
 import yaml
 from pyparsing import nestedExpr, originalTextFor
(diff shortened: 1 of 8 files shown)


langroid/vector_store/qdrantdb.py
nilspalumbo, line 1: We may wish to create a "docker" mode as well to distinguish.

langroid/parsing/parse_json.py
```diff
@@ -73,6 +74,31 @@ def add_quotes(s: str) -> str:
...
+        result = ast.literal_eval(json_string)
+        if isinstance(result, (dict, list)):
+            return result
+    except (ValueError, SyntaxError):
+        pass
+
+    # If ast.literal_eval fails or returns non-dict/list, try json.loads
+    try:
+        str = add_quotes(json_string)
+        result = json.loads(str)
```
pchalasani, line 92: @nilspalumbo see some discussions here about preferring `ast.literal_eval`:

https://community.openai.com/t/response-from-tools-call-function-arguments-is-not-consistently-valid-json/594040/2

https://community.openai.com/t/function-calls-and-streaming/263393

In any case, I changed it to use a new `parse_imperfect_json()`  fn I created in `parse_json.py`: it first tries  `literal_eval` and then `json.loads`

langroid/vector_store/qdrantdb.py
pchalasani: So we'd have a config option "docker=True/False" ? Sure, please add that.

pchalasani (changes requested):

langroid/parsing/parse_json.py
pchalasani: @nilspalumbo please see if the new tests in `test_json.py` address the scenarios that you had in mind when you made this change.

nilspalumbo: Yeah, this works! I hadn't known about this issue, looks like ideally we'd want a parser which tolerates mixed Python/JSON syntax but this is enough for now.

PR #526, opened 2024-07-11
Infer request

nilspalumbo:
When only one handled `ToolMessage` is compatible with the arguments provided by the LLM, but the LLM did not specify `request` (especially common on weaker models, but still occurs with GPT-4o), uses the only compatible function call.

### DIFF
--- a/langroid/agent/base.py
+++ b/langroid/agent/base.py
@@ -799,24 +799,23 @@ class Agent(ABC):
         ):
             handled = [self.llm_tools_map[r] for r in self.llm_tools_handled]
 
-            def can_parse(tool: type[ToolMessage]) -> bool:
+            def maybe_parse(tool: type[ToolMessage]) -> Optional[ToolMessage]:
                 try:
-                    tool.parse_obj(json_data)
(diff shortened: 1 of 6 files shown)


pchalasani:
@nilspalumbo there was an interesting edge-case triggerred by the sql-chat `GetTableNames` tool -- the tool has _no parameters_, and the `maybe_parse` was attempting to parse the _results_ of running `GetTableSchemas` (which is a dict) into `GetTableNames`, and was succeeding because the `ToolMessage` did not have `Config.extras="forbid"` enabled:

```python
json_data = {
  description = "The departments table...",
  columns = {'id': 'INTEGER', 'name': 'VARCHAR'},
}
```

This was being successfully parsed as an instance of `GetTableNames` which has no required params. 

The fix is to add this to `ToolMessage`:

```python

class ToolMessage:
    ...
    class Config:
         extra="forbid"
```

pchalasani:
great pr, thanks @nilspalumbo

pchalasani (approved):

PR #543, opened 2024-08-15
download punkt

Mohannadcse:
'punkt' tokenizer is required by `sentences = nltk.sent_tokenize(paragraph)` in `langroid/parsing/utils.py`


Mohannadcse:
I tried to solve it inside pyproject.toml, but noticed it requires a post command after `poetry install`. Then decided to do the installation inside `__init__.py`

examples/basic/chat_multi_config.py-e
```diff
@@ -0,0 +1,123 @@
+"""
```
pchalasani, line 1: this file shouldn't be in the PR

pchalasani (commented):
There's nltk download related code here:
https://github.com/langroid/langroid/blob/main/langroid/parsing/search.py#L118

this downloads nltk stuff when needed.

I don't think we should include nltk download as part of langroid import.

Since the nltk-not found issue came up in a specific example, maybe it's best to add a comment at the top of that script showing an easy cmd  to download this.

Mohannadcse:
> I don't think we should include nltk download as part of langroid import.
> 
> Since the nltk-not found issue came up in a specific example, maybe it's best to add a comment at the top of that script showing an easy cmd to download this.

It was deep in the call chain (sentences = nltk.sent_tokenize(paragraph) in langroid/parsing/utils.py), so it will be used by other examples, or future examples. Therefore, I was thinking about a general solution instead on defining that inside a specific example.

pchalasani:
> It was deep in the call chain (sentences = nltk.sent_tokenize(paragraph) in langroid/parsing/utils.py)

Can a fix be added in this function/file then ?

pchalasani:
> There is a bug in the function

then how was it working until now?

Mohannadcse:
ok, I dig more into the issue, here is my observation:
- after calling `download_nltk_resource`, the resources are downloaded, but in some circumstances, specifically for the resource `punkt`, the zip file `punkt.zip` isn't uncompressed. Therefore, `nltk.data.find` doesn't find `punkt` and raises an error.
- When I manually unzip the file `punkt.zip`, `nltk.data.find` doesn't raise any error. 

In some circumstances (not all the time), ChatGPT recommends to use the associated path (i.e, tokenizers/punkt, corpora/wordnet, corpora/stopwords) of each resource while using the API `nltk.data.find` instead of using only the resource name. 

To summarize, we don't need this PR. NLTK resources are installed using `download_nltk_resource`. However, we might just need to add a message to tell the user to check if the `punkt.zip` is uncompressed under the path `nltk_data/tokenizers`, which is based on my trials fluctuating between being inside the user home directory or `.venv`.

pchalasani:
> * but in some circumstances, specifically for the resource `punkt`, the zip file `punkt.zip` isn't uncompressed. Therefore, `nltk.data.find` doesn't find `punkt` and raises an error.

So why not check after it is downloaded and unzip if needed ?

Mohannadcse:
@pchalasani I added `punkt_tab`, noticed it's required while using the example [multi-agent-triage.py](https://github.com/langroid/langroid/commit/69fefbd134b8c1db78a3a162a4a4736a5e2cf344#diff-21f04012013b8108f1ec015f8d50d650143b22a04f7a4d7c9b2586b81aa2aa64)

pchalasani:
Resolved in recent PRs

PR #550, opened 2024-08-24
feat: task.run/run_async polymorphism support

pchalasani:
This PR supports `task.run/run_async` polymorphism, see this test: https://github.com/langroid/langroid/blob/task-run-polymorphism/tests/main/test_task_run_polymorphic.py

And also see an example of using this polymorphism to simplify a multi-agent search-critic system without inter-agent orchestration:
https://github.com/langroid/langroid/blob/task-run-polymorphism/examples/basic/multi-agent-search-critic-no-orch/main.py

The syntax is:
```
result = task.run(any_type_input, return_type=MyReturnType)
```

I looked into avoiding having to specify `return_type` and instead infer it from a result variable type-annotation, something like:

```
result: MyReturnType = task.run(some_input)
```

but it involves using unsafe things like inspecting the call frame.

@nilspalumbo Let me know if you have any ideas for enhancing etc
(No need to review the entire PR just the resulting end-user API, whether it can be enhanced, and if so, should this PR be merged and your enhancement can be on top of it)

### DIFF
--- a/langroid/agent/base.py
+++ b/langroid/agent/base.py
@@ -18,7 +18,10 @@ from typing import (
     Set,
     Tuple,
     Type,
+    TypeVar,
     cast,
+    get_args,
+    get_origin,
     no_type_check,
 )
 
@@ -46,7 +49,6 @@ from langroid.parsing.parse_json import extract_top_level_json
 from langroid.parsing.parser import Parser, ParsingConfig
 from langroid.prompts.prompts_config import PromptsConfig
 from langroid.pydantic_v1 import (
-    BaseModel,
     BaseSettings,
     Field,
     ValidationError,
@@ -56,12 +58,15 @@ from langroid.utils.configuration import settings
 from langroid.utils.constants import DONE, NO_ANSWER, PASS, PASS_TO, SEND_TO
 from langroid.utils.object_registry import ObjectRegistry
 from langroid.utils.output import status
+from langroid.utils.types import from_string, to_string
 from langroid.vector_store.base import VectorStore, VectorStoreConfig
 
 ORCHESTRATION_STRINGS = [DONE, PASS, PASS_TO, SEND_TO]
 console = Console(quiet=settings.quiet)
 
 logger = logging.getLogger(__name__)
+
+T = TypeVar("T")
 
 
 class AgentConfig(BaseSettings):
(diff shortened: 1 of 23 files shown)


langroid/agent/base.py
```diff
@@ -1260,51 +1282,112 @@ def maybe_parse(tool: type[ToolMessage]) -> Optional[ToolMessage]:
...
+            return cast(T, content_any)
+
+        tools = self.get_tool_messages(msg, all_tools=True)
+
+        if get_origin(output_type) is list:
+            list_element_type = get_args(output_type)[0]
+            if issubclass(list_element_type, ToolMessage):
+                # output_type = List[<ToolMessage],
+                # i.e. list of ToolMessage-derived objects
+                return cast(T, tools)
```
nilspalumbo, line 1369: As in the next case, we might want to filter the tools by those which are instances of `output_type`

langroid/agent/task.py
```diff
@@ -582,16 +587,44 @@ def reset_all_sub_tasks(self) -> None:
...
+        max_tokens: int = 0,
+        session_id: str = "",
+        allow_restart: bool = True,
+        return_type: Type[T],
+    ) -> Optional[T]: ...  # noqa
+
     def run(
         self,
-        msg: Optional[str | ChatDocument] = None,
+        msg: Any | None = None,
```
nilspalumbo, line 619: Here, and in `run_async`, this can just be `Any`

langroid/agent/base.py
nilspalumbo: Pushed this change, but the original may be more readable, let me know what you think

langroid/agent/task.py
nilspalumbo: Pushed this change, but the original may be more readable, let me know what you think

langroid/agent/base.py
nilspalumbo: Made this change

nilspalumbo:
Thanks! Looks great, I've added some changes which enable a cleaner interface; we can now say `task[return_type].run(...)`. Currently, the argument to `task.run()` overrides, and I've added a `default_return_type` argument to the `Task` constructor; would it be better to use the new interface exclusively?

pchalasani:
> Thanks! Looks great, I've added some changes which enable a cleaner interface; we can now say `task[return_type].run(...)`. Currently, the argument to `task.run()` overrides, and I've added a `default_return_type` argument to the `Task` constructor; would it be better to use the new interface exclusively?

the `__get_item__` idea is clever! I think we can keep the original `return_type`
Will merge once tests pass

PR #552, opened 2024-08-27
feat: run_task_batch stop on first result

pchalasani:


pchalasani:
@nilspalumbo this implements `run_task_batch` with option to stop on first result. This can be used to support async-subtasks of a task later.

nilspalumbo:
Thanks, that's a great feature! Should we add an additional function argument to `run_task_batch` that allows us to terminate at the first valid result? That would be needed for the async subtasks.

pchalasani:
> an additional function argument to `run_task_batch`

It's already there -- https://github.com/langroid/langroid/blob/batch-task-early-stop/langroid/agent/batch.py#L153

or did you mean something else?

nilspalumbo:
I mean a `Callable` argument `valid`, say, which takes the `ChatDocument` output which returns whether we should stop (say, we might want to continue if the output is `NO_ANSWER`). By default, we can just accept all however, which would work the same as the current version.

pchalasani:
> whether we should stop

Do you mean
- (A) `valid(..)` would take the result of _each of the tasks in the async-batch_ and determine if each is  "done" or should continue running? or 
- (B) `valid(.)` would check the _overall batch-result_ (i.e. any non-None result in the group) and determine whether to continue or not?

Letting an individual task in the group "continue" seems like an unclear behavior. After `r = task.run()` returns some result `r`, letting it "continue" would mean running `task.run(r)` ... so I'm confused. Or maybe you mean we take a snapshot of each task's "intermediate" result and determine which ones should stop.

nilspalumbo:
I was thinking that `valid` would operate on each of the tasks in the batch independently, and we would terminate if any returned `True`.

pchalasani:
Actually I see what you mean  - you mean a `valid()` criterion similar what we have in `task.step()` that determines whether the "step" found a valid result... I have to think about that

nilspalumbo:
Yeah that's right, we'd need this for async subtasks.

nilspalumbo:
The batch processing functions now continue until some non-`None` output is returned; they all now return `list[Optional[U]]`. The `_do_task` function now optionally returns the index (avoiding the `list.index(future)` operation). Additionally added exception handling to the `stop_on_first_result` branch and fixed a minibatch bug.

The typing updates are however, a bit messy; for instance, we always return a `list[U]` without `stop_on_first_result` but we must explicitly cast at each use as mypy cannot infer this.

pchalasani:
When I update (just pushed) the `test_task_gen_batch` to have `sleep(10)` in task 0 and task 2 (so the middle case 1 is fastest), I see two problems:
- the total time taken is longer, i.e. is dependent on `sleep(10)` whereas I expected it to be as fast as the fastest task 2 , which doesn't have a sleep.
- the task 0 finishes, and task 1 and 2 don't.

So something is still wrong with the implementation.

nilspalumbo:
Also, note on the exception handling for the `stop_on_first_result` case: `asyncio.wait` does not have a `handle_exceptions` option, so we cannot independently handle exceptions between multiple tasks (the results of a successful task may be lost if another task in the minibatch raises an exception).

nilspalumbo:
Fixed the exception handling issue for `stop_on_fiirst_result`

pchalasani:
@nilspalumbo that specific `test_task_gen_batch` still fails so there's something we're missing -- the implementation doesn't seem to be able to get what we expect to be the fastest task result. Or maybe it has something to do with test itself, i.e. `MockLM.achat` not being truly async (though the test fails even if I add a `await asyncio.sleep(0.1)`.

nilspalumbo:
I added support for a`response_fn_async`  to `MockLM` and updated the test to use it, it should work as expected now. Note though that the expected behavior with batching is that the fastest task in the first minibatch, not the fastest task overall, should finish.

pchalasani:
> it should work as expected now

Excellent, thanks!



> the fastest task in the first minibatch, not the fastest task overall, should finish.

That's right. So I guess if users want the behavior of "get the first result out of k tasks", then they shouldn't batch the tasks (i.e. choose batch size >= k)?

nilspalumbo:
Yeah, that's how it'll work by default; the behavior should be intuitive enough, but it is the only case where batching would be expected to affect the output.

Issue #563, opened 2024-09-13
0.60 release breaks groq/llama-3.1-70b-versatile due to max_completion_tokens


FilterJoe:
**Describe the bug**
Run pytest for test_agent.py or anything else that uses llama-3.1-70b-versatile through openai_gpt.py and it fails with:

unexpected keyword argument 'max_completion_tokens'

bug introduced by change in openai_gpt.py with version 0.60

**To Reproduce**
pytest -s tests/main/test_agent.py --nc --show --m groq/llama-3.1-70b-versatile

**Expected behavior**
Expected test to pass (with 1 warning), which it did with version 0.52

**Desktop (please complete the following information):**
OS: Debian 12 VM on Mac mini m2 pro

**Additional context**
If changing from max_tokens to max_completion_tokens is required for adding support to o1, then need to special case it for o1 and leave it alone for other models.

pchalasani:
> need to special case it for o1

ouch -- yes, they really made a mess with this o1 model, I had to tweak my code in a few places.
I'll patch it.

pchalasani:
awaiting tests in PR #564

pchalasani:
fixed in PR #564

PR #564, opened 2024-09-13
max tokens param rename

pchalasani:
- **fix: special-case o1 to rename max_tokens param**
- **lint: openai_gpt.py**


pchalasani:
fixes #563

PR #572, opened 2024-09-25
add: token usage metric comes from OpenAI response

kartikhans:


pchalasani:
Thanks @kartikhans please add suitable tests for this. There are existing ones for token usage, you can search for it.

PR #578, opened 2024-10-11
Allow async streaming through callback

VitaliyKRS:
The main problem is that langroid only supports streaming tokens in sync mode even in async function. This can be seen as `OpenAIGPT._stream_response_async` calls `_process_stream_event` with sync streamer callback, for this reason  `ChainlitAgentCallbacks.start_llm_stream` passes sync callback `stream_token` to `self.llm.config.streamer`  wraps `cl.Step.self.stream.stream_token(t)` into `run_sync()`  which makes it inefficient, this is especially evident if I need LLM to stream a large chunk of text, in while llm has already streamed it, run_sync still continues to stream it in ui. 

For this purpose I added the ability to pass an asynchronous streamer callback for asynchronous responses, this significantly improves the speed of streaming from the agent to the UI.

### DIFF
--- a/langroid/language_models/openai_gpt.py
+++ b/langroid/language_models/openai_gpt.py
@@ -836,10 +836,10 @@ class OpenAIGPT(LanguageModel):
                 sys.stdout.flush()
                 await self.config.streamer_async(event_args)
 
-        if event_tool_deltas is not None:
+        if event_tool_deltas is not None and not silent:
             # print out streaming tool calls, if not async
             for td in event_tool_deltas:
-                if td["function"]["name"] is not None and not silent:
+                if td["function"]["name"] is not None:
                     tool_fn_name = td["function"]["name"]
                     sys.stdout.write(
                         Colors().GREEN + "OAI-TOOL: " + tool_fn_name + ": "
--- a/langroid/language_models/openai_gpt.py
+++ b/langroid/language_models/openai_gpt.py
@@ -820,41 +820,37 @@ class OpenAIGPT(LanguageModel):
             if not silent:
                 sys.stdout.write(Colors().GREEN + event_text)
                 sys.stdout.flush()
-            await self.config.streamer(event_text)
+                await self.config.streamer_async(event_text)
         if event_fn_name:
             function_name = event_fn_name
             has_function = True
             if not silent:
                 sys.stdout.write(Colors().GREEN + "FUNC: " + event_fn_name + ": ")
                 sys.stdout.flush()
-
-            await self.config.streamer(event_fn_name)
+                await self.config.streamer_async(event_fn_name)
 
         if event_args:
             function_args += event_args
             if not silent:
                 sys.stdout.write(Colors().GREEN + event_args)
                 sys.stdout.flush()
-
-            await self.config.streamer(event_args)
+                await self.config.streamer_async(event_args)
 
         if event_tool_deltas is not None:
             # print out streaming tool calls, if not async
             for td in event_tool_deltas:
-                if td["function"]["name"] is not None:
+                if td["function"]["name"] is not None and not silent:
                     tool_fn_name = td["function"]["name"]
-                    if not silent:
-                        sys.stdout.write(
-                            Colors().GREEN + "OAI-TOOL: " + tool_fn_name + ": "
-                        )
-                        sys.stdout.flush()
-                    await self.config.streamer(tool_fn_name)
+                    sys.stdout.write(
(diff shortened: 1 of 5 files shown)


pchalasani:
Thanks @VitaliyKRS , could you point me to a chainlit script where I can see the difference in behavior and/or performance before and after this change?

pchalasani:
@VitaliyKRS your changes look reasonable, but I just ran `pytest -xvs tests/main/test_llm_async.py` and it fails. I haven't looked too deeply into it yet, but just letting you know.

langroid/language_models/openai_gpt.py
```diff
@@ -730,37 +820,41 @@ def _process_stream_event(
             if not silent:
                 sys.stdout.write(Colors().GREEN + event_text)
                 sys.stdout.flush()
-                self.config.streamer(event_text)
+            await self.config.streamer(event_text)
```
pchalasani, line 823: This should be inside the `if not silent:` block, and similarly elsewhere in this code.

pchalasani: Also, calling `self.config.streamer(event_text)` results in an error when we directly use the LLM (i.e. not through a `ChatAgent`), since the default setting of `self.config.streamer` is `noop_fn()` which is a sync fn.

pchalasani: One solution could be to have a separate `config.streamer_async` in `LLMConfig` defined as an async version of noop_fn, and use this streamer inside `_process_event_stream_async`

pchalasani (changes requested):

pchalasani:
> @VitaliyKRS your changes look reasonable, but I just ran `pytest -xvs tests/main/test_llm_async.py` and it fails. I haven't looked too deeply into it yet, but just letting you know.

I found the cause (see my other comments)

vkrasokha:
@pchalasani Thank you for the feedback, I'll look for the remark

VitaliyKRS:
@pchalasani Thank you for your feedback, I've updated the code as you requested, please check it

langroid/language_models/openai_gpt.py
```diff
@@ -708,7 +707,98 @@ def _process_stream_event(
...
+            sys.stdout.flush()
+            self.config.streamer(event_fn_name)
+
+        if event_args:
+            function_args += event_args
+            sys.stdout.write(Colors().GREEN + event_args)
+            sys.stdout.flush()
+            self.config.streamer(event_args)
+
+        if event_tool_deltas is not None:
```
pchalasani, line 744: add condition `and not silent` here, as it was in previous code.

pchalasani (changes requested):

langroid/language_models/openai_gpt.py
VitaliyKRS: Do you want to add separate `self.config.stream_quiet` for the sync version in the same way as `self.config.async_stream_quiet`?

pchalasani: No I meant let's not worry about suppressing output in _process_stream_event -- only focus on suppressing it in the async version. Does that make sense?

VitaliyKRS: I see that there is missing check in _process_stream_event_async, I misunderstood because you pointed to the sync function first.

VitaliyKRS: Fixed

pchalasani:
> Thanks @VitaliyKRS , could you point me to a chainlit script where I can see the difference in behavior and/or performance before and after this change?

Changes look good, thanks @VitaliyKRS . Any suggestions on what script I can test, to see the performance difference?

pchalasani:
@VitaliyKRS I created a temp branch + PR to run the full test suite. Will merge yours once they passs.

VitaliyKRS:
> Any suggestions on what script I can test, to see the performance difference?
Sure, I'll create one

pchalasani:
@VitaliyKRS the tests are passing. Still worth seeing if there is an example that demonstrates the speedup.

VitaliyKRS:
But anyway this should increase performance, because running asynchronous code synchronously, as in the provided run_sync function, can lead to performance issues by blocking the event loop, which prevents other tasks from being handled concurrently, negating the advantages of async programming. It introduces overhead due to context switching between threads and increases the risk of deadlocks or disrupted execution flow. Additionally, it removes the inherent parallelism of async I/O-bound operations, leading to lower throughput and higher latency.

PR #589, opened 2024-10-16
Strict decoding

nilspalumbo:
Adds support for strict tools and structured JSON outputs (supported by OpenAI as well as local LLM providers such as llama.cpp and vLLM) for improved reliability. This mechanism is enabled by default as a recovery in the presence of tool errors when possible.

Additionally, updates the `ToolMessage` schema generation for "request". Currently, "request" is an arbitrary string type, but our tool handling mechanism requires an exact match with the default value. The schema now returns an enumeration type for "request" which requires the correct value; in particular, this ensures correctness when structured outputs are enabled.


langroid/agent/chat_agent.py
```diff
@@ -581,6 +644,33 @@ def enable_message(
...
+                value: output_type  # type: ignore
+
+            output_type = OutputFormat
+
+        clone.output_format = output_type
+        if issubclass(output_type, ToolMessage):
+            clone.enabled_requests_for_inference = {
+                output_type.default_value("request")
+            }
+        if self.config.use_functions_api:
```
nilspalumbo, line 668: If an agent with a strict output format should be able to produce a function call, `output_format` must be a `ToolMessage` and `output_format` should be the only usable tool. Using the native OpenAI tool calling capabilities doesn't ensure this, but we could choose to permit both structured outputs and arbitrary tool calls with a flag.

langroid/agent/chat_agent.py
```diff
@@ -646,7 +794,21 @@ def llm_response(
             else (message.oai_tool_choice if message is not None else "auto")
         )
         with StreamingIfAllowed(self.llm, self.llm.get_stream()):
-            response = self.llm_response_messages(hist, output_len, tool_choice)
+            try:
+                response = self.llm_response_messages(hist, output_len, tool_choice)
+            except openai.BadRequestError as e:
+                if self.any_strict:
+                    self.disable_strict = True
```
nilspalumbo, line 801: As OpenAI doesn't support the full JSON schema, it is possible that enabling strict output (which we do automatically) could cause errors; this recovery mechanism is designed to catch them. I'm not aware of precise specifications of the subset of JSON schema supported, but if they exist it would be much better to validate the schemas directly.

We could detect which tools are triggering the failures from the error messages (which contain the responsible tool name) but this may be brittle.

langroid/agent/tool_message.py
```diff
@@ -22,6 +23,55 @@
...
+K = TypeVar("K")
+
+
+def remove_if_exists(k: K, d: dict[K, Any]) -> None:
+    """Removes key `k` from `d` if present."""
+    if k in d:
+        d.pop(k)
+
+
+def format_schema_for_strict(schema: Any) -> None:
```
nilspalumbo, line 35: This is used on all schemas for strict mode (both output formats and strict tool schemas), and should ensure compatibility (however, there may be edge cases as I'm not aware of precise documentation of the OpenAI JSON schema subset for strict mode). 

This may not be equivalent to the original schema, but will be in typical cases (e.g. "allOf" will occasionally be generated by Pydantic but generally only with a singleton set of types, in which case it is equivalent to "anyOf"), and the loss of "default" means that any defaults can only be expressed in an external prompt.

Given these requirements, is it safe to enable strict mode by default when supported?

pchalasani (changes requested):
Thanks for this @nilspalumbo 
Still reviewing it, but I was thinking it would be good to add a couple of example scripts to illustrate this functionality, and a short section in the docs under Notes-Updates like here: https://langroid.github.io/langroid/notes/xml-tools/

langroid/agent/task.py
```diff
@@ -732,8 +733,37 @@ def run(
         if return_type is None:
             return_type = self.default_return_type
 
+        # Take a final strict decoding step
         if return_type is not None and return_type != ChatDocument:
-            return self.agent.from_ChatDocument(final_result, return_type)
+            parsed_result = self.agent.from_ChatDocument(final_result, return_type)
```
pchalasani, line 738: we need this in the `run_async()` as well -- maybe factor it out as a `_<method>`  and use it in both

docs/tutorials/local-llm-setup.md
```diff
@@ -89,6 +89,41 @@ as the `chat_model` param in the `OpenAIGPTConfig` as in the previous section.
...
+In addition, your `llama.cpp` may be built with support for simplified management of HuggingFace models (specifically, `libcurl` support is required); in this case, `llama.cpp` will download HuggingFace models to a cache directory, and the server may be run with:
+```bash
+llama-server \
+      --hf-repo Qwen/Qwen2.5-Coder-7B-Instruct-GGUF \
+      --hf-file qwen2.5-coder-7b-instruct-q2_k.gguf
+```
+
+To use the model with Langroid, specify `llamacpp/localhost:{port}` as the `chat_model`; the default port is 8080.
+
+## Setup vLLM with a model from HuggingFace
```
pchalasani, line 109: Any restrictions on where we can use vllm, e.g. linux only, and/or is this only relevant when using a GPU?

nilspalumbo: Right now it's linux only. I don't think VLLM has any support for accelerators on M series chips; there's a PR for ARM support right now but that would still be CPU only. Currently, CPU only mode seems to require x86. 

With the new PR it may run on macOS, but llama.cpp would likely still be preferred in that case (as it supports the accelerators, etc.).

nilspalumbo: Another note is that installation with backends other than NVIDIA GPUs requires rebuilding from source.

langroid/agent/base.py
```diff
@@ -1102,12 +1115,17 @@ def get_oai_tool_calls_classes(self, msg: ChatDocument) -> List[ToolMessage]:
                     or you need to enable this agent to handle this fn-call.
                     """
                 )
+                if tool_name not in self.all_llm_tools_known:
```
pchalasani, line 1118: We should omit these two lines - it's messing up the intended logic -- i.e. if any tool is found valid, then all_errors will be set to False down below, and things will be fine. Or am I missing something?

nilspalumbo: You're right; they should be dropped.

langroid/agent/tool_message.py
pchalasani: I think it's fine -- 

- we could automatically retry with non-strict mode if there's an error (I think you already have this?)

We can also have a config `use_strict_mode_if_available` to control this behavior, and if there's an error we can print a friendly warning saying strict mode was used since it was allowed, and suggest turning off this flag.

langroid/agent/task.py
```diff
@@ -732,8 +733,37 @@ def run(
...
+
+            if (
+                parsed_result is None
+                and isinstance(self.agent, ChatAgent)
+                and self.agent._json_schema_available()
+            ):
+                strict_agent = self.agent[return_type]
+                output_args = strict_agent._function_args()[-1]
+                if output_args is not None:
```
pchalasani, line 748: Need some comment here -- why does `output_args is not None` indicate  that the result did not adhere to the schema? Maybe becomes clear later but not immediately clear.

langroid/agent/task.py
```diff
@@ -895,8 +925,37 @@ async def run_async(
         if return_type is None:
             return_type = self.default_return_type
 
+        # If possible, take a final strict decoding step
+        # when the output does not match `return_type`
         if return_type is not None and return_type != ChatDocument:
-            return self.agent.from_ChatDocument(final_result, return_type)
+            parsed_result = self.agent.from_ChatDocument(final_result, return_type)
```
pchalasani, line 931: the async is an exact dupe of the sync, so maybe use a helper function to encapsulate this code block?

langroid/language_models/azure_openai.py
```diff
@@ -8,6 +8,12 @@
     OpenAIGPTConfig,
 )
 
+azureStructuredOutputList = [
+    "2024-08-06",
```
pchalasani, line 12: there's a more recent one I think

langroid/language_models/azure_openai.py
```diff
@@ -136,20 +147,20 @@ def handle_gpt4_model(self) -> None:
         If the version is not set, it raises a ValueError indicating
         that the model version needs to be specified in the ``.env``
         file.  It sets `OpenAIChatMode.GPT4o` if the version is
-        '2024-05-13', `OpenAIChatModel.GPT4_TURBO` if the version is
-        '1106-Preview', otherwise, it defaults to setting
-        `OpenAIChatModel.GPT4`.
+        '2024-08-06' or '2024-05-13', `OpenAIChatModel.GPT4_TURBO` if
```
pchalasani, line 150: there's been a more recent `GPT4o` release

langroid/language_models/openai_gpt.py
```diff
@@ -484,8 +508,28 @@ def __init__(self, config: OpenAIGPTConfig = OpenAIGPTConfig()):
...
+        elif self.config.chat_model.startswith("vllm/"):
+            self.supports_json_schema = True
+            self.config.chat_model = self.config.chat_model.replace("vllm/", "")
+            self.api_key = VLLM_API_KEY
+            self.api_base = self.config.api_base or "http://localhost:8000/v1"
+            if not self.api_base.startswith("http"):
+                self.api_base = "http://" + self.api_base
+            if not self.api_base.endswith("/v1"):
+                self.api_base = self.api_base + "/v1"
+        elif self.config.chat_model.startswith("llamacpp/"):
```
pchalasani, line 520: add comment showing typical expected `chat_model` value

langroid/language_models/openai_gpt.py
```diff
@@ -484,8 +508,28 @@ def __init__(self, config: OpenAIGPTConfig = OpenAIGPTConfig()):
...
+            self.supports_json_schema = True
+            self.api_base = self.config.chat_model.split("/", 1)[1]
+            if not self.api_base.startswith("http"):
+                self.api_base = "http://" + self.api_base
+            self.api_key = LLAMACPP_API_KEY
         else:
             self.api_base = self.config.api_base
+            self.supports_strict_tools = self.api_base is None
```
pchalasani, line 528: comment showing why `api_base is None` means it supports strict tools

langroid/language_models/openai_gpt.py
```diff
@@ -1622,18 +1676,30 @@ def _prep_chat_completion(
                 )
             )
         if tools is not None:
+            if self.config.parallel_tool_calls is not None:
+                args["parallel_tool_calls"] = self.config.parallel_tool_calls
+
+            if any(t.strict for t in tools) and (
+                self.config.parallel_tool_calls is None
```
pchalasani, line 1683: simplify - just need:
```
any(...) and self.config.parallel_tool_calls
```

langroid/language_models/openai_gpt.py
```diff
@@ -288,6 +306,9 @@ class OpenAIGPTConfig(LLMConfig):
     chat_model: str = defaultOpenAIChatModel
     completion_model: str = defaultOpenAICompletionModel
     run_on_first_use: Callable[[], None] = noop
+    parallel_tool_calls: Optional[bool] = None
+    supports_json_schema: Optional[bool] = None
+    supports_strict_tools: Optional[bool] = None
```
pchalasani, line 311: Please comment above this line on what `strict` does -- it is clear to me it is about strict decoding, but what happens if this is False? Does it mean:
- JSON will be produced but may not adhere exactly to the schema, OR
- JSON is not even guaranteed?

I'm thinking it's the former? Maybe make these config settings in the docs.

langroid/agent/chat_agent.py
```diff
@@ -145,10 +172,22 @@ def __init__(
...
         self.llm_functions_handled: Set[str] = set()
         self.llm_functions_usable: Set[str] = set()
         self.llm_function_force: Optional[Dict[str, str]] = None
 
+        self.output_format: Optional[type[ToolMessage | BaseModel]] = None
+
+        self.saved_requests_and_tool_setings = self._requests_and_tool_settings()
+        if config.output_format is not None:
+            self.set_output_format(config.output_format)
+        # OpenAI does not support all strict schemas. We disable for
```
pchalasani, line 185: Comment may need to be fixed? 
"controls whether to disable strict schemas for this agent if strict mode causes exception"

langroid/agent/chat_agent.py
```diff
@@ -145,10 +172,22 @@ def __init__(
...
+        # this agent if an exception is thrown in strict mode
+        self.disable_strict = False
+        self.any_strict = False
```
pchalasani, line 188: short comment on this variable

langroid/agent/chat_agent.py
```diff
@@ -145,10 +172,22 @@ def __init__(
...
+        self.disable_strict_tools_set: set[str] = set()
```
pchalasani, line 189: also comment on this

langroid/agent/chat_agent.py
```diff
@@ -145,10 +172,22 @@ def __init__(
         self.system_tool_format_instructions: str = ""
 
         self.llm_functions_map: Dict[str, LLMFunctionSpec] = {}
+        self.llm_functions_class_map: Dict[str, type[ToolMessage]] = {}
```
pchalasani, line 175: is this needed since we already have `self.llm_tools_map` (defined in `agent/base.py`) ? If it's needed, suggest adding short comment about it.

langroid/agent/chat_agent.py
```diff
@@ -581,6 +655,108 @@ def enable_message(
...
         self.system_tool_instructions = self.tool_instructions()
 
+    def _requests_and_tool_settings(self) -> tuple[Optional[set[str]], bool, bool]:
+        """
+        Returns the current set of enabled requests for inference and tools configs.
+        Used for restoring setings overriden by `set_output_format`.
+        """
+        return (
+            self.enabled_requests_for_inference,
+            self.config.use_functions_api,
```
pchalasani, line 665: should we include `self.config.use_tools_api`?

langroid/agent/chat_agent.py
```diff
@@ -581,6 +655,108 @@ def enable_message(
...
+                for t in [ToolMessage, BaseModel]
+            ):
+                output_type = get_pydantic_wrapper(output_type)
+
+            if self.output_format is None and force_tools:
+                self.saved_requests_and_tool_setings = (
+                    self._requests_and_tool_settings()
+                )
+
+            self.output_format = output_type
```
pchalasani, line 737: when we get to this point are we assuming `self.output_format` is `None`?
Otherwise, we'd be over-writing it, but maybe that is what is intended anyway.

langroid/agent/chat_agent.py
```diff
@@ -623,6 +799,139 @@ def disable_message_use_except(self, message_class: Type[ToolMessage]) -> None:
...
+                    name = tool_class.default_value("request")
+                    self.disable_strict_tools_set.add(name)
+                    logging.warning(
+                        f"""
+                        Validation error occured with strict tool format.
+                        Disabling strict mode for the {name} tool.
+                        """
+                    )
+                else:
+                    self.tool_error = True
```
pchalasani, line 862: short comment to explain this, e.g. what does `tool_error` signify?

langroid/agent/chat_agent.py
```diff
@@ -685,7 +1017,22 @@ def llm_response(
...
+                response = self.llm_response_messages(hist, output_len, tool_choice)
+            except openai.BadRequestError as e:
+                if self.any_strict:
+                    self.disable_strict = True
+                    self.set_output_format(None)
+                    logging.warning(
+                        f"""
+                        OpenAI BadRequestError raised with strict mode enabled.
+                        Message: {e.message}
+                        Disabling strict mode.
```
pchalasani, line 1030: "... and retrying"

(to make clear we are re-invoking the LLM response)

langroid/agent/chat_agent.py
```diff
@@ -716,9 +1090,24 @@ async def llm_response_async(
...
+                )
+            except openai.BadRequestError as e:
+                if self.any_strict:
+                    self.disable_strict = True
+                    self.set_output_format(None)
+                    logging.warning(
+                        f"""
+                        OpenAI BadRequestError raised with strict mode enabled.
+                        Message: {e.message}
+                        Disabling strict mode.
```
pchalasani, line 1105: "... and retrying "

langroid/agent/chat_agent.py
```diff
@@ -697,6 +1044,10 @@ def llm_response(
             if isinstance(message, str)
             else message.metadata.tool_ids if message is not None else []
         )
+
+        # If using strict output format, parse the output JSON
+        self._load_output_format(response)
```
pchalasani, line 1049: since `llm_response_messages` is already doing this, do we need this here? Or is there an execution path where this is needed?

langroid/agent/chat_agent.py
```diff
@@ -730,6 +1119,10 @@ async def llm_response_async(
...
```
pchalasani, line 1124: same comment as in sync case - since the `llm_response_messages_async` is doing it already, do we need it here?

tests/main/test_task.py
```diff
@@ -673,10 +676,14 @@ def polinsky_feedback(self, msg: FeedbackTool) -> str | DonePassTool:
     assert all(s in response.content for s in strings)
 
 
-def test_task_tool_responses():
+def test_task_tool_responses(
+    test_settings: Settings,
```
pchalasani, line 680: don't need this any more since `conftest.py` now has a fixture that auto-injects these

tests/main/test_task.py
```diff
@@ -673,10 +676,14 @@ def polinsky_feedback(self, msg: FeedbackTool) -> str | DonePassTool:
...
+):
     """Test that returning ToolMessage from an entity-responder or a Task.run() are
     handled correctly"""
 
+    set_global(test_settings)
```
pchalasani, line 685: not needed (see above comment)

tests/main/test_task.py
```diff
@@ -779,3 +786,192 @@ def llm_response(
 
     result = processor_task[int].run(10)
     assert result == 11
+
+
+def test_task_output_format_sequence(
+    test_settings: Settings,
```
pchalasani, line 792: `test_settings` not needed

tests/main/test_task.py
```diff
@@ -779,3 +786,192 @@ def llm_response(
...
+):
+    """
+    Test that `Task`s correctly execute a sequence of steps
+    controlled by the agent's `output_format`, and that `output_format`
+    is handled by default without `enable_message`.
+    """
+
+    set_global(test_settings)
```
pchalasani, line 800: not needed

tests/main/test_task.py
```diff
@@ -779,3 +786,192 @@ def llm_response(
...
+
+            return str(message.a * message.b)
+
+        def increment(self, message: IncrementTool) -> str:
+            self.set_output_format(PowerTool)
+
+            return str(message.x + 1)
+
+        def power(self, message: PowerTool) -> str:
+            self.set_output_format(MultiplyTool)
```
pchalasani, line 835: we don't need to set this I think? since power is the last step

langroid/agent/chat_agent.py
```diff
@@ -623,6 +799,139 @@ def disable_message_use_except(self, message_class: Type[ToolMessage]) -> None:
             self.llm_tools_usable.discard(r)
             self.llm_functions_usable.discard(r)
 
+    def _load_output_format(self, message: ChatDocument) -> None:
+        """
+        If set, attempts to parse a value of type `self.output_format` from the message
+        contents and assigns it to `content_any`.
+        """
+        if self.output_format is not None:
```
pchalasani, line 807: This `_load_output_format` ends up getting called twice - once in `llm_response_messages` and then at the end of `llm_response` - this works fine, but it does some (minor) un-necessary work by trying to populate `content_any` all over again. I wonder if we should check if it already exists, and skip, or maybe just keep it as is just to be safe?

tests/main/test_tool_messages.py
```diff
@@ -1187,6 +1197,410 @@ def agent_response(self, msg: str | ChatDocument) -> Any:
...
+                ],
+            )
+
+        return ChatDocument(
+            content="",
+            metadata=ChatDocMetaData(sender=Entity.LLM),
+            function_call=attempt,
+        )
+
+    assert (
```
pchalasani, line 1309: in each attempt below, please add short comment on what makes that attempt wrong (e.g. wrong name, wrong args, etc)

tests/main/test_tool_messages.py
```diff
@@ -1187,6 +1197,410 @@ def agent_response(self, msg: str | ChatDocument) -> Any:
...
+                    LLMMessage(
+                        role=Role.TOOL,
+                        tool_call_id=t.id,
+                        content="error",
+                    )
+                    for t in attempt.oai_tool_calls
+                ]
+            )
+
+        agent.handle_message(attempt)
```
pchalasani, line 1267: comment:
"Simulates bad tool attempt by LLM"

tests/main/test_tool_messages.py
```diff
@@ -1187,6 +1197,410 @@ def agent_response(self, msg: str | ChatDocument) -> Any:
...
+    class BrokenStrictSchemaAgent(ChatAgent):
+        def _function_args(self) -> tuple[
+            Optional[List[LLMFunctionSpec]],
+            str | dict[str, str],
+            Optional[list[OpenAIToolSpec]],
+            Optional[dict[str, dict[str, str] | str]],
+            Optional[OpenAIJsonSchemaSpec],
+        ]:
+            functions, fun_call, tools, force_tool, output_format = (
+                super()._function_args()
```
pchalasani, line 1402: If I understood correctly, this is implementing a "broken" version of the correct `_function_args()` -- instead of this, I wonder if there's a more direct way to do the test, e.g. manually set various values?

pchalasani: Or maybe add comments to make clear exactly what is broken in this implementation

tests/main/test_tool_messages.py
```diff
@@ -1187,6 +1197,410 @@ def agent_response(self, msg: str | ChatDocument) -> Any:
...
+@pytest.mark.parametrize("use_fn_api", [True, False])
+@pytest.mark.parametrize("use_tools_api", [True, False])
+@pytest.mark.parametrize("parallel_tool_calls", [True, False])
+def test_strict_schema_mismatch(
+    use_fn_api: bool,
+    use_tools_api: bool,
+    parallel_tool_calls: bool,
+):
+    """
+    Test that validation errors triggered in strict result in disabled strict ouput.
```
pchalasani, line 1503: ouput -> output

tests/main/test_tool_messages_async.py
pchalasani, line 1: same comments as in the sync version

pchalasani (changes requested):
Thanks @nilspalumbo this is fantastic! Especially the `AnyTool` mechanism. You can ignore the comments about not needing `test_settings` (`conftest.py` now auto-inserts them, but many tests already have it, so no need to remove them if you don't want to).

langroid/language_models/openai_gpt.py
nilspalumbo: Doesn't parallel_tool_calls default to `True`? I think we need to catch the `None` case

langroid/language_models/openai_gpt.py
nilspalumbo: Added comments on what this does; however, I'm not positive whether OpenAI uses basic JSON mode for tool calls without strict. Failures to produce JSON are at least rare

langroid/agent/chat_agent.py
nilspalumbo: No need, as this isn't overridden (since we always disable `use_functions_api` which disables `use_tools_api` when we enable an output format).

langroid/agent/chat_agent.py
nilspalumbo: We are overriding it; we only consider whether it is `None` when we decide whether to save/restore tools settings.

langroid/agent/chat_agent.py
nilspalumbo: Added comments; in strict mode this means there was some schema (or perhaps token limit) error and we should disable, but in non-strict mode we can trigger strict recovery and correct.

langroid/agent/chat_agent.py
nilspalumbo: Removed (similarly in the async case)

tests/main/test_task.py
nilspalumbo: I was thinking that we should set up a cycle to allow agent reuse; this isn't needed here, so I'll remove.

tests/main/test_tool_messages.py
nilspalumbo: Hmm maybe; we could modify `ChatAgent` to have a configurable schema editing function if that's better. Added a comment.

langroid/agent/task.py
nilspalumbo: This should be using `llm_response_async`; corrected (may have overwritten the line after some changes to the sync path)

langroid/language_models/openai_gpt.py
pchalasani: you're right!

pchalasani:
@nilspalumbo I pushed a change this PR branch updating gpt4o to the latest 2024-11-20 version, maybe it will get rid of the test failures

pchalasani:
Also the fallback on gpt-4 in pytest is a bit archaic at this point, since gpt-4 is ancient and has a puny 8k context.

pchalasani (changes requested):
@nilspalumbo thanks for addressing the comments. I think there's a couple left (updating gpt-4o version name in the comments, and possibly something else)

langroid/agent/chat_agent.py
pchalasani: maybe mark this as a TODO in the code to visit later.

langroid/language_models/azure_openai.py
pchalasani: added this

langroid/language_models/azure_openai.py
pchalasani: added

langroid/language_models/openai_gpt.py
pchalasani: skip since there are a few possibilities

pchalasani:
@nilspalumbo I resolved the remaining minor comments. Let me know if this is good to merge.

nilspalumbo:
@pchalasani Thank you! I think so (except one potential remaining thing: should we automatically enable/disable `ToolMessages` when changing the output format to ensure that the instructions are generated?) If we want to do that, we could also automatically modify the system prompt for non-`ToolMessage` `output_format`s in a similar way.

pchalasani:
> @pchalasani Thank you! I think so (except one potential remaining thing: should we automatically enable/disable `ToolMessages` when changing the output format to ensure that the instructions are generated?) If we want to do that, we could also automatically modify the system prompt for non-`ToolMessage` `output_format`s in a similar way.

Yes I think that would be good to have. If it's relatively straightforward, we can include in this this PR, else next patch.

pchalasani:
> > @pchalasani Thank you! I think so (except one potential remaining thing: should we automatically enable/disable `ToolMessages` when changing the output format to ensure that the instructions are generated?) If we want to do that, we could also automatically modify the system prompt for non-`ToolMessage` `output_format`s in a similar way.
> 
> Yes I think that would be good to have. If it's relatively straightforward, we can include in this this PR, else next patch.

in either case it may be good to have this behavior configurable

nilspalumbo:
I should have a version done soon, but, as we need to restore on `output_format` change, and some of the internal `output_format` mechanisms will need to change (e.g. we won't need to have `all_llm_tools_handled` if we are operating via `enable_message`) some additional testing will be needed; maybe we should add this separately.

pchalasani:
> I should have a version done soon, but, as we need to restore on `output_format` change, and some of the internal `output_format` mechanisms will need to change (e.g. we won't need to have `all_llm_tools_handled` if we are operating via `enable_message`) some additional testing will be needed; maybe we should add this separately.

Ok
We can wait for next PR

PR #598, opened 2024-10-25
Fix broken assistants update call.

vattay:
The current OpenAI assistants API (`openai.resources.beta.assistants.update`) does not support having an existing function in the functions list, it fails with a "Function must have unique names error". Since `enable_message` will be recursively called for lists, or sequentially called when the user adds multiple tools in sequence, this will always fail with more than one tool. Instead, just add the last tool as a function. 

This does seem like a strange implementation from OpenAI as the `update` is really an append only operation for functions then.


pchalasani:
@vattay thanks for this. Would you be able to add a small test in `tests/main/test_openai_assistant.py` that would have failed prior to this fix, but succeeds now?

PR #601, opened 2024-11-06
feat: add support for Azure OpenAI Embeddings

### LINKED ISSUE #514: Add Azure embeddings
Implementation should be similar to the OpenAI embeddings

AbhishekRP2002:
- this PR aims to close #514 serving as the initial draft
- implements minor fixes ( changing the `ruff` command )


pchalasani:
Thanks @AbhishekRP2002 
please add suitable unit tests in `tests/main/test_embeddings.py`
I'll have a closer look after that.

langroid/embedding_models/models.py
```diff
@@ -23,6 +25,22 @@ class OpenAIEmbeddingsConfig(EmbeddingModelsConfig):
     context_length: int = 8192
 
 
+class AzureOpenAIEmbeddingsConfig(EmbeddingModelsConfig):
+    model_type: str = "azure-openai"
+    model_name: str = "text-embedding-ada-002"
+    api_key: str = ""
+    azure_endpoint: str = ""
```
alexagr, line 32: can you please change it to "api_base" - to align with the AzureConfig class?

langroid/embedding_models/models.py
```diff
@@ -23,6 +25,22 @@ class OpenAIEmbeddingsConfig(EmbeddingModelsConfig):
...
+    azure_deployment: Optional[str] = None
```
alexagr, line 33: can you please change it to "deployment_name" - to align with the AzureConfig class?

langroid/embedding_models/models.py
```diff
@@ -139,6 +171,72 @@ def embedding_dims(self) -> int:
...
+        Args:
+            config: Configuration for Azure OpenAI embeddings model.
+        Raises:
+            ValueError: If required Azure config values are not set.
+        """
+        super().__init__()
+        self.config = config
+        load_dotenv()
+
+        self.config.api_key = os.getenv("AZURE_OPENAI_API_KEY", "")
```
alexagr, line 194: can you please change it to:
```
    if not self.config.api_key:
        self.config.api_key = os.getenv("AZURE_OPENAI_API_KEY", "")
```
so that if user passes api_key in Config object it's not overwritten?

langroid/embedding_models/models.py
```diff
@@ -139,6 +171,72 @@ def embedding_dims(self) -> int:
...
+        if self.config.api_key == "":
+            raise ValueError(
+                """AZURE_OPENAI_API_KEY env variable must be set to use 
+            AzureOpenAIEmbeddings. Please set the AZURE_OPENAI_API_KEY value 
+            in your .env file."""
+            )
+        self.config.azure_endpoint = os.getenv("AZURE_OPENAI_ENDPOINT", "")
```
alexagr, line 201: can you please change it to:
````
        if not self.config.azure_endpoint:
            self.config.azure_endpoint = os.getenv("AZURE_OPENAI_ENDPOINT", "")
````
so that if user passes azure_endpoint in Config object it's not overwritten?

**Review by alexagr — CHANGES_REQUESTED** [review]
@AbhishekRP2002, thanks a lot for this PR - I really need this new functionality! )))
Just a few minor fixes, if you don't mind, as per my comments.

File: langroid/embedding_models/models.py (hunk 1 above)
Comment (AbhishekRP2002) [reply langroid/embedding_models/models.py:32]: the choice of args naming comes from the original Python SDK.
cc: @pchalasani 
ref: https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/embeddings?tabs=python-new

**Comment by AbhishekRP2002** [comment]
hi @pchalasani  i have add a unit test func for testing the azure openai embedding models, and have made a minor change in the boolean value representation in the toml file.

**Review by pchalasani — CHANGES_REQUESTED** [review]
Thanks @AbhishekRP2002 for this. I noticed this PR includes a bunch of files that should not be git-tracked, like `.qdrantdb/`. This likely happened because you used `git add` to add all files (or I think editors like VSCode do this automatically). Please make sure to only add relevant files. Also, we also don't want to be adding all such ignorable files into `.gitignore`, since we cannot anticipate all future ignore patterns.

**Comment by AbhishekRP2002** [comment]
> Thanks @AbhishekRP2002 for this. I noticed this PR includes a bunch of files that should not be git-tracked, like `.qdrantdb/`. This likely happened because you used `git add` to add all files (or I think editors like VSCode do this automatically). Please make sure to only add relevant files. Also, we also don't want to be adding all such ignorable files into `.gitignore`, since we cannot anticipate all future ignore patterns.

Thanks for the heads up @pchalasani  I have removed the ignorable unwanted dir / files from git tracking. Pls lemme know if there's anything else I need to adhere or make changes in.

**Comment by pchalasani** [comment]
@AbhishekRP2002 @alexagr there are common params between the azure embedding config and azure openai (AzureConfig) I think we should factor out the common params into `AzureConfig`, and then have `AzureOpenAILLMConfig` and `AzureOpenAIEmbeddingConfig` so each of these will have a field `azure_config: AzureConfig`. This way we avoid duplication of the configs, and also leverage the pydantic mechanism that gets you the configs from the env via the `env_prefix="AZURE_OPENAI_"`

what do you think?

**Comment by pchalasani** [comment]
> @AbhishekRP2002 @alexagr there are common params between the azure embedding config and azure openai (AzureConfig) I think we should factor out the common params into `AzureConfig`, and then have `AzureOpenAILLMConfig` and `AzureOpenAIEmbeddingConfig` so each of these will have a field `azure_config: AzureConfig`. This way we avoid duplication of the configs, and also leverage the pydantic mechanism that gets you the configs from the env via the `env_prefix="AZURE_OPENAI_"`
> 
> what do you think?

Actually never mind, let's not change AzureConfig or refactor it, as it will be a breaking change. 
For now, in `models.py` let's get the env vars with the right names (for those that are common with AzureConfig)

File: langroid/embedding_models/models.py (hunk 1 above)
Comment (pchalasani) [reply langroid/embedding_models/models.py:32]: I agree with @alexagr to keep the names consistent with AzureConfig

File: langroid/embedding_models/models.py (hunk 2 above)
Comment (pchalasani) [reply langroid/embedding_models/models.py:33]: Same here, let's align with AzureConfig's existing names

File: langroid/embedding_models/models.py [hunk 5]
```diff
@@ -139,6 +171,74 @@ def embedding_dims(self) -> int:
...
+        if not self.config.api_key:
+            self.config.api_key = os.getenv("AZURE_OPENAI_API_KEY", "")
+        if self.config.api_key == "":
+            raise ValueError(
+                """AZURE_OPENAI_API_KEY env variable must be set to use 
+            AzureOpenAIEmbeddings. Please set the AZURE_OPENAI_API_KEY value 
+            in your .env file."""
+            )
+        if not self.config.azure_endpoint:
+            self.config.azure_endpoint = os.getenv("AZURE_OPENAI_ENDPOINT", "")
```
pchalasani, line 203: use AZURE_OPENAI_API_BASE

langroid/embedding_models/models.py
```diff
@@ -139,6 +171,74 @@ def embedding_dims(self) -> int:
...
+        if self.config.azure_endpoint == "":
+            raise ValueError(
+                """AZURE_OPENAI_ENDPOINT env variable must be set to use
```
pchalasani, line 206: same, fix comment accordingly

langroid/embedding_models/models.py
```diff
@@ -139,6 +171,74 @@ def embedding_dims(self) -> int:
...
+        Args:
+            config: Configuration for Azure OpenAI embeddings model.
+        Raises:
+            ValueError: If required Azure config values are not set.
+        """
+        super().__init__()
+        self.config = config
+        load_dotenv()
+
+        if not self.config.api_key:
```
pchalasani, line 194: simplify using `x = x or ...` pattern

langroid/embedding_models/models.py
```diff
@@ -139,6 +171,74 @@ def embedding_dims(self) -> int:
...
+            self.config.api_key = os.getenv("AZURE_OPENAI_API_KEY", "")
+        if self.config.api_key == "":
+            raise ValueError(
+                """AZURE_OPENAI_API_KEY env variable must be set to use 
+            AzureOpenAIEmbeddings. Please set the AZURE_OPENAI_API_KEY value 
+            in your .env file."""
+            )
+        if not self.config.azure_endpoint:
```
pchalasani, line 202: simplify using `x = x or ...` pattern

AbhishekRP2002: Done @pchalasani

AbhishekRP2002:
hi @pchalasani  any updates on this ?

pchalasani:
> hi @pchalasani any updates on this ?

This PR is also including the exa changes, we should keep it separate.

langroid/embedding_models/models.py
pchalasani: @AbhishekRP2002 please make this `api_base` to align with `AzureCongig`

langroid/embedding_models/models.py
pchalasani: please change name as suggested

PR #607, opened 2024-11-13
Fix neo4j schema flag

Mohannadcse:
`init_state()` is called twice inside `neo4j_chat_agent`.
the flag `tried_schema` isn't constantly set 

@pchalasani: doesn't seem `addressing_prefix=SEND_TO` works correctly because the agent in `dependency_chatbot.py` doesn't address the user.

### DIFF
--- a/examples/kg-chat/dependency_chatbot.py
+++ b/examples/kg-chat/dependency_chatbot.py
@@ -269,7 +269,9 @@ def main(
             use_tools=tools,
             use_functions_api=not tools,
             addressing_prefix=SEND_TO,
-            llm=lm.azure_openai.AzureConfig(),
+            llm=lm.OpenAIGPTConfig(
+                chat_model=model or lm.OpenAIChatModel.GPT4o,
+            ),
         ),
     )
 
--- a/examples/kg-chat/dependency_chatbot.py
+++ b/examples/kg-chat/dependency_chatbot.py
@@ -43,7 +43,10 @@ from langroid.agent.special.neo4j.neo4j_chat_agent import (
     Neo4jChatAgentConfig,
     Neo4jSettings,
 )
-
+from langroid.agent.special.neo4j.tools import (
+    cypher_retrieval_tool_name,
+    graph_schema_tool_name,
+)
 from langroid.utils.constants import NO_ANSWER, SEND_TO
 from langroid.utils.configuration import set_global, Settings
 from langroid.agent.tool_message import ToolMessage
@@ -53,6 +56,8 @@ from langroid.agent.task import Task
 from cypher_message import CONSTRUCT_DEPENDENCY_GRAPH
 
 app = typer.Typer()
+
+web_search_name = GoogleSearchTool.default_value("request")
 
 
 class DepGraphTool(ToolMessage):
@@ -69,6 +74,9 @@ class DepGraphTool(ToolMessage):
     package_name: str
 
 
+construct_dependency_graph_name = DepGraphTool.default_value("request")
+
+
 class VisualizeGraph(ToolMessage):
     request = "visualize_dependency_graph"
     purpose = """
@@ -78,6 +86,9 @@ class VisualizeGraph(ToolMessage):
     package_type: str
     package_name: str
     query: str
+
+
+visualize_dependency_graph_name = VisualizeGraph.default_value("request")
 
 
 class DependencyGraphAgent(Neo4jChatAgent):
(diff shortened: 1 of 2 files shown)


examples/kg-chat/dependency_chatbot.py
pchalasani, line 1: @Mohannadcse this script gave an err when I ran it --
```
ConnectionError: Failed to initialize Neo4j connection: Cannot resolve address 927d9aab.databases.neo4j.io:7687
```

It should be more straightforward to run it, without having to set up anything. Can you see if you can modify it so that it uses a docker-based set up like in `test_neo4j_chat_agent.py`?

langroid/agent/special/neo4j/neo4j_chat_agent.py
```diff
@@ -87,12 +87,13 @@ def __init__(self, config: Neo4jChatAgentConfig):
         self._import_neo4j()
         self._initialize_db()
         self._init_tools_sys_message()
-        self.init_state()
+        # self.init_state() # already called inside _init_tools_sys_message()
```
pchalasani, line 90: we should leave it in here, since we may call `init_state()` to reset state after the agent has been initialized.

langroid/agent/special/neo4j/neo4j_chat_agent.py
```diff
@@ -87,12 +87,13 @@ def __init__(self, config: Neo4jChatAgentConfig):
...
 
     def init_state(self) -> None:
         super().init_state()
         self.current_retrieval_cypher_query: str = ""
-        self.tried_schema: bool = False
+        if not self.tried_schema:
```
pchalasani, line 95: this is superfluous

langroid/agent/special/neo4j/neo4j_chat_agent.py
```diff
@@ -398,7 +400,8 @@ def graph_schema_tool(
 
     def _init_tools_sys_message(self) -> None:
         """Initialize message tools used for chatting."""
-        self.tried_schema = False
+        if not self.tried_schema:
```
pchalasani, line 403: superfluous

examples/kg-chat/dependency_chatbot.py
Mohannadcse: Seems the Neo4j instance is down. 
We already have docker instructions [here](https://github.com/langroid/langroid/tree/main/examples/kg-chat#requirements) 
The docker based setup gave me errors, because the newer docker version doesn't support the command `docker-compose`, it's replaced in the new docker version with `docker compose`. So I had to make some modifications to get neo4j docker working.

langroid/agent/special/neo4j/neo4j_chat_agent.py
Mohannadcse: I had a case where DB exists, so `tried_schema` will be true after calling `graph_schema_tool` from `_initialize_db`. But when `_init_tools_sys_message` is called, which is immediately after `_init_tools_sys_message`, it sets `tried_schema` to false at [this line](https://github.com/langroid/langroid/blob/main/langroid/agent/special/neo4j/neo4j_chat_agent.py#L401)

langroid/agent/special/neo4j/neo4j_chat_agent.py
Mohannadcse: yes, I should have removed this after commenting out `self.init_state()`

langroid/agent/special/neo4j/neo4j_chat_agent.py
pchalasani: but the code is superfluous - if it's false, you're setting it to false, which is superfluous

examples/kg-chat/dependency_chatbot.py
pchalasani: there should be a way to have it work without having to set up/update local env. The test works in github-actions and locally for me, so not sure what the problem is.

Mohannadcse: Please see this Docker announce ment of docker compose V2 https://www.docker.com/blog/announcing-compose-v2-general-availability/

Mohannadcse: > there should be a way to have it work without having to set up/update local env

To ensure I understood correctly, is the plan now to avoid using local env for neo4j?

pchalasani: I meant, we should make the tests/examples work without requiring people to have specific versions of docker on their local machine. In the docker v2 link, it shows what needs to be done.

pchalasani:
> doesn't seem `addressing_prefix=SEND_TO` works correctly

`addressing_prefix` needs to be set in the agent config as well. See the other example scripts. I did not update this specific dependency-chatbot script.

examples/kg-chat/dependency_chatbot.py
pchalasani, line 1: the name `retrieval_query` needs to be changed, to derive from the actual tool. The tool names have changed, and they should not be hard-coded in the prompts. See the `chat-neo4j.py` example.

Mohannadcse:
> > doesn't seem `addressing_prefix=SEND_TO` works correctly
> 
> `addressing_prefix` needs to be set in the agent config as well. See the other example scripts. I did not update this specific dependency-chatbot script.

Thanks for the tip, I did try previously and didn't work, that's why I didn't push this change.
BTW, `ArangoDB` chat examples don't contain this setting `addressing_prefix` under the agent config, it's only set under task config. But I noticed you have incorporated some insturctions/implementations under `llm_response` in the `arangodb_agent`

```
dependency_agent = DependencyGraphAgent(
        config=Neo4jChatAgentConfig(
            chat_mode=True,
            neo4j_settings=neo4j_settings,
            show_stats=False,
            use_tools=tools,
            use_functions_api=not tools,
            llm=lm.OpenAIGPTConfig(
                chat_model=model or lm.OpenAIChatModel.GPT4o,
            ),
            addressing_prefix=SEND_TO,
        ),
    )
```

Mohannadcse:
@pchalasani addressed your comments and suggestions. Pls let me know if I missed anything.

pchalasani (changes requested):
@Mohannadcse what is the goal of this PR?
in main, the `test_neo4j_chat_agent.py passes`, and the example `chat-neo4j.py` works, but the `dependency_chatbot.py` fails, partly because I did not update it after the changes to `Neo4jChatAgent`. In your PR, this example still fails due to a neo4j setup issue. It will be good to get this example working using an automatic docker setup as in the neo4j test and chat example.

Mohannadcse:
> @Mohannadcse what is the goal of this PR?

Fix the problem that I mentioned yes

langroid/agent/special/neo4j/neo4j_chat_agent.py
```diff
@@ -87,12 +87,12 @@ def __init__(self, config: Neo4jChatAgentConfig):
...
         self._initialize_db()
         self._init_tools_sys_message()
-        self.init_state()
+        # self.init_state() # already called inside _init_tools_sys_message()
 
     def init_state(self) -> None:
         super().init_state()
         self.current_retrieval_cypher_query: str = ""
-        self.tried_schema: bool = False
+        self.tried_schema: bool = self.tried_schema or False
```
pchalasani, line 95: this is superfluous, it doesn't do anything

langroid/agent/special/neo4j/neo4j_chat_agent.py
```diff
@@ -398,7 +399,7 @@ def graph_schema_tool(
 
     def _init_tools_sys_message(self) -> None:
         """Initialize message tools used for chatting."""
-        self.tried_schema = False
+        self.tried_schema = self.tried_schema or False
```
pchalasani, line 402: superfluous

Mohannadcse: this prevents `self.tried_schema` from being false when it's already set to true after calling `_initialize_db` and there is a database exists

PR #610, opened 2024-11-16
"true" async implementation of user_response_async() and agent_response_async()

alexagr:
"True" asynchronous implementation of `user_response_async` and `agent_response_async` methods in `Agent`.
Most of the common code between sync and async implementation is refactored into private methods (see, for example, `_handle_message_final`).
Asynchronous tool handlers may be implemented via `{tool.name}_async` methods in Agent class/instance.
Implicitly fallback to synchronous callbacks/handlers, if asynchronous ones are not implemented.


pchalasani:
Thanks @alexagr will look in detail. Will be good to add tests that exercise these changes.

alexagr:
@pchalasani modified test_tool_messages_async.py to use handle_message_async() and made one of the tools use the new _async handler

pchalasani:
@nilspalumbo could you please take a look

langroid/agent/base.py
```diff
@@ -690,6 +691,60 @@ def user_response(
...
+        self,
+        msg: Optional[str | ChatDocument] = None,
+    ) -> Optional[ChatDocument]:
+        """
+        Asynch version of `user_response`. See there for details.
+        """
+        if not self.user_can_respond(msg):
+            return None
+
+        if self.callbacks.get_user_response_async is not None:
```
nilspalumbo, line 704: As in the synchronous path, we need to skip when `default_user_response` is set.

langroid/agent/base.py
```diff
@@ -649,23 +657,16 @@ def user_response(
...
-                user_msg = Prompt.ask(
-                    f"[blue]{self.indent}"
-                    + self.config.human_prompt
-                    + f"\n{self.indent}"
-                ).strip()
+            return False
+
+        return True
 
+    def _user_response_final(
```
nilspalumbo, line 664: The original code includes a line `user_msg = user_msg.strip() or default_user_msg.strip()`; dropping it here means that we do not ensure that a not-None response is returned on `need_human_response`. @pchalasani 
could this cause problems?

langroid/agent/base.py
```diff
@@ -649,23 +657,16 @@ def user_response(
         interactive = self.interactive or settings.interactive
```
nilspalumbo: Diffs seem confused here; this line is new. I do not expect that we want to force-enable interactive mode whenever `settings.interactive` is set; we may allow it to set defaults but this would force even agents with explicitly-disabled interactivity to be interactive. 

@pchalasani I don't see any existing use of `settings.interactive`, where do we use it? If we don't use it anywhere, we could change the type to `Optional[bool]` and only override if set.

langroid/agent/base.py
```diff
@@ -690,6 +691,60 @@ def user_response(
...
+        """
+        Asynch version of `user_response`. See there for details.
+        """
+        if not self.user_can_respond(msg):
+            return None
+
+        if self.callbacks.get_user_response_async is not None:
+            user_msg = await self.callbacks.get_user_response_async(prompt="")
+        else:
+            user_msg = await self.callbacks.get_user_response(prompt="")
```
nilspalumbo, line 724: We need to consider the condition when no sync callback is set. If it isn't, we must block on synchronous user input as in the sync `user_response` method.

In addition, no `await` should be used here as that callback should not be an async function.

langroid/agent/base.py
pchalasani: @nilspalumbo yes that is why I simplified that line, to not use `settings.interactive` -- there was a conflict I just resolved, it should no longer be here

langroid/agent/base.py
```diff
@@ -1109,65 +1164,18 @@ def tool_validation_error(self, ve: ValidationError) -> str:
...
+        There were one or more errors in your attempt to use the
+        TOOL or function_call named '{tool_name}':
         {bad_field_errors}
         Please write your message again, correcting the errors.
         """
 
-    def handle_message(
-        self, msg: str | ChatDocument
-    ) -> None | str | OrderedDict[str, str] | ChatDocument:
+    def _check_multiple_orch_tools(
```
nilspalumbo, line 1204: Perhaps call it `_get_multiple_orch_tool_errs` or similar given the return type.

langroid/agent/base.py
pchalasani: yes the diff-viewer in pycharm is getting confused

langroid/agent/base.py
pchalasani: The original code did not have a `user_can_respond` method; the PR has it, and it's setting `user_msg` but it's not being used anywhere. We need a closer look at what is going on.

nilspalumbo:
Thanks, this will be a nice update! In addition to the comments, we need a few more things: we need additional testing, both of (simulated) async user responses and additional tests of async tools (in particular, we need a test with use in a `Task`).

In addition, we need to have async handlers automatically generated in `_get_tool_list()` when defined on a `ToolMessage`; however, I can add that from another branch.

langroid/agent/base.py
nilspalumbo: Sounds good, thanks! I had to resort to running `diff` manually on blocks of code to see the changes

langroid/agent/base.py
```diff
@@ -1511,6 +1643,37 @@ def _maybe_truncate_result(
             ) + truncate_warning
             return result
 
+    async def handle_tool_message_async(
```
pchalasani, line 1646: This method is not awaiting anything and seems to be just a copy of `handle_tool_message` -- do we need this?

pchalasani: never mind, missed that it *is* awaiting handler_method. Pls ignore

langroid/agent/base.py
nilspalumbo: Yeah, it mainly consists of some extracted code from `user_response` to avoid code duplication (it's `need_human_response or interactive`). But that does mean that if we need to ensure not-None on `need_human_response` we can't use it directly.

pchalasani:
@nilspalumbo does this intersect much with your PR ? In any case, we would merge yours first.

nilspalumbo:
@pchalasani Added async handler generation to `_get_tool_list()`. Also, it seems that some of the code was mangled by the merge (see the new `user_can_respond()`, we may have to do a manual merge

nilspalumbo:
> @nilspalumbo does this intersect much with your PR ? In any case, we would merge yours first.

Not much, but it will cause merge conflicts, in particular because I modified `handle_message` etc. to track failures in order to trigger strict recovery.

langroid/agent/base.py
alexagr: agreed - changing

alexagr:
> Thanks, this will be a nice update! In addition to the comments, we need a few more things: we need additional testing, both of (simulated) async user responses and additional tests of async tools (in particular, we need a test with use in a `Task`).

@nilspalumbo I've already made changes to test_tool_messages_async.py - so that it tests a tool with both sync and async handlers in an async Task. What else do you need? Can you please point me to some existing sync test that you want me to modify / create an async version of?

langroid/agent/base.py
alexagr: Ok - I see that the confusion comes from the automatic merge that broke a few things too.
In my latest commit I fixed these merger errors and added a slightly refactored version of the missing line (from #611) to the beginning of _user_response_final() method.

alexagr:
@nilspalumbo @pchalasani A new commit addresses all issues that were raised and fixes bugs introduced by the automatic merge with "main".

nilspalumbo:
Thanks! Looks good; a good test would be a variant of `test_task_batch` in `test_batch.py` in which the LLM is prompted to make a tool call with the provided argument (along the lines of `test_task_gen_batch_exceptions`). The handler should use `asyncio.sleep` to simulate computation; the test should ensure that the fastest of the tasks finishes first to confirm that we are not blocking.

For async user responses, set a constant callback as `callbacks.get_user_response_async` and another sync callback and ensure that the async callback is being used (as in `test_task_default_human_response` in `test_task.py`); an additional test in `test_batch.py` which validates that no blocking is occuring as above would be a good addition.

pchalasani:
@alexagr thanks for this PR. We will first merge #589 after @nilspalumbo addresses my review, and then there will be conflicts we would have to resolve.

pchalasani:
@nilspalumbo please let me know once you're done reviewing

nilspalumbo:
@alexagr thanks for adding the tests! They look good, and should be enough after a few updates that I've added.

@pchalasani I'm done now, and will start Pytest.

pchalasani:
> @pchalasani is there any chance the slow test is purely due to caching? Else seems that the update may be causing some problems.

I think the tests are failing because when we run on someone's fork, the gh secrets are not shared. So to run tests, one of the core devs needs to push to a new branch and run them. This is what I've done for other PRs.

nilspalumbo:
@pchalasani I've done that, should I make a duplicate PR or can this PR be associated with the new branch?

pchalasani:
> should I make a duplicate PR

duplicate PR from your branch

nilspalumbo:
Sounds good; I've also added a manual request option to the Pytest workflow, should I drop that?

pchalasani:
> Sounds good; I've also added a manual request option to the Pytest workflow, should I drop that?

probably yes

nilspalumbo:
OK; the new PR is #630.

nilspalumbo (approved):
@pchalasani tests are passing on #630, I think it's ready to merge.

Thanks @alexagr, this'll be a nice update!

nilspalumbo:
Closed, merged in #630.

PR #615, opened 2024-11-18
feat(docagent): hypothetical questions

dfm88:
### Hypothetical Question Index

Despite there being no papers supporting this technique, Hypothetical Question Indexing is mentioned in several blogs as an advanced RAG technique. It somewhat contrasts with Hypothetical Answer (HyDE) and intervenes during the embedding saving phase, specifically trying to solve the problem where the question posed on an embedding contains words distant from it. Therefore, it associates different questions to increase the probability of retrieval.

Some link:
- https://www.falkordb.com/blog/advanced-rag/
- https://youtu.be/sGvXO7CVwc0?t=158


langroid/agent/special/doc_chat_agent.py
```diff
@@ -405,6 +419,10 @@ def ingest_docs(
         if self.vecdb is None:
             raise ValueError("VecDB not set")
 
+        if self.config.hypothetical_questions:
+            question_docs = [self.llm_hypothetical_question(doc) for doc in docs]
```
pchalasani, line 423: This will be very slow when there are hundreds/thousands of chunks. We should use an appropriate batch method, see `langroid/agent/batch.py`

langroid/agent/special/doc_chat_agent.py
```diff
@@ -126,6 +126,20 @@ class DocChatAgentConfig(ChatAgentConfig):
     # https://arxiv.org/pdf/2212.10496.pdf
     # It is False by default; its benefits depends on the context.
     hypothetical_answer: bool = False
+    # Flag to enable Hypothetical Question Indexing,
+    # which uses a language model to generate questions for each data chunk,
+    # this might help in finding relevant chunks for a given question.
+    hypothetical_questions: bool = False
+    hypothetical_questions_nr: int = 3
```
pchalasani, line 133: => `num_hypothetical_questions`

langroid/agent/special/doc_chat_agent.py
```diff
@@ -126,6 +126,20 @@ class DocChatAgentConfig(ChatAgentConfig):
...
+    hypothetical_questions_default_prompt: str = f"""
```
pchalasani, line 134: => `hypothetical_questions_prompt`

langroid/agent/special/doc_chat_agent.py
```diff
@@ -860,6 +878,26 @@ def llm_hypothetical_answer(self, query: str) -> str:
...
+            raise ValueError("LLM not set")
+
+        with status("[cyan]LLM generating hypothetical questions..."):
+            with StreamingIfAllowed(self.llm, False):
+                prompt = self.config.hypothetical_questions_default_prompt % {
+                    "passage": chunk.content
+                }
+            questions = self.llm_response_forget(prompt).content
+        return Document(
+            content=questions,
```
pchalasani, line 893: The content is just the questions, which is fine since it helps match the query, but we lose the original content. I know you are keeping it in the metadata, but `original_content` is not used in the normal flow of this agent anywhere. 

I think it's better to _augment_ the `content` field value with the `questions`, and also demarcate those questions in some special way, so that during passage retrieval, the questions are ignored. I.e. only use the questions for matching/ranking, but ignore in retrieval.

pchalasani: The other advantage of augmenting (i.e. combining chunk content + questions) is that now we can match with both the question and content, rather than having to deal with de-duplication, e.g. when we find a match for questions_5 that links to chunk_3 and also find a match to chunk_3.

pchalasani:
Thanks for this @dfm88 
Please also add suitable tests in `test_doc_chat_agent.py`

langroid/agent/chat_agent.py
dfm88, line 1: Here I had to change the signature of 
`llm_response_forget` and `llm_response_forget_async` (which might also have sense) because I had MyPy problem in passing the `llm_response_forget_async` method to `run_batch_agent_method`

langroid/agent/special/doc_chat_agent.py
dfm88: done

langroid/agent/special/doc_chat_agent.py
dfm88: done

langroid/agent/special/doc_chat_agent.py
dfm88: Thanks for the feedback! 

I've implemented the suggested improvements like this:

1. Combined content+questions in a single field instead of separating them, using a consistent marker (`GENERATED_QUESTIONS_MARKER`) to demarcate them.

2. Added `clean_generated_content()` method in the so that questions are used in ranking but not showed in retrieval

pchalasani: Thank you @dfm88 ! Added to my queue of things to look at this week.

langroid/agent/special/doc_chat_agent.py
```diff
@@ -860,6 +879,60 @@ def llm_hypothetical_answer(self, query: str) -> str:
...
+        """
+        if self.llm is None:
+            raise ValueError("LLM not set")
+
+        if not self.config.hypothetical_questions:
+            return docs
+
+        with status("[cyan]Generating hypothetical questions for chunks..."):
+            # Process chunks in parallel using run_batch_agent_method
+            questions_batch = run_batch_agent_method(
```
pchalasani, line 899: there could be 1000+ chunks so we should use a batched version of this method. I know there's no `batch_size` param in this method, so we'd have to add that to `run_agent_method_batch` and related onces

langroid/agent/special/doc_chat_agent.py
```diff
@@ -1315,6 +1388,35 @@ def get_relevant_extracts(self, query: str) -> Tuple[str, List[Document]]:
 
         return query, extracts
 
+    def clean_generated_content(self, passages: List[Document]) -> List[Document]:
```
pchalasani, line 1391: suggest rename - `remove_hypothetical_questions` to be more clear

langroid/agent/special/doc_chat_agent.py
```diff
@@ -126,6 +128,20 @@ class DocChatAgentConfig(ChatAgentConfig):
     # https://arxiv.org/pdf/2212.10496.pdf
     # It is False by default; its benefits depends on the context.
     hypothetical_answer: bool = False
+    # Flag to enable Hypothetical Question Indexing,
+    # which uses a language model to generate questions for each data chunk,
+    # this might help in finding relevant chunks for a given question.
+    hypothetical_questions: bool = False
```
pchalasani, line 134: add config:
hypothetical_questions_batch_size: int = 50

tests/main/test_doc_chat_agent.py
```diff
@@ -939,3 +940,165 @@ def gen_task(i: int):
...
         a.clear()
+
+
+@pytest.mark.parametrize(
+    "vecdb", ["lancedb", "qdrant_local", "qdrant_cloud", "chroma"], indirect=True
+)
+@pytest.mark.parametrize(
+    "hypothetical_questions, expect_cleaned", [(True, True), (False, False)]
+)
+def test_clean_generated_content(
```
pchalasani, line 951: ==> `test_remove_generated_questions`

langroid/agent/special/doc_chat_agent.py
dfm88: I thoughy it might need for other cleaning stuff so I used a generic name, but ok I'll make it specific for this task

tests/main/test_doc_chat_agent.py
pchalasani, line 1: It would be nice if you can think of a small test that shows the benefit of adding the hyp questions.

Or better, an example scripts under `examples/docqa/` would be great.

pchalasani (changes requested):
Thanks @dfm88 I just reviewed it.

dfm88:
Hi @pchalasani , sorry for the delay in making changes on this. Here an overview of the changes, since it was not a straightforward task to handle batch number:

I saw you added a function in `batch.py` file to handle batches, that actually was handling only sync code, I slughtly changed it to also handle aysnc code and try to reuse it in the `run_batch_agent_method` 

The most challenging part was ensuring correct behavior when stop_on_first_result=Truem so I maintained strict task ordering by preserving original indices.
Force batch_size=1 when stop_on_first_result=True to prevent later tasks from completing first
Use an async lock to ensure thread-safe access when storing the first completed result
Carefully reconstruct the result list to preserve the original task positions

For normal operation (when stop_on_first_result=False), the code fully leverages the batching capabilities.

I have also added an example of when leveraging Hypothetical Questions might be beneficial. Here another interesting article [Advanced RAG Optimization: Aligning Question and Document Embedding Spaces with Hypothetical Questions](https://blog.epsilla.com/demystifying-rag-empowered-chat-agents-aligning-question-and-document-embedding-spaces-with-5710c4218464)

I know there are a lot of changes, but tests using batch module are passing. Let me know if you'd like me to clarify any part of these changes or have suggestions for improvement.

pchalasani:
Thank you @dfm88 will check in a few days.

langroid/agent/batch.py
```diff
@@ -281,50 +297,64 @@ def run_batch_agent_method(
     agent_cls = type(agent)
     agent_name = agent_cfg.name
 
-    async def _do_task(input: str | ChatDocument, i: int) -> Any:
-        agent_cfg.name = f"{agent_cfg.name}-{i}"
+    first_result = None
```
pchalasani, line 300: Thanks for this @dfm88.

Is there a reason the implementation of `stop_on_first_result` here is different from the one in `run_batch_task_gen` -- is this somehow better? This will help me better understand + review.

dfm88: Hi @pchalasani,

thank you for pointing this out. I actually didn't notice the `run_batch_task_gen` implementation when I was working on this PR. Looking at it now, I can see it already provides a good handling of batch async tasks, so I basically took inspiration from that to implement `batch_size` also in `run_batch_agent_method` in my last commit.

So I could revert also the changes I did to `langroid/agent/batch.py::run_batch_function`.

Finally I fixed the tests where I still was using the previous naming for the doc agent method relative to hypothetical questions.

pchalasani: Will review soon thanks !

langroid/agent/batch.py
```diff
@@ -312,17 +317,40 @@ async def _do_all() -> List[Any]:
...
+
+        for batch in batches:
+            start_idx = len(results)
+            complete_str = f", {start_idx} complete" if start_idx > 0 else ""
+            msg = (
+                f"[bold green]Running {len(items)} copies of "
+                f"{agent_name}{complete_str}..."
+            )
+
+            if stop_on_first_result and any(results):
```
pchalasani, line 347: I think we should discard None results here, just like we do in the `run_batch_task_gen` fn.

Also more generally, since this `_do_all` functions in the two places are so similar, I think we should avoid code duplication and have a common fn and use it in both places. What do you think @dfm88 ?

examples/docqa/hypothetical_questions.py
```diff
@@ -0,0 +1,154 @@
...
+    doc_agent = DocChatAgent(config=config)
+    print("[blue]Welcome to the document chatbot!")
+    doc_agent.ingest_doc_paths(
+        "https://www.italia.it/en/italy/things-to-do/pasta-types-italian-formats-and-recipes"
+    )
+
+    doc_task = Task(
+        doc_agent,
+        interactive=False,
+        name="DocAgent",
```
pchalasani, line 102: set `single_round=True` so it exits after generating the answer -- this avoids the current behavior of "stuck for 5 steps"

examples/docqa/hypothetical_questions.py
pchalasani, line 1: I ran this with and without hyp questions and got the exact same  result:

image

pchalasani (changes requested):

langroid/agent/special/doc_chat_agent.py
```diff
@@ -81,6 +82,8 @@ def apply_nest_asyncio() -> None:
 You are a helpful assistant, helping me understand a collection of documents.
 """
 
+GENERATED_QUESTIONS_MARKER = "--- Generated Questions ---"
```
pchalasani, line 85: The chunk-embeddings will be distracted by the presence of the words "generated questions", and it would thus affect retrieval as well. Instead, this should be set to something non-semantic but unique like 
```
<##-##-##>
```

To be even safer, we could try to make the vecdb aware of the presence of this marker and strip it out before embedding, but that is messy (we have to propagate the hypo questions flag all the way into the `VectorStoreConfig` -- we can skip that for now.

pchalasani (changes requested):

examples/docqa/hypothetical_questions.py
dfm88: Oh yeah, I see. So given that Hypothetical Question works well in particular scenarios, (e.g., specialized jargon, very technical documents, Q-A) I provided a new example on a langorid python file (doc_chat_agent.py -Inception?-), and the user query `"How does the system generate responses in conversation mode?"` got a more appropriate answer using HQ.

Without HQ:
WITHOUT_HQ

With HQ
WITH_HQ

pchalasani: > doc_chat_agent.py -Inception?-

Didn't understand what you meant by `inception` ?

pchalasani:
@dfm88 this is going to be very useful, as you say in cases with very technical/domain-specific terms that are not captured well by generic embeddings. I changed your example script to the medical domain to illustrate the benefit more clearly and explicitly, see the `docagent-hypoq` branch:
https://github.com/langroid/langroid/blob/docagent-hypoq/examples/docqa/hypothetical_questions.py

I added explanatory notes at the top of the script.

NOTE - you need to pull latest from main (there is a minor fix in `doc_chat_agent.py` and a needed method in `mytypes.py`) and then pull the `hypothetical_questions.py` script from the `docagent-hypoq` branch.

dfm88:
> @dfm88 this is going to be very useful, as you say in cases with very technical/domain-specific terms that are not captured well by generic embeddings. I changed your example script to the medical domain to illustrate the benefit more clearly and explicitly, see the `docagent-hypoq` branch: https://github.com/langroid/langroid/blob/docagent-hypoq/examples/docqa/hypothetical_questions.py
> 
> I added explanatory notes at the top of the script.
> 
> EDIT - Please merge the changes from branch `docagent-hypoq` into your branch. The `docagent-hypoq` branch also includes some fixes to `doc_chat_agent.py` and a new method `Document.from_string` in `mytypes.py`

Great  thanks! I did the merge

pchalasani:
@dfm88 I am making some changes on top of this.

dfm88:
> @dfm88 I've made a new PR with my changes on top, #683 I made the design more flexible, to support any type of chunk enrichment, and also allow a different LLM to be used for the chunk enrichment. Essentially, we can can pass a config for a "chunk enrichment agent".

Yeah I see, it makes a lot of sense to me, thanks!

pchalasani:
> Yeah I see, it makes a lot of sense to me, thanks!

Thank you for leading this PR !

pchalasani:
finished in #683

PR #612, opened 2024-11-16
change: parameter name to 'message' in llm_response and llm_response_async

dfm88:
Uniformized to `message` the parameter in the `llm_response` agents classes methods.

PR #616, opened 2024-11-18
chore: parameter name to 'message' in llm_response and llm_response_async (clone of #612)

pchalasani:
This is a clone of #612 to force gh actions to run, to see if tests pass.

### DIFF
--- a/examples/basic/multi-agent-triage.py
+++ b/examples/basic/multi-agent-triage.py
@@ -147,9 +147,9 @@ def main(model: str = ""):
     class CoursesAgent(lr.agent.special.DocChatAgent):
         def llm_response(
             self,
-            query: None | str | ChatDocument = None,
+            message: None | str | ChatDocument = None,
         ) -> Optional[ChatDocument]:
-            answer = super().llm_response(query)
+            answer = super().llm_response(message)
             if answer is None:
                 return None
             return self.create_llm_response(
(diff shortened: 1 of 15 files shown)


pchalasani:
@dfm88 thanks for this PR! The tests are passing so the changes in the core lib are good. I also visually checked the changes in the examples, and they look good. Curious, how did you find all these cases -- did you use some type of strict checker?

dfm88:
> @dfm88 thanks for this PR! The tests are passing so the changes in the core lib are good. I also visually checked the changes in the examples, and they look good. Curious, how did you find all these cases -- did you use some type of strict checker?

Hi @pchalasani thanks :).
So about the different parameters names, I realized it because in my DocChatAgent child class I had to call the llm_response from ChatAgent so

```python
Class ConversationMemoryDocAgent(DocChatAgent):
    ...
    def llm_response(
        self,
        query: None | str | ChatDocument = None,
    ) -> Optional[ChatDocument]:
        # this was raising TypeError since passing the named argument the call 
        # to the CahtAgent corresponding method raised 
        # DocChatAgent.llm_response() got an unexpected keyword argument 'query'
        super().llm_response(query=query) 
```

About the replace in the code I first searched in all the codebase for `"def llm_response"` to find all places where it was defined  and after that I searched for its usage `"llm_response("` to check if anywhere it was used with named arguments

PR #621, opened 2024-11-25
Strict decoding enhancements

nilspalumbo:
Automatically enables `output_format` use and handling and generates instructions for the system message. Additionally, updates `SQLChatAgent` to use the `AnyTool` mechanism instead of the helper agent mechanism on fallback if strict JSON mode is available.


pchalasani:
thanks @nilspalumbo ! let me know when ready to review.

nilspalumbo:
Should be ready now! The test failure seems to be some quota issue for embeddings.

langroid/agent/chat_agent.py
```diff
@@ -737,16 +789,95 @@ def set_output_format(
...
+                        self.llm_functions_usable
+                    )
+                    is_handled = name in self.llm_tools_handled.union(
+                        self.llm_functions_handled
+                    )
+
+                    if is_copy:
+                        # We must copy `llm_tools_usable` so the base agent
+                        # is unmodified
+                        self.llm_tools_usable = copy.copy(self.llm_tools_usable)
```
pchalasani, line 810: should this be inside a `if use:` block, similar to `if handle` ?

langroid/agent/chat_agent.py
```diff
@@ -708,15 +729,46 @@ def set_output_format(
...
+        copy certain fields to ensure that we do not overwrite the main agent's
+        setings.
+        """
+        # Disable usage of an output format which was not specifically enabled
+        # by `enable_message`
+        if self.enabled_use_output_format is not None:
+            self.disable_message_use(self.enabled_use_output_format)
+            self.enabled_use_output_format = None
+
+        # Disable usage of an output format which did not specifically have
```
pchalasani, line 755: typo: usage -> handling

langroid/agent/chat_agent.py
```diff
@@ -737,16 +789,95 @@ def set_output_format(
...
+                    # it should be disabled when `output_format` is changed
+                    if not is_usable:
+                        self.enabled_use_output_format = output_type
+
+                    # If `output_type` was not already handled, record that
+                    # it should be disabled when `output_format` is changed
+                    if not is_handled:
+                        self.enabled_handling_output_format = output_type
+
+                generated_tool_instructions = name in self.llm_tools_usable.union(
```
pchalasani, line 836: `... =  is_usable`

langroid/agent/chat_agent.py
```diff
@@ -627,6 +644,16 @@ def enable_message(
             if handle:
                 self.llm_tools_handled.add(t)
                 self.llm_functions_handled.add(t)
+
+                # Track whether `output_format` handling was
+                # explicitly enabled and should not be disabled when
+                # `output_format` changes
+                if self.enabled_handling_output_format is not None:
+                    if (
```
pchalasani, line 652: shorten with `and` instead of nested `if`

langroid/agent/chat_agent.py
```diff
@@ -647,6 +674,12 @@ def enable_message(
                         set `_allow_llm_use=True` when you define the tool.
                         """
                     )
+                # Track whether `output_format` use was explicitly
+                # enabled and should not be disabled when
+                # `output_format` changes
+                if self.enabled_use_output_format is not None:
+                    if self.enabled_use_output_format.default_value("request") == t:
```
pchalasani, line 681: shorten with `and` instead of nested `if`

langroid/agent/chat_agent.py
```diff
@@ -179,8 +187,16 @@ def __init__(
         self.output_format: Optional[type[ToolMessage | BaseModel]] = None
 
         self.saved_requests_and_tool_setings = self._requests_and_tool_settings()
+        # Tracks if we have enabled use of an output format
```
pchalasani, line 190: Needs better explanation, took me a while to grok it; suggest this  (maybe you need to edit it if not quite right):

When this is not None and equals a tool msg T, this means that 
(a) T has been set as the `output_format` of this agent, AND 
(b) T has been "enabled for use" ONLY for enforcing this output format.
Thus when the `output_format` changes, T will need to be "disabled for use".

langroid/agent/chat_agent.py
```diff
@@ -179,8 +187,16 @@ def __init__(
...
+        # We disable use when the output format is changed for any such tool
+        # where `enable_message` was not called explicitly
+        self.enabled_use_output_format: Optional[type[ToolMessage]] = None
+        # As above but used to disable handling
```
pchalasani, line 194: Change comment to -- 

As above but deals with "enabled for handling" rather than "enabled for use".

langroid/agent/chat_agent.py
```diff
@@ -627,6 +644,16 @@ def enable_message(
...
+
+                # Track whether `output_format` handling was
+                # explicitly enabled and should not be disabled when
+                # `output_format` changes
+                if self.enabled_handling_output_format is not None:
+                    if (
+                        self.enabled_handling_output_format.default_value("request")
+                        == t
+                    ):
+                        self.enabled_handling_output_format = None
```
pchalasani, line 656: Comment right above the `... = None`:

`t` was designated as "enabled for handling" ONLY for `output_format` enforcement, but we are explicitly enabling it for handling here, so we set the variable to None.

langroid/agent/chat_agent.py
```diff
@@ -627,6 +644,16 @@ def enable_message(
             if handle:
                 self.llm_tools_handled.add(t)
                 self.llm_functions_handled.add(t)
+
+                # Track whether `output_format` handling was
```
pchalasani, line 648: remove this comment, replace with suggested comment inside if block

langroid/agent/chat_agent.py
```diff
@@ -647,6 +674,12 @@ def enable_message(
                         set `_allow_llm_use=True` when you define the tool.
                         """
                     )
+                # Track whether `output_format` use was explicitly
+                # enabled and should not be disabled when
+                # `output_format` changes
+                if self.enabled_use_output_format is not None:
+                    if self.enabled_use_output_format.default_value("request") == t:
+                        self.enabled_use_output_format = None
```
pchalasani, line 682: comment inside if block, right above this None assignment:

t was designated as "enabled for use" ONLY for output_format enforcement, but we are explicitly enabling it for use here, so we set the variable to None.

pchalasani (changes requested):
@nilspalumbo thanks for this update. I added/edited some comments in various places to help/improve (my) understanding of the code. 

There are a couple code questions I had.

Also the `test_sql_chat_agent` is failing.

pchalasani:
@nilspalumbo this is the error I'm getting locally with `test_sql_chat_agent.py`:

```
tests/main/sql_chat/test_sql_chat_agent.py:162: in _test_sql_chat_agent
    result = task.run(prompt, turns=turns)
langroid/agent/task.py:688: in run
    self.step()
langroid/agent/task.py:1107: in step
    result = self.response(r, turns)
langroid/agent/task.py:1427: in response
    result = response_fn(self.pending_message)
langroid/agent/base.py:421: in agent_response
    results = self.handle_message(msg)
langroid/agent/base.py:1238: in handle_message
    results = [self.handle_tool_message(t, chat_doc=chat_doc) for t in tools]
langroid/agent/base.py:1238: in <listcomp>
    results = [self.handle_tool_message(t, chat_doc=chat_doc) for t in tools]
langroid/agent/base.py:1583: in handle_tool_message
    raise e
langroid/agent/base.py:1577: in handle_tool_message
    maybe_result = handler_method(tool)
langroid/agent/base.py:314: in <lambda>
    setattr(self, tool, lambda obj: obj.response(self))
langroid/agent/chat_agent.py:1061: in response
    agent.set_output_format(None)
langroid/agent/chat_agent.py:774: in set_output_format
    self.config = copy.deepcopy(self.config)
/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:172: in deepcopy
    y = _reconstruct(x, memo, *rv)
/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:271: in _reconstruct
    state = deepcopy(state, memo)
/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:146: in deepcopy
    y = copier(x, memo)
/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:231: in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:146: in deepcopy
    y = copier(x, memo)
/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:231: in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:172: in deepcopy
    y = _reconstruct(x, memo, *rv)
/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:271: in _reconstruct
    state = deepcopy(state, memo)
/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:146: in deepcopy
    y = copier(x, memo)
/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:231: in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:172: in deepcopy
    y = _reconstruct(x, memo, *rv)
/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:271: in _reconstruct
    state = deepcopy(state, memo)
/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:146: in deepcopy
    y = copier(x, memo)
/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:231: in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:172: in deepcopy
    y = _reconstruct(x, memo, *rv)
/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:271: in _reconstruct
    state = deepcopy(state, memo)
/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:146: in deepcopy
    y = copier(x, memo)
/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:231: in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:172: in deepcopy
    y = _reconstruct(x, memo, *rv)
/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:271: in _reconstruct
    state = deepcopy(state, memo)
/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:146: in deepcopy
    y = copier(x, memo)
/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:231: in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

x = <module 'sqlite3.dbapi2' from '/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/sqlite3/dbapi2.py'>
memo = {13451490704: PromptsConfig(max_tokens=1000), 13452568960: ['\n\n', '\n', ' ', ''], 13452608064: {'_sa_instance_state'...tanceState object at 0x323b27a10>}, 13452609168: <sqlalchemy.orm.identity.WeakInstanceDict object at 0x322dbd250>, ...}
_nil = []

    def deepcopy(x, memo=None, _nil=[]):
        """Deep copy operation on arbitrary Python objects.

        See the module's __doc__ string for more info.
        """

        if memo is None:
            memo = {}

        d = id(x)
        y = memo.get(d, _nil)
        if y is not _nil:
            return y

        cls = type(x)

        copier = _deepcopy_dispatch.get(cls)
        if copier is not None:
            y = copier(x, memo)
        else:
            if issubclass(cls, type):
                y = _deepcopy_atomic(x, memo)
            else:
                copier = getattr(x, "__deepcopy__", None)
                if copier is not None:
                    y = copier(memo)
                else:
                    reductor = dispatch_table.get(cls)
                    if reductor:
                        rv = reductor(x)
                    else:
                        reductor = getattr(x, "__reduce_ex__", None)
                        if reductor is not None:
>                           rv = reductor(4)
E                           TypeError: cannot pickle 'module' object
```

pchalasani:
The above `test_sql_chat_agent.py` failure disappears when I use `gpt-4`, but worth looking at what exec-path is causing the pickle failure.

pchalasani:
Also I was able to reproduce the `test_task::test_output_format_sequence_async.py` that is failing on  gh actions (see image). The LLM is producing the wrong tool structure repeatedly - 

image

pchalasani:
Dug into these test failures - they have to do with the fact that `gpt-4` doesn't have structured output:

- the failure in `test_output_format_sequence_async.py` occurs in gh actions when the original test with `gpt-4o` failed and it's being retried with `gpt-4` -- the original failure is probably an LLM deviation that we can't do much about.
- the failure in `test_sql_chat_agent.py` I am getting locally is with `gpt-4o-2024-11-20`, and it disappeared with `gpt-4` because with the latter model we don't activate the strict-decoding, and the test passes via retries using the old mechanisms. This is also why it passes in GH actions. So this failure with `gpt-4o-2024-11-20` is a real failure that needs to be fixed.

pchalasani:
> * failure with `gpt-4o-2024-11-20` is a real failure that needs to be fixed.

The root cause is that `copy.deepcopy(self.config)` in `ChatAgent.set_output_format` is failing b/c for a `SQLChatAgent`, the `self.config` contains a field `database_session: Session` which is not pickle-able. Should we use something like `dill` to try to do `deepcopy` regardless of what `Config` may contain?

nilspalumbo:
Hmm, we may be able to replace the deep copy of the config with `config.copy()`; that should fix the issue.

nilspalumbo:
That seems to fix the failure; do we want to replace other uses of `deepcopy` on configs with the same? `config.copy()` is a shallow copy, but handles `BaseModel` correctly by descending into the object's `__dict__`, which is enough for the strict copy mechanism but perhaps not for `agent.clone()`.

pchalasani:
> but perhaps not for `agent.clone()`

let's do whatever needs to be done just for the strict decoding to work right. Agent.clone already has a caveat

PR #626, opened 2024-11-29
multi-agent debate example with fixes from the feedback

adam-shams:
Submitting for review to contribute

Overview
--------
This example is a debate system powered by LLMs using Langroid, enabling structured debates on various topics such as use of AI in healthcare, education, intellectual property, and societal biases. This can be used for other kinds of debates as well. The program creates and manages agents that represent opposing sides of a debate, interact with users, and provide constructive feedback based on established debate criteria.

Features
--------
- Multiple Debate Topics:
  - AI in Healthcare
  - AI and Intellectual Property
  - AI and Societal Biases
  - AI as an Educator
- Agent-Based Interaction:
  - Pro and Con agents for each topic simulate structured debate arguments.
- Customizable LLM Configuration:
  - Support for different LLMs like GPT-4o, GPT-4, and Mistral.
- Feedback Mechanism:
  - Provides structured feedback on debate performance based on key criteria.
- Interactive or Autonomous Mode:
  - Users can either control interactions manually or let agents autonomously continue debates.


examples/multi-agent-debate/config.py
```diff
@@ -0,0 +1,70 @@
...
+        "1: GPT4o\n"
+        "2: GPT4\n"
+        "3: Mistral: mistral:7b-instruct-v0.2-q8_0a\n"
+        "Enter 1, 2, or 3:",
+        choices=["1", "2", "3"],
+        default="1"
+    )
+
+    model_map = {
+        "1": "GPT4o",
```
pchalasani, line 39: use the actual openai mdl names here , e.g. `gpt-4`, `gpt-4o`, then directly use it to set `chat_model`, no need for `getattr`

examples/multi-agent-debate/config.py
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    if chat_model_option == "3":
+        chat_model = "ollama/mistral:7b-instruct-v0.2-q8_0"
+        base_llm_config = lm.OpenAIGPTConfig(
+            chat_model=chat_model,
+            chat_context_length=16_000  # Only set for Ollama model
+        )
+    else:
+        chat_model = getattr(lm.OpenAIChatModel, model_map[chat_model_option])
```
pchalasani, line 50: avoid getattr, see prev comment

examples/multi-agent-debate/agents.py
```diff
@@ -0,0 +1,172 @@
...
+        vecdb=None,
+        system_message=system_message,
+
+    )
+
+    return lr.ChatAgent(config)
+
+
+# Agents for AI in Healthcare
+def create_pro_ai_agent(base_llm_config: OpenAIGPTConfig) -> lr.ChatAgent:
```
pchalasani, line 39: avoid defining a fn for each agent type. see later comment in main.py

pchalasani: have a dict that holds the system messages, keyed by agent type

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,137 @@
...
+    logger.info("Setting up agents and tasks...")
+    global_settings = get_global_settings(debug, nocache)
+    lr.utils.configuration.set_global(global_settings)
+    base_llm_config = get_base_llm_config()
+
+    agents = {}
+    tasks = {}
+
+    for _, (topic_name, pro_key, con_key) in DEBATE_TOPICS.items():
+        agents[pro_key] = globals()[f"create_{pro_key}_agent"](base_llm_config)
```
pchalasani, line 49: no need for globals() etc.
If you have system msgs keyed by topic type, you can directly call the generic `create_agent` with the appropriate system msg here, rather than calling the topic-specific agent-creator. Imagine if there are 100 different topics, you wouldn't want to have 100 different specialized agent-definition fns -- instead, you'd have 100 sys msgs keyed by topic, and call the generic agent-defn fn, with the specific sys msg

examples/multi-agent-debate/main_chainlit.py
```diff
@@ -0,0 +1,264 @@
+import chainlit as cl
```
pchalasani, line 1: avoid code duplication, import from parent directory, after refactoring common parts if needed.

examples/multi-agent-debate/tasks.py
pchalasani, line 1: no need for tasks

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,137 @@
...
+# Set up the Typer application
+app = typer.Typer()
+
+# Set up logging
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+lr.utils.logging.setup_colored_logging()
+
+# Mappings for dynamic task and agent creation
+DEBATE_TOPICS = {
```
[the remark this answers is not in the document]
adam-shams: refactored in new commit

examples/multi-agent-debate/agents.py
adam-shams: done

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,137 @@
...
+import langroid as lr
+from agents import (
+    create_pro_ai_agent, create_con_ai_agent,
+    create_pro_ip_agent, create_con_ip_agent,
+    create_pro_bias_agent, create_con_bias_agent,
+    create_pro_edu_agent, create_con_edu_agent,
+    create_feedback_agent,
+)
+from tasks import (
+    create_pro_ai_task, create_con_ai_task,
```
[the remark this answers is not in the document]
adam-shams: done in new commit

examples/multi-agent-debate/tasks.py
adam-shams: removed in new commit

examples/multi-agent-debate/main.py
adam-shams: this was a good suggestion. completely refactored to dynamically generate agents based on system_messages.json.

examples/multi-agent-debate/config.py
```diff
@@ -0,0 +1,70 @@
+from rich.prompt import Prompt
+import langroid.language_models as lm
+import langroid.utils.configuration
+
+
+# Define the streaming handler function here
+def handle_streaming_output(token: str):
```
pchalasani, line 7: Is this meant for chainlit? I don't see the chainlit-related files any more. Maybe they were accidentally omitted from the PR ?

adam-shams: I added this for chainlit streaming support, however I removed the main_chainlit.py from the project. Wanted to make sure first to get the CLI version working well and plan to add chainlit version some time in the future. Currently handle_streaming_output is called from main.py: 
agent_config = get_base_llm_config(streamer=handle_streaming_output).

More likely added when integrating chainlit. I can delete this from config.py and update agent_config = get_base_llm_config() or we can leave it. Open to suggestions. Thanks for reviewing the rest of the code.

pchalasani: yes best to delete this if not relevant for this version

examples/multi-agent-debate/config.py
```diff
@@ -0,0 +1,65 @@
...
+        "1: GPT4o\n"
+        "2: GPT4\n"
+        "3: Mistral: mistral:7b-instruct-v0.2-q8_0a\n"
+        "Enter 1, 2, or 3:",
+        choices=["1", "2", "3"],
+        default="1"
+    )
+
+    model_map = {
+        "1": "gpt-4o",
```
pchalasani, line 44: use `lm.OpenAIChatModel.GPT4o` etc here, so you get the right model names

examples/multi-agent-debate/config.py
```diff
@@ -0,0 +1,65 @@
...
+            stream=True,                # Enable streaming outputs
+            streamer=streamer,          # Set the streaming handler
+        )
+    else:
+        chat_model = model_map[chat_model_option]
+        base_llm_config = lm.OpenAIGPTConfig(
+            chat_model=chat_model,
+            max_output_tokens=1500,  # Adjusted to prevent truncation
+            stream=True,            # Enable streaming outputs
+            streamer=streamer,      # Set the streaming handler
```
pchalasani, line 63: may not need this any more since you're removing the chainlit-related things?

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,204 @@
...
+                # Build the conversation context
+                if is_user_turn:
+                    opponent_arguments = ai_arguments
+                else:
+                    opponent_arguments = student_arguments
+
+                context = "\n".join(opponent_arguments[-1:]) if opponent_arguments else "Start of debate."
+
+                # Prepare the full prompt
+                full_context = f"{current_agent.config.system_message}\n\nOpponent's argument:\n{context}"
```
pchalasani, line 138: The agent already has a `system_message`, so this is re-filling the sys msg as a user-role msg. 
In general try to leverage Langroid's Agent functionality -- it has a `message_history` object that starts with sys msg and alternating user/asst msgs. It automatically accumulates dialog so you should try to use that. When stepping in debugger, look at the `self.message_history` variable to understand more. I don't think you need to manually accumulate the arguments at each turn, unless you're doing something very custom (which I don't think you are).

examples/multi-agent-debate/config.py
```diff
@@ -0,0 +1,65 @@
...
+    Returns:
+        Settings: Langroid configuration settings.
+    """
+    return langroid.utils.configuration.Settings(
+        debug=debug,
+        cache=not nocache,
+    )
+
+
+def get_base_llm_config(streamer=None):
```
pchalasani, line 23: remove streamer?

pchalasani: ... here and other streamer-related code

examples/multi-agent-debate/config.py
adam-shams: agreed, removed.

examples/multi-agent-debate/config.py
adam-shams: > use the actual openai mdl names here , e.g. `gpt-4`, `gpt-4o`, then directly use it to set `chat_model`, no need for `getattr`

do you recommend `lm.OpenAIChatModel.GPT4o` be used for `chat_model_option` instead of `model_map` or somewhere else?

pchalasani: I just meant in the `model_map` -- "gpt-4o" is not an actual model name in langroid -- we use "gpt-4o-2024-11-20" to make sure it points to a fixed model. But if you use `lm.OpenAIChatModel.GPT4o` you can be sure it is a valid name

pchalasani:
@adam-shams regarding the model-name issue, it's fixed in latest langroid we pushed today 0.24.1

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,325 @@
...
+        for turn in range(max_turns):
+            if llm_delegate:
+                # Autonomous LLM debate
+                current_agent = user_agent if is_user_turn else ai_agent
+                agent_role = user_side if is_user_turn else ai_side
+                opponent_arguments = (
+                    ai_arguments if is_user_turn else student_arguments
+                )
+                context = (
+                    "\n".join(opponent_arguments[-1:])
```
pchalasani, line 241: Here and elsewhere I think we can avoid this manual accumulation of conversation. E.g., we have the following input-output sequence of  msgs:

```
Student (or agent proxy) (StudentSysMsg) =>  S1  (i.e. student sees sys msg, says "S1")

AI (AI_Sys_msg, S1) =>  A1 (i.e. AI sees AI_Sys_Msg, S1, responds with A1)

Student (StudentSysMsg, S1, A1) => S2

AI(AI_SysMsg, S1, A1, S2) => A2
...
```

At each turn, each side is responding based on their own system msg, and **every** intervening msg of both sides, so you can simply pass on the `llm_response` of each agent to the other. I suggest as a simple exercise just set up two agents and try this. In fact we could probably create a task around each agent and have one task be subtask of the other, and count the turns and exit when needed. I'll see if I can mock up a simple example in next few days. You can try as well. I don't know if I'm over-simplifying what you're trying to do, but it will be nice to leverage langroid's built-in features if possible. I'll have to give it a closer look.

pchalasani: For example you could probably re-work the teacher/student example in the langroid tour into a debate example: https://langroid.github.io/langroid/tutorials/langroid-tour/
The twist in what you're doing is the final feedback, and for that I think we could pass on a digest of the final responder's msg history to the feedback agent.

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,300 @@
...
+    """Execute the main debate logic.
+
+    Orchestrates the debate process, including setup, user input, LLM agent
+    interactions, and final feedback. Handles both user-guided and LLM-
+    delegated debates.
+
+    Raises:
+        Exception: If an unexpected error occurs during the debate process.
+    """
+    try:
```
pchalasani, line 185: no need to enclose the whole thing in try/except since you're re-raise the exception anyway.

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,300 @@
...
+            user_side, ai_side = "Con", "Pro"
+
+        logger.info(
+            f"Starting debate on topic: {topic_name}, taking the {user_side} side. "
+            f"LLM Delegate: {llm_delegate}"
+        )
+
+        is_user_turn: bool = True
+
+        if side == "pro":
```
pchalasani, line 251: I'm pretty sure the pro and con branches can be unified to make the code more compact, since there's a lot of similar-looking code.

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,300 @@
...
+            else:
+                user_input = Prompt.ask(
+                    "Your argument (or type 'f' for feedback, 'done' to end):"
+                )
+                pro_agent.llm_response(user_input)
+            pro_agent.enable_message(DoneTool)
+            pro_agent_task = lr.Task(pro_agent, interactive=False)
+            con_agent_task = lr.Task(con_agent, interactive=False, single_round=True)
+            pro_agent_task.add_sub_task(con_agent_task)
+            pro_agent_task.run(turns=max_turns)
```
pchalasani, line 266: `task.run()` with no starting msg may not work with some models (e.g. gemini) which throw an err when the first msg after the sys msg is not a user-role msg. So you can pass in a dummy msg like `task.run("get started", turns=...)` as I did in my example

pchalasani (changes requested):
Code is looking much clearer and really leveraging langroid features! I had a couple suggestions.

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,300 @@
...
+            agent_name = pro_agent
+            pro_agent.clear_history()
+            logger.info(f"\n{side} Agent ({topic_name}):\n")
+            if llm_delegate:
+                logger.info("Autonomous Debate Selected")
+            else:
+                user_input = Prompt.ask(
+                    "Your argument (or type 'f' for feedback, 'done' to end):"
+                )
+                pro_agent.llm_response(user_input)
```
pchalasani, line 261: Note that when the `pro_agent_task` is run, it resets the msg history (except sys msg), so to prevent that, you should use `restart=False` when you create the task down below.

pchalasani: Actually , even simpler — omit this llm-response line and just pass the user_input as the initial msg when calling the pro task run method below

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,300 @@
...
+            f"LLM Delegate: {llm_delegate}"
+        )
+
+        is_user_turn: bool = True
+
+        if side == "pro":
+            agent_name = pro_agent
+            pro_agent.clear_history()
+            logger.info(f"\n{side} Agent ({topic_name}):\n")
+            if llm_delegate:
```
pchalasani, line 255: I think in case of `llm_delegate =False`, you actually don't want the LLM of the agent to say anything at all, and instead just the user enters their side of the argument? So when you initialize the corresponding agent, you can set the llm component to None (an agent need not have an LLM), so its only responders would be `user_response` and `agent_response` (that latter is not used in your example since there are no tools). Alternatively, you can define a subclass of `ChatAgent` and override the `llm_response` method to return None. But it's easier to just set `llm` to None in the agent (or agent config, check which one works).

examples/multi-agent-debate/main.py
adam-shams: gemini is working now. thanks for this!

examples/multi-agent-debate/main.py
adam-shams: working on it

pchalasani:
@adam-shams let me know when this ready to review

adam-shams:
@pchalasani ready for review. Thank you for all your feedback

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,391 @@
...
+    """
+
+    global_settings = get_global_settings(nocache=True)
+    lr.utils.configuration.set_global(global_settings)
+    same_llm: bool = is_same_llm_for_all_agents()
+    llm_delegate: bool = is_llm_delegate()
+
+    # Get base LLM configuration
+    if same_llm:
+        shared_agent_config: OpenAIGPTConfig = get_base_llm_config("for main LLM")
```
pchalasani, line 280: I think the word "for" should be dropped since the template already has "for" in it

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,391 @@
...
+        """
+        return ChatAgent(
+            ChatAgentConfig(
+                llm=llm_config,
+                name=name,
+                system_message=system_message,
+            )
+        )
+
+    pro_agent = create_chat_agent("Pro", pro_agent_config, system_messages.messages[pro_key].message +
```
pchalasani, line 319: here and in the next 3 lines, break the args into multiple lines for easier readability.

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,391 @@
...
+        logger.info("Autonomous Debate Selected")
+        interactive_setting = False
+    else:
+        logger.info("Manual Debate Selected with an AI Agent")
+        interactive_setting = True
+        user_input: str = Prompt.ask(
+            "Your argument (or type 'f' for feedback, 'done' to end):"
+        )
+        user_agent.llm = None  # User message without LLM completion
+        user_agent.llm_response(user_input)
```
pchalasani, line 353: Not sure why we would call this when llm is None

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,391 @@
...
+from langroid.agent.chat_agent import ChatAgent, ChatAgentConfig
+from langroid.agent.task import Task
+from langroid.agent.tools.orchestration import DoneTool
+from langroid.agent.tools.google_search_tool import GoogleSearchTool
+from langroid.utils.logging import setup_logger
+from langroid.utils.configuration import set_global
+from config import get_base_llm_config, get_global_settings
+from models import SystemMessages, Message
+
+DEFAULT_SYSTEM_MESSAGE_ADDITION = f"""
```
pchalasani, line 17: This will _not_ work no matter how much we plead 😀 - this is because LLMs have no ability to differentiate real from fake URLs. Also you are asking it to use sources from a web search, but you are not enabling the agent with a web-search tool (you're only enabling it for the validation agent at the end). A better way to set this up is to enable both the debate agents to use a web-search tool. The `task.run()` loop handles a web-search tool call, returns results to the LLM. Your sys msgs will have to instruct the LLM to support its args with references from a web search, indicating which tool to use (see comment below). Also when the sub-task agent uses a tool, `single_round=True` won't work (since it will just get llm response and return, without handling the tool), so you have to instruct it to use the DoneTool when done. You can search for `DoneTool` and `MetaphorSearchTool` under `examples` to see how those are used. All in all, the addition of a tool somewhat complicates the setup, so it is up to you if you want to get into it. (tip - install the `ripgrep` tool for fast command-line search using `rg MetaphorSearchTool -t py`)

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,391 @@
...
+            6. Technical Execution: Identify if major arguments were addressed or dropped and check consistency.
+            7. Debate Etiquette: Evaluate professionalism, respect, and demeanor.
+            8. Final Focus: Judge the strength of closing speeches, how well they summarize the case, 
+            and justify a winner.
+            Provide constructive feedback for each debater, 
+            summarizing their performance and declaring a winner with justification.
+            """
+GOOGLE_SEARCH_SYSTEM_MESSAGE = f"""
+            You are a helpful assistant. Extract all the links in references, ensure the URLs looks valid.  
+            Use the GoogleSearchTool to validate the references.
```
pchalasani, line 43: When this tool is enabled, langroid generates instructions using the tool's `request` field as the name of the tool, so you have to say something like 

> use the TOOL `{GoogleSearchTool.name()}` ...

examples/multi-agent-debate/main.py
adam-shams: This makes sense. Got the key for MetaphorSearchTool. Will implement in the future V2 version of debate system.  For right now I removed the reference to Web Search.

adam-shams:
@pchalasani ready for review. I have removed the GoogleSearch agent and instead incorporated the GoogleSearchTool and MetaphorSearchTool in the feedback agent. The idea is that after the debate is analyzed the two tools are used to provide user links to gain additional knowledge about the subject. For the future, I am thinking about passing these urls to DocChat Agent with the urls and the user can interact with a grounded sources, but that's for V2 of this system. Thanks again for all your help and feedback.

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,407 @@
...
+from langroid.agent.task import Task
+from langroid.agent.tools.google_search_tool import GoogleSearchTool
+from langroid.agent.tools.metaphor_search_tool import MetaphorSearchTool
+from langroid.utils.logging import setup_logger
+from langroid.utils.configuration import set_global
+from config import get_base_llm_config, get_global_settings
+from models import SystemMessages, Message
+
+DEFAULT_SYSTEM_MESSAGE_ADDITION = f"""
+            PLEASE PROVIDE REFERENCES WITH URLS FOR ALL YOUR ARGUMENTS.
```
pchalasani, line 18: We shouldn't ask the LLM to produce URLs, they will almost never be reliable, unless they are well-known top-level names. Also, I don't think you're asking the feedback agent to verify the URLs, so best to drop.
It's best to not even say "ensure real sources" when we are not giving it access to docs. The best it can do is to produce text from its own baked-in "knowledge" in its weights.

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,407 @@
...
+            4. Rebuttals: Analyze how well each side refutes their opponent's arguments.
+            5. Persuasion: Assess communication quality, tone, rhetorical effectiveness, and emotional/ethical appeals.
+            6. Technical Execution: Identify if major arguments were addressed or dropped and check consistency.
+            7. Debate Etiquette: Evaluate professionalism, respect, and demeanor.
+            8. Final Focus: Judge the strength of closing speeches, how well they summarize the case, 
+            and justify a winner.
+            Provide constructive feedback for each debater, 
+            summarizing their performance and declaring a winner with justification.
+            After providing the above feedback:
+            Use the TOOL {GoogleSearchTool.name()} to search the web for 2 references for each pro and con.
```
pchalasani, line 41: Any reason to use both Google and Metaphor search? To keep it simple, suggest using just one, preferably Metaphor.

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,407 @@
...
+    ai_task = Task(ai_agent, interactive=False, single_round=True)
+    user_task.add_sub_task(ai_task)
+    user_task.run("get started", turns=max_turns)
+
+    # Determine the last agent based on turn count and alternation
+    # Note: user_agent and ai_agent are dynamically set based on the chosen user_side
+    last_agent = ai_agent if max_turns % 2 == 0 else user_agent
+
+    # Generate feedback summary and declare a winner using feedback agent
+    validation_message: str = last_agent.message_history
```
pchalasani, line 388: this will have type `List[LLMMessage]`, not `str`.

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,407 @@
...
+    if not validation_message:
+        logger.warning("No agent message history found for the last agent")
+        validation_message = "No last agent history available to validate."
+
+    feedback_agent.clear_history()
+    feedback_agent.enable_message(GoogleSearchTool)
+    feedback_agent.enable_message(MetaphorSearchTool)
+    feedback_agent_task = Task(feedback_agent, interactive=False)
+    feedback_agent_task.run(validation_message)
```
pchalasani, line 397: The `validation_message` is a List of `LLMMessage` which will get converted to a string -- this may be ok since it will preserve each side's arguments. As long as you're aware of this.

adam-shams: Acknowledged

examples/multi-agent-debate/main.py
adam-shams: Updated the type

examples/multi-agent-debate/main.py
adam-shams: I was just comparing Metaphor and Google, but agree with the simplification. I've removed Google

examples/multi-agent-debate/main.py
adam-shams: Cleaned up the prompt.

adam-shams:
@pchalasani Thank you for the feedback! Resolved all comments in new commit.

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,399 @@
...
+    """Prompt the user to decide on LLM delegation.
+
+    Asks the user whether the LLM should autonomously continue the debate
+    without requiring user input.
+
+    Returns:
+        bool: True if the user chooses LLM delegation, otherwise return False.
+    """
+    return Confirm.ask(
+        "Would you like the LLM to autonomously continue the debate without "
```
pchalasani, line 246: this question is confusing - you're asking two questions, then y/n. Make the question crisp so that it is clear that a yes/no answer applies

examples/multi-agent-debate/config.py
```diff
@@ -0,0 +1,134 @@
...
+        config_agent_name (str): The name of the agent being configured, used
+                                 in the prompt to personalize the message.
+
+    Returns:
+        str: The user's selected option as a string, corresponding to one of the
+             predefined model choices (e.g., "1", "2", ..., "10").
+
+    """
+    return Prompt.ask(
+        f"Select a Model for {config_agent_name}:\n"
```
pchalasani, line 62: for some reason , even when I select "same LLM for all agents", it asks me twice "Select a Model for main LLM:"

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,399 @@
...
+        logger.info("Autonomous Debate Selected")
+        interactive_setting = False
+    else:
+        logger.info("Manual Debate Selected with an AI Agent")
+        interactive_setting = True
+        user_input: str = Prompt.ask(
+            "Your argument (or type 'f' for feedback, 'done' to end):"
+        )
+        user_agent.llm = None  # User message without LLM completion
+        user_agent.user_message = user_input
```
pchalasani, line 368: this does not get used -- when you call `user_task.run(x)`, then `x` is used as the first user-role msg

examples/multi-agent-debate/README.md
```diff
@@ -0,0 +1,107 @@
...
+       export OPENAI_API_KEY=OPEN AI KEY
+       export GEMINI_API_KEY=GEMiNi API KEY
+       export METAPHOR_API_KEY=METAPHOR_API_KEY
+4. Please read the following page for more information: https://langroid.github.io/langroid/quick-start/setup/
+
+Usage
+-----
+Run the CLI Application
+Start the application with:
+   python main.py
```
pchalasani, line 74: Suggest making this runnable from the root of the repository, 

```bash
python examples/multi-agent-debate/main.py
```
When I run from root, I get an err saying `generation_config.json` not found.

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,399 @@
...
+    # Determine user's side and assign user_agent and ai_agent based on user selection
+    agents = {"pro": (pro_agent, con_agent, "Pro", "Con"), "con": (con_agent, pro_agent, "Con", "Pro")}
+    user_agent, ai_agent, user_side, ai_side = agents[side]
+    logger.info(
+        f"Starting debate on topic: {topic_name}, taking the {user_side} side. "
+        f"LLM Delegate: {llm_delegate}"
+    )
+
+    # Clear Agents memory
+    user_agent.clear_history()
```
pchalasani, line 353: I don't think `clear_history()` is needed anywhere, since we're not re-running the agents/tasks in same session

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,399 @@
...
+    # Generate feedback summary and declare a winner using feedback agent
+    validation_message: List = last_agent.message_history
+    if not validation_message:
+        logger.warning("No agent message history found for the last agent")
+        validation_message = "No last agent history available to validate."
+
+    feedback_agent.clear_history()
+    feedback_agent.enable_message(MetaphorSearchTool)
+    feedback_agent_task = Task(feedback_agent, interactive=False)
+    feedback_agent_task.run(validation_message)
```
pchalasani, line 389: Since there is no clear done-condition, this gets stuck in a loop, and exits after 5 steps. The `validation_message` must include instructions to say DONE after Step 3.

examples/multi-agent-debate/models.py
```diff
@@ -0,0 +1,23 @@
...
+    Attributes:
+        topic (str): The topic of the message.
+        message (str): The content of the message.
+    """
+    topic: str
+    message: str
+
+
+class SystemMessages(BaseModel):
+    """Represents a collection of system messages.
```
pchalasani, line 17: I suggest also including instructions to limit the responses of both sides to max 3 points expressed as single sentences (or something reasonable) so it's easy for people to try it out without having to go through a lot of slop. You can add instrucs in the readme saying this is configurable and people can increase it to see a realistic debate.

examples/multi-agent-debate/main.py
adam-shams: done

examples/multi-agent-debate/main.py
adam-shams: Done

examples/multi-agent-debate/config.py
adam-shams: Done

examples/multi-agent-debate/main.py
adam-shams: Done

examples/multi-agent-debate/main.py
adam-shams: Done

examples/multi-agent-debate/README.md
adam-shams: Done

adam-shams:
@pchalasani Ready for review. Thanks again for all the feedback!

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,407 @@
...
+    # Generate feedback summary and declare a winner using feedback agent
+    validation_message: List = last_agent.message_history
+    if not validation_message:
+        logger.warning("No agent message history found for the last agent")
+        validation_message = "No last agent history available to validate."
+
+    feedback_agent.enable_message(MetaphorSearchTool)
+    feedback_agent.enable_message(DoneTool)
+    feedback_agent_task = Task(feedback_agent, interactive=False)
+    feedback_agent_task.run(validation_message)
```
pchalasani, line 397: I forgot to mention one thing here - the message history will contain the system message at the start, causing confusion to the feedback agent. You may want to add code to create a nicely annotated text from each side of the argument, by going thru the message_history. Something like

```
Pro: ... 

Con: ...

Pro:  ...

Con: ...
```

I'm adding a new experimental langroid feature later to langroid that adds a notion of an agent "listening" to one or more other agents and accumulating context so this type of manual stitching is avoided. This would be much like the bluesky social-network where people "follow" others. Your work made me realize there's this limitation in langroid that an agent can only get message from a **sinle** "parent" or peer task.
But that will be added later, so for now you'll need to this manually.

pchalasani (changes requested):

examples/multi-agent-debate/main.py
adam-shams: Thanks! This was a good idea. Ready for review

examples/multi-agent-debate/config.py
```diff
@@ -0,0 +1,134 @@
...
+        "1: gpt-4o\n"
+        "2: gpt-4\n"
+        "3: gpt-4o-mini\n"
+        "4: gpt-4-turbo\n"
+        "5: gpt-4-32k\n"
+        "6: gpt-3.5-turbo-1106\n"
+        "7: Mistral: mistral:7b-instruct-v0.2-q8_0a\n"
+        "8: Gemini: gemini-1.5-flash\n"
```
pchalasani, line 70: may want to switch to 2.0-flash

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,451 @@
...
+        pro_agent_config: OpenAIGPTConfig = get_base_llm_config("for Pro Agent")
+        con_agent_config: OpenAIGPTConfig = get_base_llm_config("for Con Agent")
+        feedback_agent_config: OpenAIGPTConfig = get_base_llm_config("feedback", temperature=0.2)
+
+    system_messages: SystemMessages = load_system_messages("examples/multi-agent-debate/system_messages.json")
+    topic_name, pro_key, con_key, side = select_topic_and_setup_side(system_messages)
+
+    # Prompt for number of debate turns
+    max_turns: int = int(
+        Prompt.ask(f"How many turns should the debate continue for? (Defalut is {DEFAULT_TURN_COUNT})",
```
pchalasani, line 347: typo `defalut`

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,451 @@
...
+
+    # Generate feedback summary and declare a winner using feedback agent
+    validation_message: List = last_agent.message_history
+    if not validation_message:
+        logger.warning("No agent message history found for the last agent")
+        validation_message = "No last agent history available to validate."
+
+    feedback_agent.enable_message(MetaphorSearchTool)
+    feedback_agent.enable_message(DoneTool)
+    feedback_agent_task = Task(feedback_agent, interactive=False)
```
pchalasani, line 433: not used (overridden later)

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,451 @@
...
+
+    # Determine the last agent based on turn count and alternation
+    # Note: user_agent and ai_agent are dynamically set based on the chosen user_side
+    last_agent = ai_agent if max_turns % 2 == 0 else user_agent
+
+    # Generate feedback summary and declare a winner using feedback agent
+    validation_message: List = last_agent.message_history
+    if not validation_message:
+        logger.warning("No agent message history found for the last agent")
+        validation_message = "No last agent history available to validate."
```
pchalasani, line 429: not used

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,451 @@
...
+    feedback_agent.enable_message(MetaphorSearchTool)
+    feedback_agent.enable_message(DoneTool)
+    feedback_agent_task = Task(feedback_agent, interactive=False)
+
+    formatted_history = parse_and_format_message_history(last_agent.message_history)
+    logger.info(f"Formatted Message History:\n{formatted_history}")
+
+    # Pass formatted history to the feedback agent
+    feedback_agent_task = Task(feedback_agent, interactive=False)
+    feedback_agent_task.run([formatted_history])
```
pchalasani, line 440: pass history directly, not list

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,451 @@
...
+            "Your argument (or type 'f' for feedback, 'done' to end):"
+        )
+        user_agent.llm = None  # User message without LLM completion
+        user_agent.user_message = user_input
+
+    # Set up langroid tasks and run the debate
+    user_task = Task(user_agent, interactive=interactive_setting, restart=False)
+    ai_task = Task(ai_agent, interactive=False, single_round=True)
+    user_task.add_sub_task(ai_task)
+    if llm_delegate:
```
pchalasani, line 416: flipped -- needs `not llm_delegate` (fails with gemini in delegate mode since user_message is None)

pchalasani (changes requested):
Tried with gemini 1.5 pro, the feedback agent ends prematurely due to `max_output_tokens` param default value of 1024 tokens -- you can see that in the output log:

image


And it doesn't get a chance to output the DoneTool so the agent exits after being stuck for 5 steps. 

To make it easier for people to try - I suggesting setting `max_output_tokens=5000` wherever necessary. And in various prompts have the agents  output at most 2 points or similar. The final task of the feedback agent is a bit complex and smaller models struggle -- it has to first give feedback, then web search, then give its own arguments. You could simplify this final task if you want, or leave as is.

examples/multi-agent-debate/main.py
adam-shams: fixed!

examples/multi-agent-debate/main.py
adam-shams: fixed

examples/multi-agent-debate/main.py
adam-shams: done

examples/multi-agent-debate/main.py
adam-shams: done

adam-shams:
@pchalasani Ready for review. I refactored the feedback message and created a new agent to do MetaphorSearch since not everyone has a MetaphorSearch key so you can still get feedback. I increased the max_output_tokens to 10k but still see it getting stuck after 5 steps even with gpt-4o
Screenshot 2024-12-25 at 5 39 43 PM
Screenshot 2024-12-25 at 5 41 27 PM
It seems to complete the tasks despite this message. Thanks!

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,504 @@
...
+    """Prompt the user to decide on LLM delegation.
+
+    Asks the user whether the LLM should autonomously continue the debate
+    without requiring user input.
+
+    Returns:
+        bool: True if the user chooses LLM delegation, otherwise return False.
+    """
+    return Confirm.ask(
+        "Do you have the Metaphor Search API Key set, and should I search to provide pro and con summaries for your topic?",
```
pchalasani, line 304: simplify this question -- too long

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,504 @@
...
+    topic_name, pro_key, con_key, side = select_topic_and_setup_side(system_messages)
+
+    # Prompt for number of debate turns
+    max_turns: int = int(
+        Prompt.ask(f"How many turns should the debate continue for? (Default is {DEFAULT_TURN_COUNT})",
+                   default=DEFAULT_TURN_COUNT)
+    )
+
+    # Generate the system message
+    metaphor_search_agent_system_message = generate_metaphor_search_agent_system_message(system_messages, pro_key,
```
pchalasani, line 426: avoid long vbl and fn names

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,504 @@
...
+
+from config import get_base_llm_config, get_global_settings
+from models import SystemMessages, Message
+
+DEFAULT_SYSTEM_MESSAGE_ADDITION = f"""
+            DO NOT REPEAT ARGUMENTS THAT HAVE BEEN PREVIOUSLY GENERATED 
+            AND CAN BE SEEN IN THE DEBATE HISTORY PROVIDED. 
+            """
+FEEDBACK_AGENT_SYSTEM_MESSAGE = f"""  
+            STEP 1: Provide Thorough Debate Feedback
```
pchalasani, line 24: there's only 1 step, so no need to say STEP 1

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,504 @@
...
+    # Determine the last agent based on turn count and alternation
+    # Note: user_agent and ai_agent are dynamically set based on the chosen user_side
+    last_agent = ai_agent if max_turns % 2 == 0 else user_agent
+
+    # Generate feedback summary and declare a winner using feedback agent
+
+    if not last_agent.message_history:
+        logger.warning("No agent message history found for the last agent")
+
+    feedback_agent_task = Task(feedback_agent, interactive=False)
```
pchalasani, line 486: set `single_round=True` so we avoid the "stuck for 5 steps warning" -- this ends the task as soon as the agent LLM gives feedback

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,504 @@
...
+            4. Rebuttals: Analyze how well each side refutes their opponent's arguments.
+            5. Persuasion: Assess communication quality, tone, rhetorical effectiveness, and emotional/ethical appeals.
+            6. Technical Execution: Identify if major arguments were addressed or dropped and check consistency.
+            7. Debate Etiquette: Evaluate professionalism, respect, and demeanor.
+            8. Final Focus: Judge the strength of closing speeches, how well they summarize the case, 
+            and justify a winner.
+            Provide constructive feedback for each debater, 
+            summarizing their performance and declaring a winner with justification.   
+            """
+METAPHOR_SEARCH_AGENT_SYSTEM_MESSAGE_TEMPLATE = """
```
pchalasani, line 41: I don't think this is giving sufficient context to the metaphor agent -- we need a preamble stating the background, i.e. that the task is to find good references for arguing two sides of a debate.

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,504 @@
...
+                     processes, emphasizing the need for autonomy and ethical oversight.
+        
+            M2: SOURCE: ...
+            EXTRACT: ...
+            SUMMARY:
+        
+            DO NOT MAKE UP YOUR OWN SOURCES; ONLY USE SOURCES YOU FIND FROM A WEB SEARCH. 
+            ENSURE STEP 2 IS COMPLETED BEFORE STARTING STEP 3
+            
+            STEP 3: Argue Pro and Con Cases
```
pchalasani, line 65: STEP 2 since there are only 2.

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,504 @@
...
+            As an expert debater, your goal is to eloquently argue for both the Pro and Con cases using the web-search sources generated in Step 2.
+            Write at least 10 sentences for each side.
```
pchalasani, line 67: 5 sentences

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,504 @@
...
+        
+            M2: SOURCE: ...
+            EXTRACT: ...
+            SUMMARY:
+        
+            DO NOT MAKE UP YOUR OWN SOURCES; ONLY USE SOURCES YOU FIND FROM A WEB SEARCH. 
+            ENSURE STEP 2 IS COMPLETED BEFORE STARTING STEP 3
+            
+            STEP 3: Argue Pro and Con Cases
+            As an expert debater, your goal is to eloquently argue for both the Pro and Con cases using the web-search sources generated in Step 2.
```
pchalasani, line 66: Step 2 -> Step 1

examples/multi-agent-debate/main.py
pchalasani, line 1: This file has become too big. For readability I suggest -- 

- move the user-prompt stuff to utils.py
- move sys msgs to system_messages.py

You want people looking at your code to quickly understand what is going on.

examples/multi-agent-debate/system_messages.json
pchalasani, line 1: At least in PyCharm the system msgs are shown in a single super-long line, without wrapping.
Maybe use triple-quoted multiline strings

examples/multi-agent-debate/main.py
pchalasani, line 426: avoid long vbl and fn names. One suggestion `x_agent` -> `x_task` instead of `x_agent_task`.

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,504 @@
...
+
+    feedback_agent_task = Task(feedback_agent, interactive=False)
+    formatted_history = parse_and_format_message_history(last_agent.message_history)
+    feedback_agent_task.run(formatted_history)  # Pass formatted history to the feedback agent
+
+    if metaphor_search:
+        metaphor_search_agent_task = Task(metaphor_search_agent, interactive=False)
+        metaphor_search_agent.enable_message(MetaphorSearchTool)
+        metaphor_search_agent.enable_message(DoneTool)
+        metaphor_search_agent_task.run("run the search")
```
pchalasani, line 494: this agent gets stuck after:

```
- LLM -> search tool
-  agent_response -> results
- LLM -> final debate arguments with citations
```
It ends up working fine since langroid has a stalled loop check. But you can avoid that by having a `handle_message_fallback` method defined in a `MetaphorSearchAgent` subclassed from `ChatAgent`. You can use a variant of the definition here -- 
https://langroid.github.io/langroid/FAQ/#how-can-i-handle-an-llm-forgetting-to-generate-a-toolmessage

In your case you can just return `AgentDoneTool` to end the task, e.g.

```python
from langroid.agent.tools.orchestration import AgentDoneTool
def handle_message_fallback(
        self, msg: str | ChatDocument
    ) -> str | ChatDocument | None:
        """Handle scenario where LLM did not generate any Tool"""
        if isinstance(msg, ChatDocument) and msg.metadata.sender == Entity.LLM:
            return AgentDoneTool(content="done")
        return None

```

(Search for `handle_fallback_method` in `examples/` folder)

examples/multi-agent-debate/main.py
adam-shams: fixed

examples/multi-agent-debate/main.py
adam-shams: fixed

examples/multi-agent-debate/main.py
adam-shams: done

examples/multi-agent-debate/main.py
adam-shams: done

examples/multi-agent-debate/main.py
adam-shams: done

examples/multi-agent-debate/main.py
adam-shams: done

examples/multi-agent-debate/main.py
adam-shams: done

examples/multi-agent-debate/system_messages.json
adam-shams: done in system_messages.py

examples/multi-agent-debate/main.py
adam-shams: done

examples/multi-agent-debate/main.py
adam-shams: fixed. thank you!

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,254 @@
...
+        logger.warning("No agent message history found for the last agent")
+
+    feedback_task = Task(feedback_agent, interactive=False, single_round=True)
+    formatted_history = parse_and_format_message_history(last_agent.message_history)
+    feedback_task.run(formatted_history)  # Pass formatted history to the feedback agent
+
+    metaphor_search: bool = is_metaphor_search_key_set()
+    if metaphor_search:
+
+        def handle_message_fallback(
```
pchalasani, line 233: not needed

pchalasani (changes requested):
When I ran it with gpt4o it, the metaphor agent got references, then LLM showed the references, but then did not do the final step of arguing the pro and con sides with the references. Overall I think the task of the metaphor is not crisply defined. You may want to clearly specify the task at the top of the sys msg so that it actually completes both steps. Or you can simplify this agent's task to simply get useful references. 

And then run a couple times with gpt4o to make sure it's doing what you want.

Also there are several unused imports etc. You can run `ruff check <filename> --fix` to auto-fix some of these. 


image

adam-shams:
@pchalasani Thanks for the feedback. I fixed the prompt and I'm getting consistent results on gpt-4o with metaphor search 
MetaphoreSearchAgentResult

examples/multi-agent-debate/main.py
```diff
@@ -0,0 +1,239 @@
...
+)
+
+
+class MetaphorSearchChatAgent(ChatAgent):
+    def handle_message_fallback(
+            self, msg: str | ChatDocument
+    ) -> AgentDoneTool | None:
+        """Handle scenario where LLM did not generate any Tool"""
+        if isinstance(msg, ChatDocument) and msg.metadata.sender == Entity.LLM:
+            return AgentDoneTool(content="done")
```
pchalasani, line 41: When I ran with gpt4o it again got refs but did not argue the 2 sides. 
One way to handle that is to change this line to return a message:

```python
return f"Have you presented pro and con arguments based on your search results? If so, use the TOOL `{DoneTool.name()}` to indicate you're finished. Otherwise, argue both sides and then send the `{DoneTool.name()}`
```

pchalasani: the LLM will see this "nudge" and respond accordingly

adam-shams: I changed the return message and tested the output. Screenshot attached 
DoneTool_return_message

pchalasani (approved):
Great work @adam-shams I can merge this.

pchalasani:
@adam-shams there are lint errors, could you please make a new PR addressing those?
You have to run `make check` and it will show the errors that weren't auto-fixable.

PR #630, opened 2024-12-05
Async methods

nilspalumbo:
Duplicate of  #610.


langroid/agent/base.py
```diff
@@ -322,6 +323,52 @@ def _get_tool_list(
...
             )
 
+        async_tool_name = f"{tool}_async"
+        if (
+            hasattr(message_class, "handle_async")
+            and inspect.isfunction(message_class.handle_async)
+            and not hasattr(self, async_tool_name)
+        ):
+            has_chat_doc_arg = (
+                len(inspect.signature(message_class.handle_async).parameters) > 2
```
pchalasani, line 333: shouldn't this be > 1 just like in the "handle" case?

pchalasani (changes requested):
This was a very carefully done PR, thanks @alexagr and thanks @nilspalumbo for reviewing!
I had just one comment on the code. Once resolved I'll merge

langroid/agent/base.py
alexagr: @pchalasani I think you are right, but it's not something that I've changed; this line was modified on 2028/08/11 by you as part of #539.

pchalasani: Right, there are two cases of `has_doc_chat_arg`, one for `handle` and one for `response`, the first one checks for > 1 param and the second one checks for > 2 params. In the **async** version you added, you're using > 2 in both cases. I'll push my change and merge. Thanks @alexagr

nilspalumbo: Yes, you're right, thanks for catching this!

Issue #627, opened 2024-12-01
Custom tool handler method


alexagr:
**Is your feature request related to a problem? Please describe.**

Langroid currently assumes that tool handlers are class methods with names identical to tool's name (`request` attribute).
This is somewhat unsafe - as tool name may be identical to an existing method of Agent class.
Also for use-cases where tools are defined dynamically (e.g. based on some "configuration data" stored in external database) this requires creation of multiple dynamic methods in Agent object - which may be a challenge for garbage collection.

**Describe the solution you'd like**
Add a new optional "handler" attribute to `ToolMessage`:
```
class ToolMessage(ABC, BaseModel):
    ...
    handler: Optional[str] = None
```

When the `handler` attribute is not `None`, use it to find the handler method instead of default `request` attribute value.

The signature of the `handler` method will be as follows:
```
def handler(self, msg: ToolMessage, request: str)
```
Use of `request` parameter (name of the tool) will allow creation of a single "tool handler" for multiple tools.

**Describe alternatives you've considered**
I initially thought about simply adding "tool_" prefix to the default tool handler - but this is a "breaking change" and thus IMO not a good idea.

pchalasani:
@alexagr thanks for this suggestion. 

Suppose we have

```python
class MyTool(ToolMessage):
    request: str = "my_tool"
    handler: str|None = "my_handler"
```

Are you saying in this case we would handle this by runing `agent.my_handler(msg: MyTool, request="my_tool")` ?

If that's not what you had in mind can you create the right sample code?

alexagr:
@pchalasani yes, exactly

PR #633, opened 2024-12-07
custom tool handler

alexagr:
Implement "custom tool handler" method - as described in issue #627.

Sample usage:
```
class MyToolMessage(ToolMessage):
    request: str = "my_tool"
    handler: str = "tool_handler"

class MyAgent(ChatAgent):
    def tool_handler(self, message: ToolMessage, tool_name: str, chat_doc: ChatDocument = None) -> str:
        if tool_name == "my_tool":
            # do something

```
I've also modified test_tool_messages and test_tool_messages_async to use the new "custom handler" for some tools.

### DIFF
--- a/langroid/agent/base.py
+++ b/langroid/agent/base.py
@@ -1760,20 +1760,11 @@ class Agent(ABC):
             chat_doc is not None
             and "chat_doc" in inspect.signature(handler_method).parameters
         )
-        has_tool_name_arg = ("tool_name" in
-                             inspect.signature(handler_method).parameters)
         try:
-            if has_tool_name_arg:
-                if has_chat_doc_arg:
-                    maybe_result = await handler_method(tool, tool_name=tool_name,
-                                                        chat_doc=chat_doc)
-                else:
-                    maybe_result = await handler_method(tool, tool_name=tool_name)
+            if has_chat_doc_arg:
+                maybe_result = await handler_method(tool, chat_doc=chat_doc)
             else:
-                if has_chat_doc_arg:
-                    maybe_result = await handler_method(tool, chat_doc=chat_doc)
-                else:
-                    maybe_result = await handler_method(tool)
+                maybe_result = await handler_method(tool)
             result = self.to_ChatDocument(maybe_result, tool_name, chat_doc)
         except Exception as e:
             # raise the error here since we are sure it's
@@ -1812,20 +1803,11 @@ class Agent(ABC):
             chat_doc is not None
             and "chat_doc" in inspect.signature(handler_method).parameters
         )
-        has_tool_name_arg = ("tool_name" in
-                             inspect.signature(handler_method).parameters)
         try:
-            if has_tool_name_arg:
-                if has_chat_doc_arg:
-                    maybe_result = handler_method(tool, tool_name=tool_name,
-                                                  chat_doc=chat_doc)
-                else:
-                    maybe_result = handler_method(tool, tool_name=tool_name)
+            if has_chat_doc_arg:
+                maybe_result = handler_method(tool, chat_doc=chat_doc)
             else:
-                if has_chat_doc_arg:
-                    maybe_result = handler_method(tool, chat_doc=chat_doc)
-                else:
-                    maybe_result = handler_method(tool)
+                maybe_result = handler_method(tool)
             result = self.to_ChatDocument(maybe_result, tool_name, chat_doc)
         except Exception as e:
             # raise the error here since we are sure it's
--- a/langroid/agent/base.py
+++ b/langroid/agent/base.py
@@ -278,17 +278,27 @@ class Agent(ABC):
         if not issubclass(message_class, ToolMessage):
             raise ValueError("message_class must be a subclass of ToolMessage")
         tool = message_class.default_value("request")
+
+        """
+        if tool has handler method explicitly defined - use it,
+        otherwise use the tool name as the handler
+        """
+        if hasattr(message_class, "handler"):
+            handler = message_class.default_value("handler")
+        else:
+            handler = tool
+
         self.llm_tools_map[tool] = message_class
         if (
             hasattr(message_class, "handle")
             and inspect.isfunction(message_class.handle)
-            and not hasattr(self, tool)
+            and not hasattr(self, handler)
         ):
             """
             If the message class has a `handle` method,
-            and agent does NOT have a method with the same name as the tool,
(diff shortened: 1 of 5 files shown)


langroid/agent/base.py
```diff
@@ -1738,18 +1749,31 @@ async def handle_tool_message_async(
...
             and "chat_doc" in inspect.signature(handler_method).parameters
         )
+        has_tool_name_arg = ("tool_name" in
+                             inspect.signature(handler_method).parameters)
         try:
-            if has_chat_doc_arg:
-                maybe_result = await handler_method(tool, chat_doc=chat_doc)
+            if has_tool_name_arg:
+                if has_chat_doc_arg:
+                    maybe_result = await handler_method(tool, tool_name=tool_name,
```
pchalasani, line 1768: `tool_name` is a direct function of `tool`, so it seems redundant to have this arg I think? See my comment in the test

tests/main/test_tool_messages_async.py
```diff
@@ -76,8 +77,16 @@ class MessageHandlingAgent(ChatAgent):
     def file_exists(self, message: FileExistsMessage) -> str:
         return "yes" if message.filename == "requirements.txt" else "no"
 
-    def python_version(self, PythonVersionMessage) -> str:
-        return DEFAULT_PY_VERSION
+    def tool_handler(
+        self,
+        message: ToolMessage,
+        tool_name: str,
```
pchalasani, line 83: since we are intending `tool_handler` to be specifically for the `PythonVersionMessage` tool, why introduce `tool_name`? Do you see any scenario where we would want to have this arg?

alexagr: @pchalasani you are absolutely right! Sorry, my bad. Pushed a fixed version.

langroid/agent/base.py
alexagr: @pchalasani you are absolutely right! Sorry, my bad. Pushed a fixed version.

langroid/agent/base.py
```diff
@@ -278,17 +278,27 @@ def _get_tool_list(
         if not issubclass(message_class, ToolMessage):
             raise ValueError("message_class must be a subclass of ToolMessage")
         tool = message_class.default_value("request")
+
+        """
+        if tool has handler method explicitly defined - use it,
+        otherwise use the tool name as the handler
+        """
+        if hasattr(message_class, "handler"):
```
pchalasani, line 286: we should use `_handler` instead of `handler` since we don't want `handler` to be part of the schema passed to the LLM (i.e. we don't want LLM to generate `handler` field). 

Also a note on this should be added to the `chat-agent-tool.md` in the quick-start docs. But let's wait until we know we really need this feature (see comment later)

pchalasani (changes requested):
@alexagr thanks for this PR. I wanted to revisit the motivation for this, which I think I'm not fully understanding. 

Pasting from the related issue, you wrote:

> Langroid currently assumes that tool handlers are class methods with names identical to tool's name (request attribute).
This is somewhat unsafe - 
> - [1] as tool name may be identical to an existing method of Agent class. 
> - [2] Also for use-cases where tools are defined dynamically (e.g. based on some "configuration data" stored in external database) this requires creation of multiple dynamic methods in Agent object - which may be a challenge for garbage collection.

Let's take point [1] - say I have an agent A with method `blah`, then when we create a tool T, we **know** we shouldn't set `request` to `blah`. Also, say there's an existing tool T with method `blah`, then when we create agents A1, A2 to use T, we'll have to take care not to have methods `blah`. I think you're saying these considerations of avoiding method-name-clash are a burden, so we use `handler` method to avoid this? But the exact same considerations apply to `handler`, right?

Point [2] about dynamic tool definitions. Here also whatever concerns there are about name-clash with `request` seem to apply to `handler`.

Overall, I understand you want to separate the notion of "tool name" from "tool handler". I'd like to know a scenario you have in mind where we'd leverage that. 

@nilspalumbo please weigh in when you get a chance.

nilspalumbo:
@alexagr I need a bit more clarification as well.

I assume the intention for [1] is that, if there is a name clash, renaming the handler method and `handler` for all tools is simpler than changing `request` and any handler methods defined on the agent for all tools?

Is the goal for [2] that, rather than dynamically redefining a handler method on the agent, to instead redefine some `handler` method as tools are enabled and disabled, allowing the earlier iterations of `handler` to be garbage collected?

Currently, we do not delete the handler methods created when a tool is enabled, even if handling is later disabled, as the memory usage of handler methods will be negligible in typical use-cases. As the size of the required `handler` method would grow with the function count, the potential memory savings are limited. Perhaps this might become relevant with a combination of batch processing and large-scale automated tool generation; is this your intended use case?

nilspalumbo:
Use of union types in a single tool definition already enables much of this functionality (see the `_get_any_tool_message` method of `ChatAgent`).

alexagr:
@pchalasani @nilspalumbo I was referring to a possible clash with base `Agent` class methods - e.g. `agent_response`. 

But let me describe my use-case - as it'll be easier to understand where I'm coming from. I'm writing a Web portal where end-users define agents (prompts / LLMs / ...), tools (URLs / authentication / schema) and upload documents. I'm using langroid as an "engine" to run the chat sessions following these definitions. With the current tools implementation I can't simply create `MyAgent` class with some generic `tool_handler` method - instead I have to create classes dynamically for every session (because configuration changes all the time). Also I can't just use the tool name "as provided by user" - due to the possible naming clash - instead I'm currently adding "tools_" prefix to it - which is pretty messy, since this modified tool name appears in "agent logs" that I expose to users for troubleshooting purposes. Also I'm not sure how well garbage collector cleans up all this mess - though I may be wrong here...

Bottom line - I think that (optional) separation between tool name (as "exposed" to LLM) and handler name (that is "dictated" by underlying code structure) could be quite useful.

pchalasani:
@alexagr I discussed with @nilspalumbo and we decide to proceed with this, since it's a minor change, albeit a bit custom. It will help if you could please:

- add a short illustrative example for how it helps to use the `_handler` attribute in a `ToolMessage`, and 
- add a brief note at `docs/notes/tool_message_handlers.md` and link the example to this note.

alexagr:
@pchalasani thank you - done

pchalasani:
Thanks @alexagr that is a nice illustrative example! Closing as replaced by #643  (I added some sentences to the docs)

PR #634, opened 2024-12-10
fix: invalid default for start_llm_stream_async

johannestang:
Running `task.run_async(...)` fails if `async_stream_quiet=False` because the default for `callbacks.start_llm_stream_async` doesn't return a function as it should, so an exception is raised when it is called (which only happens if `async_stream_quiet=False`).


pchalasani:
Good catch, thank you @johannestang 
I added some fixes on top of yours in #636, so closing this

PR #635, opened 2024-12-10
fix: better handling of errors when parsing func args (non-streaming)

johannestang:
Currently, parsing function args of type `bool` fails when `stream=False`. The reason is that `LLMFunctionCall.from_dict` uses `ast.literal_eval` which fails as booleans are lowercase in JSON but must be capitalized in Python. In streaming mode this does not happen as `parse_imperfect_json` is used which uses `repair_json` as a fallback when `ast.literal_eval` fails. This PR proposes to also use `parse_imperfect_json` in the non-streaming case.


pchalasani:
Thanks @johannestang - I added a differentiating test (i.e. fails before PR, passes after) in #637, so closing this. Also tests can't be run from a fork (since gh secrets like OpenAI keys aren't shared), so tests can only be run from main fork.

PR #638, opened 2024-12-11
Add Llama.cpp Embeddings Support

Kwigg:
This PR adds support for using llama.cpp server as a provider of embeddings.

Although llama.cpp server provides an OpenAI compatible endpoint for any embeddings model, tiktoken is often not the right tokeniser so it's not possible to use the OpenAIEmbeddings pathway.

In order to account for how the tokeniser changes based on the model, the implementation uses llama.cpp server's endpoints for tokenisation as opposed to trying to handle it locally - it'll use a lot more API calls, but the benefit of being locally hosted is this shouldn't really matter!

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.8.3
+  rev: v0.8.4
   hooks:
     - id: ruff
--- a/docs/notes/llama-cpp-embeddings.md
+++ b/docs/notes/llama-cpp-embeddings.md
@@ -1,0 +1,30 @@
+# Local embeddings provision via llama.cpp server
+
+As of Langroid v0.X.X, you can use llama.cpp as provider of embeddings
+to any of Langroid's vector stores, allowing access to a wide variety of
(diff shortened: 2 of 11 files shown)


pchalasani:
Thanks @Kwigg! Will look in the coming days

pchalasani:
@Kwigg could you please:
- make a simple example script at `examples/docqa/chat-llama-cpp-emdbed.py` with basic instructions at the top, that demonstrates the usage of this, or alternatively, add this as an optional setting in `examples/docqa/chat.py` with basic instructions at the top, 
- a short doc page at `docs/notes/llama-cpp-embeddings.md` pointing to this example,
- update the nav section in `mkdocs.yaml` to refer to this doc page

pchalasani (commented):
Requesting tests, examples, docs as in the comments

Kwigg:
I believe I've made the requested changes, please could you have a look over them? 

I also refactored the embedding models.py, as I noticed that only OpenAI embeddings functioned in the examples/docqa/chat.py example with a ChromaDB vector store. Previously, ChromaDB would complain that the function didn't match its expected function signature and throw up errors. This change should fix that!

pchalasani (changes requested):
Thanks @Kwigg 
I'll review once you've added pytest-based tests in  `tests/extras/llama-cpp-embeddings.py`

pchalasani:
Thank you @Kwigg  -- fixed a minor conflict and created PR #648 which duplicates this, and also enables running tests. Will merge once passed.

pchalasani:
Closing, since dupe #648 was merged.

Issue #534, opened 2024-07-25
Add Azure Token Provider setting


svanschalkwyk:
My Azure subscription requires an Azure Token to enable access to my models. 
The Azure configuration in the examples and the relevant class do not provide that. 

What we need is something similar to the OpenAI in pypi: 
`class AzureOpenAI(BaseAzureClient[httpx.Client, Stream[Any]], OpenAI):
    @overload
    def __init__(
        self,
        *,
        azure_endpoint: str,
        azure_deployment: str | None = None,
        api_version: str | None = None,
        api_key: str | None = None,
        azure_ad_token: str | None = None,
        azure_ad_token_provider: AzureADTokenProvider | None = None,
        organization: str | None = None,
        timeout: float | Timeout | None | NotGiven = NOT_GIVEN,
        max_retries: int = DEFAULT_MAX_RETRIES,
        default_headers: Mapping[str, str] | None = None,
        default_query: Mapping[str, object] | None = None,
        http_client: httpx.Client | None = None,
        _strict_response_validation: bool = False,
    ) -> None:
`

angsabrina:
Running into the same issue, would love to see this change!

sanny12444:
hi

sanny12444:
[](url)https://github.com/langroid/langroid/issues/534#issue-2431060061

johannestang:
Release [0.29.0](https://github.com/langroid/langroid/releases/tag/0.29.0) addresses this. Have a look at [https://langroid.github.io/langroid/notes/custom-azure-client/](https://langroid.github.io/langroid/notes/custom-azure-client/).

PR #645, opened 2024-12-16
azure openai: bring your own client

johannestang:
I'm facing the same problem as in issue #534, this is how I've addressed it in my fork. Instead of adding more options to the AzureConfig class, I've added an option for providing custom clients.

Example:
```python
import os

import azure.identity as azure_identity
import azure.identity.aio as azure_identity_async
from dotenv import load_dotenv
from openai import AsyncAzureOpenAI, AzureOpenAI

import langroid as lr
import langroid.language_models as lm

load_dotenv()


def get_azure_openai_client():
    return AzureOpenAI(
        api_version="2024-02-15-preview",
        azure_endpoint=os.environ["AZURE_OPENAI_API_BASE"],
        azure_ad_token_provider=azure_identity.get_bearer_token_provider(
            azure_identity.DefaultAzureCredential(),
            "https://cognitiveservices.azure.com/.default",
        ),
    )


def get_azure_openai_async_client():
    return AsyncAzureOpenAI(
        api_version="2024-02-15-preview",
        azure_endpoint=os.environ["AZURE_OPENAI_API_BASE"],
        azure_ad_token_provider=azure_identity_async.get_bearer_token_provider(
            azure_identity_async.DefaultAzureCredential(),
            "https://cognitiveservices.azure.com/.default",
        ),
    )


lm_config = lm.AzureConfig(
    azure_openai_client_provider=get_azure_openai_client,
    azure_openai_async_client_provider=get_azure_openai_async_client,
)

if __name__ == "__main__":
    agent = lr.ChatAgent(lr.ChatAgentConfig(llm=lm_config))
    task = lr.Task(agent, interactive=False)
    task.run("Who is the president of the United States? Reply and end with DONE")
```

For me, this implementation has an added benefit: I use [Langfuse](https://github.com/langfuse/langfuse) for observability, and their [integration](https://langfuse.com/docs/sdk/python/decorators) relies on using a custom openai client which I can now easily pass to Langroid. If this is a use-case you also want to support, then it might be worth considering make a similar change to the (non-Azure) openai config class.

### DIFF
--- a/langroid/language_models/azure_openai.py
+++ b/langroid/language_models/azure_openai.py
@@ -1,3 +1,5 @@
+from typing import Callable
+
 from dotenv import load_dotenv
 from httpx import Timeout
 from openai import AsyncAzureOpenAI, AzureOpenAI
@@ -42,6 +44,10 @@ class AzureConfig(OpenAIGPTConfig):
     model_version: str = ""  # is used to determine the cost of using the model
     api_base: str = ""
 
+    # Alternatively, bring your own clients:
+    azure_openai_client_provider: Callable[[], AzureOpenAI] | None = None
+    azure_openai_async_client_provider: Callable[[], AsyncAzureOpenAI] | None = None
+
     # all of the vars above can be set via env vars,
     # by upper-casing the name and prefixing with `env_prefix`, e.g.
     # AZURE_OPENAI_API_VERSION=2023-05-15
@@ -69,20 +75,6 @@ class AzureGPT(OpenAIGPT):
         load_dotenv()
         super().__init__(config)
         self.config: AzureConfig = config
-        if self.config.api_key == "":
-            raise ValueError(
-                """
-                AZURE_OPENAI_API_KEY not set in .env file,
-                please set it to your Azure API key."""
-            )
-
-        if self.config.api_base == "":
-            raise ValueError(
-                """
-                AZURE_OPENAI_API_BASE not set in .env file,
-                please set it to your Azure API key."""
-            )
-
         if self.config.deployment_name == "":
             raise ValueError(
                 """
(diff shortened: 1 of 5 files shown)


langroid/language_models/azure_openai.py
```diff
@@ -98,6 +90,42 @@ def __init__(self, config: AzureConfig):
                 please set it to chat model name in your deployment."""
             )
 
+        if (
+            self.config.azure_openai_client_provider
+            and self.config.azure_openai_async_client_provider
```
pchalasani, line 95: Might be good to separate these two, i.e. if only one of them is specified, use that one for the client. If you think we should enforce that either none or both are specified, it may be good to raise an error with a friendly msg saying that only one of these was specified, and the other must also be specified.

pchalasani (changes requested):
Thanks @johannestang this is really useful. Could you please address my comment and also copy over the example you gave into a script as `examples/basic/chat-azure-client.py`?
Also a short note at `docs/notes/custom-azure-client.md` and corresponding entry in the nav section of `mkdocs.yml` would be nice to have.

angsabrina:
Hi @johannestang will this change work with InteractiveBrowserCredential() parameter for get_bearer_token_provider? Thank you for making this change!

langroid/language_models/azure_openai.py
johannestang: Yeah, that's a good point. I've changed it now so you only need to supply one, but if you only supply one, a warning will be given.

johannestang:
> Hi @johannestang will this change work with InteractiveBrowserCredential() parameter for get_bearer_token_provider? Thank you for making this change!

I don't normally use InteractiveBrowserCredential(), but I just gave it a try and it appears to work 😃

pchalasani:
Thank you @johannestang ! I've fixed a couple mypy errs and made a new PR which also allows tests to be run.

pchalasani:
Closing, replaced by #647

PR #646, opened 2024-12-16
support use of existing embedding model in VectorStore

alexagr:
Some embedding models - e.g. `FastEmbed` - do a LOT during their initialization.
Therefore for multi-tenant / multi-agent environments it makes sense to re-use the same model for all `VectorStore` instances (used by `Agent` instances). However current `VectorStore` implementation doesn't allow this - as it calls `VectorStore.create(config.embedding)` on instance initialization.

This simple patch adds `embedding_model` attribute to `VectorStoreConfig` class. When this attribute is specified, existing embedding model is used during `VectorStore` initialization.

And just to small clarification: while it is possible to use the same `VectorStore` for multiple agents, in many cases it's not appropriate - as each agent has it's own documents/vector-data.

### DIFF
--- a/langroid/vector_store/base.py
+++ b/langroid/vector_store/base.py
@@ -30,6 +30,7 @@ class VectorStoreConfig(BaseSettings):
     embedding: EmbeddingModelsConfig = OpenAIEmbeddingsConfig(
         model_type="openai",
     )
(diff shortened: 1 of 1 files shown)


pchalasani:
Thanks @alexagr !

PR #662, opened 2025-01-05
uv works for installation

### LINKED ISSUE #595: Poetry -> uv to manage deps
We could either replace the poetry instructions with those for uv, or include both:

https://docs.astral.sh/uv/

abab-dev:
@pchalasani I have ported the poetry pyproject.toml to uv. It works for local dev setup. 

Steps to follow 

Install uv using pip if not already.

To create dev environment 
`uv venv --python 3.11 .venv`

`uv sync --group dev`

To sucessfully pass the tests 

`uv sync --group dev  --extra neo4j --extra arango --extra momento --extra sql`

To build 

`uv build`

currently using pdm as build backend 

Few caveats
[project.optional-dependencies](https://docs.astral.sh/uv/concepts/projects/dependencies/#optional-dependencies): Published optional dependencies, or "extras" roughly map to poetry optional dependencies and poetry.extras
[dependency-groups](https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-groups): Local dependencies for development roughly maps to poetry.group 

ran the tests and I get two error as I dont have azure api key and openai api key 

image

also build is successful

Screenshot from 2025-01-05 22-37-29

partially fixes #595


pchalasani:
> Few caveats
> [project.optional-dependencies](https://docs.astral.sh/uv/concepts/projects/dependencies/#optional-dependencies): Published optional dependencies, or "extras" roughly map to poetry optional dependencies and poetry.extras
> [dependency-groups](https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-groups): Local dependencies for development roughly maps to poetry.group

@abab-dev  Thank you! I didn't quite understand what you meant by the caveats? Did you mean the mapping is not "exact" and only "rough" ?

abab-dev:
yes mapping is not exact as I have to copy full dependencies every time i was creating extra group like 
`
hf-transformers = [
    "sentence-transformers<3.0.0,>=2.2.2",
    "torch<3.0.0,>=2.0.0",
    "transformers<5.0.0,>=4.40.1",
    "huggingface-hub<0.22.0,>=0.21.2",
]
`


`
all = [
    "pdf2image<2.0.0,>=1.17.0",
    "pdfplumber<0.11.0,>=0.10.2",
    "pymupdf<2.0.0,>=1.23.3",
    "pypdf>=5.1.0",
    "pytesseract<0.4.0,>=0.3.10",
    "python-docx<2.0.0,>=1.1.0",
    "unstructured[docx,pdf,pptx]<0.10.18,>=0.10.16",
    "sqlalchemy<3.0.0,>=2.0.19",
    "psycopg2<3.0.0,>=2.9.7",
    "pymysql<2.0.0,>=1.1.0",
    "sentence-transformers<3.0.0,>=2.2.2",
    "torch<3.0.0,>=2.0.0",
    "transformers<5.0.0,>=4.40.1",
    "huggingface-hub<0.22.0,>=0.21.2",
    "chromadb<=0.4.23,>=0.4.21",
    "metaphor-python<0.2.0,>=0.1.23",
    "neo4j<6.0.0,>=5.14.1",
    "python-arango",
    "arango-datasets",
    "litellm<2.0.0,>=1.30.1",
    "chainlit<2.0.0,>=1.3.2",
    "python-socketio<6.0.0,>=5.11.0",
    "fastembed<0.4.0,>=0.3.1",
]
`


you can see lot of duplication which is not case with poetry  and 

also if you only want to declare optional package and
not use it with `uv sync --extra package_name` you can just declare under optional dependencies like  ` "litellm<2.0.0,>=1.30.1"`

otherwise poetry.group is compatible with `dependancy-groups` in uv.

pchalasani:
> I have to copy full dependencies every time i was creating extra group

Even in poetry we have to explicitly list all dependencies for each extra group, i.e. if `a` is an extra and `b` is an extra, then we can't simply define "all" as `[a, b]`, i.e. we have to explicitly list the deps in each extra group. 

The only additional burden with `uv` seems to be in the specification of version ranges, since it doesn't accept caret versions like `^1.4.0`, so we have to specify the full acceptable range.

So I guess I'm not clear what you mean by "duplication" here:

> you can see lot of duplication which is not case with poetry and

Also I'm not fully understanding this statement:

> also if you only want to declare optional package and
> not use it with uv sync --extra package_name you can just declare under optional dependencies like  > " litellm<2.0.0,>=1.30.1"

Are you saying that uv allows defining optional dependency `blah` without necessarily having an "extra" associated with it? If that's what you meant then it's not import for us right?

Just trying to make sure I'm very clear what you're saying since this is an important change, thanks for bearing with me :)

pchalasani:
> To build
> 
> `uv build`
> 
> currently using pdm as build backend

When I do `uv build` I see a lot of lines saying `Adding ...` , and this includes lots of useless paths. I don't know if this is normal, and how to prevent it. 

```
Building source distribution...
 - Adding LICENSE -> LICENSE
 - Adding README.md -> README.md
 - Adding langroid/py.typed -> langroid/py.typed
 - Adding .pdm-build/pyproject.toml -> pyproject.toml
 - Adding tests/.DS_Store -> tests/.DS_Store
 - Adding tests/.logs/repo_loader/download_log.json -> tests/.logs/repo_loader/download_log.json
 - Adding tests/.pytest_cache/.gitignore -> tests/.pytest_cache/.gitignore
 - Adding tests/.pytest_cache/CACHEDIR.TAG -> tests/.pytest_cache/CACHEDIR.TAG
 - Adding tests/.pytest_cache/README.md -> tests/.pytest_cache/README.md
 - Adding tests/.pytest_cache/v/cache/lastfailed -> tests/.pytest_cache/v/cache/lastfailed
 [truncated: many more such lines]
```

pyproject.toml
```diff
@@ -1,201 +1,250 @@
...
+    "mkdocs-rss-plugin<2.0.0,>=1.8.0",
+    "mkdocs-gen-files<1.0.0,>=0.4.0",
+    "mkdocs-literate-nav<1.0.0,>=0.6.0",
+    "mkdocs-section-index<1.0.0,>=0.3.5",
+    "mkdocs-jupyter<1.0.0,>=0.24.1",
+    "griffe<1.0.0",
+]
+
+[tool.pdm.build]
+includes = ["pyproject.toml", "langroid/py.typed"]
```
pchalasani, line 244: I think we should add:
```
excludes = ["tests/"]
```

pchalasani (changes requested):
If we move away from poetry, the version-bumping in the Makefile needs to change:

```
.PHONY: bump-patch
bump-patch:
	@poetry version patch
	@git commit pyproject.toml -m "Bump version"

.PHONY: bump-minor
bump-minor:
	@poetry version minor
	@git commit pyproject.toml -m "Bump version"

.PHONY: bump-major
bump-major:
	@poetry version major
	@git commit pyproject.toml -m "Bump version"
```

and 

```
.PHONY: release
release:
	@VERSION=$$(poetry version | cut -d' ' -f2) && gh release create $${VERSION} dist/*
```

pchalasani:
> version-bumping

relevant uv discussion - https://github.com/astral-sh/uv/issues/6298

abab-dev:
I realized I was not clear enough.

> Even in poetry we have to explicitly list all dependencies for each extra group, i.e. if `a` is an extra and `b` is an extra, then we can't simply define "all" as `[a, b]`, i.e. we have to explicitly list the deps in each extra group.

 I was just referring to the fact that poetry lets your declare optional dependencies and you can freely use them just by referring to their name. 

> > you can see lot of duplication which is not case with poetry and
> 
> Also I'm not fully understanding this statement:

In case of uv I had to do lot of diffing get dependencies right which was bad user experience. I was declaring them explicitly every time and there was no way to refer back to them. This kind of breaks the mental model we have with poetry. Except that tbh I do not know their internal working.


> Are you saying that uv allows defining optional dependency `blah` without necessarily having an "extra" associated with it? If that's what you meant then it's not import for us right?

Yes, I get it now. This does not matter, because it is not import.

abab-dev:
> If we move away from poetry, the version-bumping in the Makefile needs to change:


I am considering commitizen for version bumping cz bump --increment PATCH or MAJOR or MINOR are available. It can also be integrated with gh workflows. 
[bump command](https://commitizen-tools.github.io/commitizen/commands/bump/)

pchalasani:
> > If we move away from poetry, the version-bumping in the Makefile needs to change:
> 
> I am considering commitizen for version bumping `cz bump --increment PATCH` or MAJOR or MINOR are available. [commitzen bump command](https://commitizen-tools.github.io/commitizen/commands/bump/)
> 
> see this repo for overview https://github.com/Fraunhofer-AISEC/gallia/blob/1672f25d6b5e20325850ee2ac7b2af4e955c1e27/justfile

Commitizen looks great! I also like `justfile` -- maybe that can be in a later PR

abab-dev:
@pchalasani  Should I just modify the README and quickstart guide and mark this pr ready for review or modify GitHub workflows as well

abab-dev:
@pchalasani  Added git workflows. Please test and provide feedback. I think it only makes sense to change documentation when everything works as expected

pchalasani:
> I think it only makes sense to change documentation when everything works as expected

Agreed. Will do soon, thanks.

.github/workflows/pytest.yml
```diff
@@ -78,49 +78,44 @@ jobs:
...
 
     - name: Install dependencies
       run: |
         sudo apt-get update
         sudo apt-get install -y libmagic1
         sudo apt-get install -y poppler-utils tesseract-ocr
-
-    - name: Load cached Poetry installation
-      uses: actions/cache@v4
-      with:
```
pchalasani, line 90: I guess it's ok to omit this caching since uv build is fast anyway?

.gitignore
```diff
@@ -32,6 +32,7 @@ share/python-wheels/
 .installed.cfg
 *.egg
 MANIFEST
+uv.lock
```
pchalasani, line 35: I was thinking we should include `uv.lock` in tracked files (just like we were git-tracking `poetry.lock`) since it ensures contributors have the same exact pkg versions. What do you think @abab-dev ?

pchalasani:
@abab-dev I made a dupe of this PR to test the gh workflow

.github/workflows/pytest.yml
abab-dev: @pchalasani 
peotry w/o cache : https://github.com/langroid/langroid/actions/runs/12495980351/job/34867250739?pr=653 time:1m46s

peotry with cache: https://github.com/langroid/langroid/actions/runs/12601730026/job/35123410912 time:35s

uv with enable-cache:true but not explicit actions/cache : https://github.com/abab-dev/langroid/actions/runs/12649839216/job/35247092728 time:35s

uv with actions/cache: https://github.com/abab-dev/langroid/actions/runs/12649614195/job/35246411285 time:35s

.gitignore
abab-dev: yes it was mistake will include in next pr

pchalasani:
@abab-dev I made a comment in the dupe of this PR #663  about `uv sync` ignoring version bound for pytest, which is causing test failure since `--cov` is not recognized. Also I don't know if it ignores version bound for other deps, haven't checked.

pchalasani:
@abab-dev whenever you're ready, please make it <kbd> ready for review </kbd> and I'll test out, thanks!

abab-dev:
@pchalasani Update docs now? or in different PR.

pchalasani:
> Update docs now? or in different PR.

We should update docs in same PR, but first I'll test the gh workflow in the other dupe PR

> Also lot of poetry install commands still in codebase

Ok, let's remove or change them to uv once I test the workflows.


> also should keep the `actions/cache` section as comments in `pytest.yml` for reference or delete it.

Ok

pchalasani:
> I'll test the gh workflow in the other dupe PR

@abab-dev confirmed the workflows succeeded in the dupe, so please proceed with the rest of the items, thank you!

abab-dev:
@pchalasani  Lot of work is in async. So please before you go offline give out comments about poetry install. They are contextual in the sense when user is using poetry as package manager and using langroid as package vs when dev is working on langroid to make some changes also some pip commands are also there so should we replace them with uv pip (uv pip compatible interface)?

pchalasani:
> @pchalasani Lot of work is in async. So please before you go offline give out comments about poetry install. They are contextual in the sense when user is using poetry as package manager and using langroid as package vs when dev is working on langroid to make some changes also some pip commands are also there so should we replace them with uv pip (uv pip compatible interface)?

I see what you mean -- for dev instructions, we can restrict them to uv, but for a user, we should retain some flexibility in instructions, and allow pip/poetry/uv, is that what you mean? If so, I agree. I am updating some of the poetry-related files in the dupe PR, will ping when done.

pchalasani:
@abab-dev please pull from `add-uv-dupe` branch, I updated these files to mention `uv` alongside `poetry` (or instead of `poetry`):

```
modified:   README.md
modified:   docs/quick-start/setup.md
modified:   docs/tutorials/postgresql-agent.md
modified:   examples/docqa/streamlit-app/README.md
```

The main thing for you to take care of is the instructions in `CONTRIBUTING.md`. Or let me know if there is any other place we should handle.

CONTRIBUTING.md
```diff
@@ -115,33 +115,31 @@ Emergent autonomous scientific research capabilities of large language models ht
...
-poetry env use python3.11
 
 # use poetry to install dependencies (these go into .venv dir)
-poetry install --with dev
+uv sync --dev 
 ```
 
 Important note about dependencies management:
-> As of version 0.18.3, we are starting to include the `poetry.lock` file as part of 
+> As of version 0.18.3, we are starting to include the `uv.lock` file as part of
```
abab-dev, line 139: update this version once we merge into main and bump version

abab-dev:
Few considerations,
`uv add package` adds package to pyproject.toml modifies the uv.lock and installs stuff
`uv sync ` installs the stuff from pyproject.toml
`uv pip install` installs the package to venv but does nothing to uv.lock of pyproject.toml

docs/tutorials/postgresql-agent.md
```diff
@@ -29,11 +29,13 @@ automatically extracting schemas from the database.
...
 the `.env` file). Note that to use the SQLChatAgent with a PostgreSQL database,
 you need to install the `langroid[postgres]` extra, e.g.:
-    - `pip install langroid[postgres]` or 
-    - `poetry add langroid[postgres]` or 
-    - `poetry install -E postgres`.
 
-    If this gives you an error, try `pip install psycopg2-binary` in your virtualenv.
+    - `pip install "langroid[postgres]"` or 
+    - `poetry add "langroid[postgres]"` or `uv add "langroid[postgres]"`
+    - `poetry install -E postgres` or `uv pip install --extra postgres`
```
abab-dev, line 35: `uv pip install --extra` does not work and it demands -r pyproject.toml file

docs/quick-start/setup.md
```diff
@@ -57,8 +60,9 @@ For many practical scenarios, you may need additional optional dependencies:
         - `sudo apt-get install libpq-dev` on Ubuntu,
         - `brew install postgresql` on Mac, etc.
     - Install langroid with the postgres extra, e.g. `pip install langroid[postgres]`
-      or `poetry add langroid[postgres]` or `poetry install -E postgres`.
-      If this gives you an error, try `pip install psycopg2-binary` in your virtualenv.
+      or `uv add "langroid[postgres]"` or `uv pip install --extra postgres`.
```
abab-dev, line 63: `uv pip install --extra` does not work and it demands -r pyproject.toml file

pchalasani: I'll update both of these in the dupe PR

pchalasani:
Thank you @abab-dev , finally got it done here, after a few adjustments to the makefile 

https://github.com/langroid/langroid/releases/tag/0.33.3

abab-dev:
@pchalasani Very glad, I could help! Really hope to contribute more.

PR #663, opened 2025-01-06
add uv dupe

pchalasani:
- **uv works for installation**
- **exclude tests from build**
- **add uv run and cz bump in Makefile**
- **Github workflows updated with uv**


pchalasani:
@abab-dev  for some reason in the gh `pytest.yml` workflow, it installed the latest `pytest` -

```
 + pytest==8.3.4
```
and ignored the version bound < 8.0.0, and this latest version doesn't recognize the `--cov` and similar args.

I don't know why the version bound was ignored.

pchalasani:
completed in PR #662

PR #672, opened 2025-01-14
add event type to stream callback

johannestang:
I'm building a solution where multiple agents interact via ToolMessages in order to come up with an answer for a user. Any given agent may be the one replying to the user. As soon as an agent is ready to respond to the user, the answer should be streamed to the user and not relayed back to a "controlling" agent (in order to minimize waiting time). I've implemented this using callbacks (based on [callbacks/chainlit.py](https://github.com/langroid/langroid/blob/main/langroid/agent/callbacks/chainlit.py)) which stream to a web app.

My problem is that function calls (i.e. any communication between agents) should never be streamed to the user. In order to achieve this I need to provide the event type to the callback function. This PR implements that functionality.

Note: While I have updated [callbacks/chainlit.py](https://github.com/langroid/langroid/blob/main/langroid/agent/callbacks/chainlit.py) to reflect this change, I have not tested it.

### DIFF
--- a/langroid/agent/callbacks/chainlit.py
+++ b/langroid/agent/callbacks/chainlit.py
@@ -20,6 +20,7 @@ from chainlit.logger import logger
 
 import langroid as lr
 import langroid.language_models as lm
+from langroid.language_models import StreamEventType
 from langroid.utils.configuration import settings
 from langroid.utils.constants import NO_ANSWER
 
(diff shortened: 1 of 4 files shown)


pchalasani:
Thanks @johannestang.
Does it make sense to add a simple script (under  `examples/basic`) that illustrates this? Or is that not easy to do given that you're using it in the context of a webapp?

pchalasani:
@johannestang I added a small fix to `chainlit.py` and made a dupe PR #673

pchalasani:
@johannestang - I merged it in PR #673 , so closing this.

johannestang:
@pchalasani Thanks for merging the PR. Regarding an example: I don't think it would be easy to make a small one. I might try to make a larger one, when I get the time.

PR #677, opened 2025-01-16
sql chat structured fixes

pchalasani:
- **minor comment fixes**
- **chat_agent.py strict recovery: restore orig msg after llm responds**
- **sql_chat_agent.py: adjust recovery and helper agent**

### DIFF
--- a/langroid/agent/chat_agent.py
+++ b/langroid/agent/chat_agent.py
@@ -539,6 +539,20 @@ class ChatAgent(Agent):
                 self.message_history[i].content = message
                 break
 
+    def delete_last_message(self, role: str = Role.USER) -> None:
+        """
+        Delete the last message that has role `role` from the message history.
+        Args:
+            role (str): role of message to delete
+        """
+        if len(self.message_history) == 0:
+            return
+        # find last message in self.message_history with role `role`
+        for i in range(len(self.message_history) - 1, -1, -1):
+            if self.message_history[i].role == role:
+                self.message_history.pop(i)
+                break
+
     def _create_system_and_tools_message(self) -> LLMMessage:
         """
         (Re-)Create the system message for the LLM of the agent,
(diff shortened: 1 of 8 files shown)


langroid/agent/chat_agent.py
```diff
@@ -1294,15 +1318,25 @@ async def llm_response_async(
                 instructions=True,
             )
             recovery_message = self._strict_recovery_instructions(AnyTool)
+            augmented_message = message
```
pchalasani, line 1321: @nilspalumbo I noticed in some cases these `AnyTool` instructions cause the agent to directly use this tool rather than the original tools, so I added this code to restore the original user msg, so that the `AnyTool` instructions don't persist in msg history. Let me know what you think

pchalasani: Also fixed some of the structured recovery tests -- those tests were failing due to 2 reasons - 

- when the LLM re-generates the "intended" tool, it's not what *we* think it should generate, e.g. sometimes it picks a tool based on arg name not on tool name - so I renamed some or the args to "cats" and "cows"
- the `maybe_parse` mechanism (best-guess tool name inferred based on args match) sometimes interferes with testing structured recovery -- I changed some tools to get around this, i.e. to ensure `maybe_parse` doesn't result in a valid tool call so we can test structured recovery in isolation.

nilspalumbo: Hmm, I didn't see that earlier, maybe it's due to a 4o update? I'd have preferred it if it returned the best fit tool based on args and name (perhaps adding some chain of thought before generating the tool may help), but this should work for now.

I think it makes sense to drop those recovery instruction for the most part, however it may make sense to preserve some message stating that the agent should correct its last tool call as the generated tool may not correspond to the actual provided message. We could even modify the history to show the base tool corresponding to the `AnyTool` generated.

PR #680, opened 2025-01-18
Updated multi-agent-debate with Chainlit

adam-shams:
@pchalasani ready for review when you have the time

### DIFF
--- a/examples/multi-agent-debate/chainlit_utils.py
+++ b/examples/multi-agent-debate/chainlit_utils.py
@@ -1,0 +1,363 @@
+import chainlit as cl
+import logging
+from models import SystemMessages
+from typing import Tuple, Optional
+from config import MODEL_MAP
+from utils import extract_topics
+
+DEFAULT_TURN_COUNT = 2
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+
+def parse_boolean_response(response: str) -> bool:
+    """
+    Parse user's response into a boolean value.
+    Args:
+        response (str): User input as a string.
+    Returns:
+        bool: True for yes/true, False for no/false.
+    """
+    response = response.strip().lower()
+    if response in ["yes", "y", "true"]:
+        return True
+    elif response in ["no", "n", "false"]:
+        return False
+    raise ValueError("Invalid boolean response")
+
+
+async def is_same_llm_for_all_agents() -> bool:
+    """
+    Ask the user if they want to use the same LLM for all agents.
+    Returns:
+        bool: True if yes, False if no.
+    """
+    same_llm = await cl.AskUserMessage(
+        content="Do you want to use the same LLM for all agents? (yes/no)", timeout=10
+    ).send()
+    if same_llm:
+        try:
+            result = parse_boolean_response(same_llm["output"])
+            if result:
+                await cl.Message(
+                    content="Great! Same LLM will be used for all agents."
+                ).send()
+            else:
+                await cl.Message(
+                    content="Ok. Different LLMs will be configured for the agents."
+                ).send()
+            return result
+        except ValueError:
+            await cl.Message(
+                content="Invalid response. Please answer yes or no."
+            ).send()
+            return await is_same_llm_for_all_agents()  # Retry on invalid input
+    else:
+        await cl.Message(
+            content="You didn't respond in time. Defaulting to 'no'."
+        ).send()
+        return False
+
+
+async def select_max_debate_turns() -> int:
+    """
+    Ask the user to select the maximum number of turns for debates.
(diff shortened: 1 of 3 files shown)


pchalasani:
Thanks Adam, will look!

examples/multi-agent-debate/chainlit_utils.py
```diff
@@ -0,0 +1,363 @@
...
+from models import SystemMessages
+from typing import Tuple, Optional
+from config import MODEL_MAP
+from utils import extract_topics
+
+DEFAULT_TURN_COUNT = 2
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
```
pchalasani, line 12: when the user's choices are simple like yes or no, see if you can use `AskUserAction` instead of `AskUserMessage`:
https://docs.chainlit.io/api-reference/ask/ask-for-action

I think it lets the user click a button, if I remember right.

pchalasani: Also I would remove the timeouts - these make it hard to keep up

pchalasani (changes requested):
The start of the feedback is not clear in the UI:

image

Also the response from the other side is hidden by default, I think you have to set `cot="full"` in the `.chainlit/config.toml`. I don't recall their options.

In general chainlit have made some odd design choices and changes that mess up some of the multi-agent/tool flow displays, e.g. I've complained here and elsewhere:
https://github.com/Chainlit/chainlit/issues/1052

pchalasani:
> do I need to change anything else in the config.toml?

Not sure, I haven't kept up with their changes - you may want to look in their docs or discord etc, on how to enable the sub-task response to be visible by default

adam-shams:
Screenshot 2025-01-20 at 4 59 54 PM
@pchalasani  The response from the other side is showing after setting cot=“full”  however it repeats the response in the dropdown for (pro -> con ) section.

pchalasani:
> however it repeats the response in the dropdown for (pro -> con ) section.

Ah yes I didn't release the displayed msg was from the other side - so we don't in fact need to expand the sub-task response. Which brings up another issue - for some reason the names of the agents are missing. Chainlit have made some change in how this name can be customized, or maybe they made not customizable ( I remember seeing some issue about that, maybe you can find it, or see the docs)

adam-shams:
> Ah yes I didn't release the displayed msg was from the other side - so we don't in fact need to expand the sub-task response. Which brings up another issue - for some reason the names of the agents are missing. Chainlit have made some change in how this name can be customized, or maybe they made not customizable ( I remember seeing some issue about that, maybe you can find it, or see the docs)

I am checking their docs but no luck so far. I think we will have to change Langroid's `ChainlitTaskCallbacks` class. in `langroid/agent/callbacks/chainlit.py` 

FYI, I ran `examples/chainlit/multi-agent.py` and Chainlit is exhibiting the same behavior. Do you remember the Chainlit version when the `examples/chainlit/multi-agent.py` was implemented? Just want to see their implementation when it worked.

pchalasani:
> Do you remember the Chainlit version when the `examples/chainlit/multi-agent.py` was implemented?

I think it was 0.201.x , not sure, I mentioned it in the chainlit issue I linked to. Or you can see commit history of `chainlit.py`

adam-shams:
image

@pchalasani I suppressed subtask_response by subclassing the ChainlitTaskCallbacks class to get around Chainlit framework bug related to steps. It works around the bug and is not the most elegant solution but makes it usable.

pchalasani:
> get around Chainlit framework bug related to steps.

great work! the display does look a lot cleaner, with clear labeling of Pro Con etc. Too bad we have to do all this to get around chainlit's questionable design choices. There's another framework https://reflex.dev/ I came across but haven't had time to dig in.

pchalasani:
The dialog looks much nicer with the buttons. I noticed the input place holder is too prominent - it should be in a faint color - 

image

pchalasani:
when it times out pick a default, not "no topic"

image

pchalasani:
When I pick autonomous debates, it should not ask me which side I want to be on - 

image

Also for the prior config questions up top, only the answer is visible, the original question disappears. But this may be too hard to fix and we don't want to go into another rabbit hole of chainlit tweaks, so leave it unless it's a pain.

pchalasani:
Same here, only answers visible, can't see the question -- 

image

adam-shams:
> The dialog looks much nicer with the buttons. I noticed the input place holder is too prominent - it should be in a faint color -
> 
> image

chat.input.placeholder is a default setting in chainlit so that will require langroid level changes or a css as it will impact all langroid examples, so holding off on that

PR #681, opened 2025-01-19
Pgvtest

abab-dev:


abab-dev:
Regarding your questions, I wanted to imply that all tests which I wrote are passing. I did not thoroughly test crud. Also set_collection(replace=True) has same effect as creating new collection if it does not exits.

Issue #280, opened 2023-10-31
Add pinecone vectorstore


badal8381:
Add pinecone vectorstore

pchalasani:
@badal8381 did you get a chance to look at this?

pchalasani:
add test in `test_vector_stores.py`

pchalasani:
Should implement an interface very similar to this:

https://github.com/langroid/langroid/blob/main/langroid/vector_store/qdrantdb.py

coretado:
Hello @pchalasani, if no one is working on this interface currently could I work on this issue?

pchalasani:
@coretado thank you, yes this is open!

pchalasani:
done in #685

PR #685, opened 2025-01-20
Add Pinecone interface for vector stores

coretado:
- Links to issue #280


coretado:
Initial commit with tests passing locally, working on doc example next

pchalasani:
Thanks @coretado I assume this is ready for review?

coretado:
Hello @pchalasani, yep fire away

docs/notes/pinecone.md
```diff
@@ -0,0 +1,131 @@
+# How to setup Langroid and Pinecone Serverless
```
pchalasani, line 1: Thanks for this well-made documentation!

tests/main/test_vector_stores.py
```diff
@@ -267,13 +279,19 @@ class MyDocument(Document):
     assert docs_and_scores[0][0].content == "cow"
 
     # test collections: create, list, clear
-    coll_names = [f"test_junk_{i}" for i in range(3)]
+    if isinstance(vecdb, PineconeDB):
+        # pinecone only allows lowercase alphanumeric with "-" characters
+        coll_names = [f"test-junk-{i}" for i in range(3)]
```
pchalasani, line 284: maybe we can just use hyphen for all vecdbs? Or do the others only allow underscore? Not a big deal either way.

langroid/vector_store/pineconedb.py
```diff
@@ -0,0 +1,363 @@
...
+    cloud: bool = True
+    collection_name: str | None = "temp"
+    embedding: EmbeddingModelsConfig = OpenAIEmbeddingsConfig()
+    dimension: int = 1536
+    spec: ServerlessSpec = ServerlessSpec(cloud="aws", region="us-east-1")
+    deletion_protection: Literal["enabled", "disabled"] | None = None
+    metric: str = "cosine"
+    pagination_size: int = 100
+
+class PineconeDB(VectorStore):
```
pchalasani, line 33: since pinecone is an optional dep, we need to do the import(s) in here, similar to how we do for ChromaDB, and
throw a LangroidImportException to get a friendly err msg

langroid/vector_store/pineconedb.py
```diff
@@ -0,0 +1,363 @@
...
+        batch_size = self.config.batch_size
+
+        for i in range(0, len(documents), batch_size):
+            try:
+                if namespace:
+                    index.upsert(vectors=vectors[i:batch_size], namespace=namespace)
+                else:
+                    index.upsert(vectors=vectors[i:batch_size])
+            except PineconeApiException as e:
+                logger.error(f"Unable to add batch of documents between indices {i} and {batch_size}")
```
pchalasani, line 238: These long lines will cause lint errors; please make sure to run the lints and checks -- `make check`

coretado:
Sounds like a plan, thank you!

pchalasani (changes requested):
@coretado please add a pinecone option in `examples/docqa/chat.py` like we did for the other vec-dbs. Thanks!

pchalasani:
@coretado  please note this: https://github.com/langroid/langroid/releases/tag/0.37.4

langroid/vector_store/pineconedb.py
coretado: Added an import guard similar to the `lancedb` and `weaviate` interfaces as the API exception class from the pinecone package is used outside of the vector db instantiation if that's alright

pchalasani:
> Hello @pchalasani, most recent merge has considerations for the changed reference of the embedding function and embedding dimension from `0.37.4`, and has added the pinecone option to `examples/docqa/chat.py`. Also made updates for the optional import and ran the linters.

Much appreciated! Will look in coming days.

pchalasani (commented):
@coretado I did an initial review, looks good! 
Probably nothing to do with your implementation, but I find the pinecone test to be both slow and flake. I pushed some changes (including an important bug-fix in batching in `add_documents`) to this branch https://github.com/langroid/langroid/tree/add-pinecone-vector-store

I got an API key and ran the first test, and it was slow, plus it fails non-deterministically with no doc matches found. I have no idea why.

coretado:
@pchalasani Thank you for catching that! Hmm, I wonder if the eventual consistency of Pinecone on serverless is leading to that behavior. Will merge your update and take a look at those tests.

pchalasani:
I think the tests didn't catch the batching bug -- maybe worth tweaking the number of docs or batch size in the tests to make sure some test would have failed with the buggy version. It's
funny I had the exact same bug several months ago in a different vector db so I knew to look for it.

coretado:
Makes sense, I can look into triggering that behavior in one of the tests (perhaps the `test_vector_stores::test_vector_stores_access` function). Ah so you had the intuition immediately 🚀

pchalasani:
> Will merge your update and take a look at those tests.

Let me know what you find. My test still fails.

coretado:
Hello @pchalasani, it seems like the [battle against eventual consistency](https://community.pinecone.io/t/there-can-be-a-slight-delay-before-new-or-changed-records-are-visible-to-queries/4912) in serverless indexes are shared by other folks. For serverless Pinecone, their docs mention [either checking the log sequence number](https://docs.pinecone.io/guides/data/data-freshness/check-data-freshness) (LSN), or that the number of vectors matches what you would expect.

It feels like it might be hard to determine the exact number of vectors you're expecting to be reflected in an eventually consistent query in a real world use case if a user's Pinecone instance is being updated frequently.

Their docs direct you to the [/vectors/upsert](https://docs.pinecone.io/reference/api/2024-10/data-plane/upsert) API call to get the LSN, but the python SDK does not return this information in the response.

coretado:
One approach to resolve the flaky tests could be to check that the vector IDs used per test instance exist before running the validations on retrieved documents. I've tried separating the Pinecone tests into a dedicated class using a single serverless DB instance to limit the setup time and added polling to search for the expected document IDs.

With tests limited to N=20 documents, hopefully the RU usage isn't astronomical. The class for the Pinecone test takes about 2-3 minutes but does pass consistently. I've added these changes to the branch and would appreciate your thoughts on this approach.

langroid/vector_store/pineconedb.py
```diff
@@ -0,0 +1,428 @@
...
+                    )
+                else:
+                    index.upsert(vectors=vectors[i : i + batch_size])
+            except PineconeApiException as e:
+                logger.error(
+                    f"Unable to add of docs between indices {i} and {batch_size}"
+                )
+                logger.error(e)
+
+    def vectors_resolvable(
```
coretado, line 263: Reading this again, it certainly seems like it could be supplanted in the tests by simply calling `get_documents_by_ids` in a polling pattern and ensuring that the ids from this call match the ids for the documents used in the test. Can probably cull this function in a future branch update if reusing the defined class interface function in the tests is preferable.

tests/main/test_vector_stores.py
```diff
@@ -56,6 +93,7 @@ def vecdb(request) -> VectorStore:
             collection_name="test-" + embed_cfg.model_type,
             storage_path=qd_dir,
             embedding=embed_cfg,
+            batch_size=10,
```
coretado, line 96: Adding a batch_size override to 10 to test that the vector db's used in the `test_vector_stores_access` can batch upsert the expected documents

pchalasani:
> I've tried separating the Pinecone tests into a dedicated class using a single serverless DB instance to limit the setup time and added polling to search for the expected document IDs.

@coretado Thanks, the tests do seem to pass (thought it started off very slow, then paused a long time  5 passed, inexplicably, etc). But I really don't want to make a special case for pinecone with all the horrible code duplication. I thought Pinecone was an "enterprise-ready" vecdb so this is shocking. For the moment can you please revert and mark those tests as skip using 
```python
pytest.param("pinecone_serverless", marks=pytest.mark.skip)
```
like I did for weaviatedb in `main` branch, until we figure out a good way to deal with this?

coretado:
Certainly, removed the class based test for pinecone and added skip params for `pinecone_serverless`

pchalasani:
closing, dupe in #722  for testing

PR #692, opened 2025-01-23
feat: support reasoning LLMs - capture reasoning_content; closes #691

### LINKED ISSUE #691: OpenAI-compatible API support for reasoning LMs (o1, R1, QwQ, etc)
For Reasoning LMs, the chat-completion message or chunks can have `reasoning_content`.
We should:
- capture and print this during streaming
- insert this as `LLMResponse.reasoning` (in addition to `LLMResponse.message`)
- insert this into the cached OpenAIResponse object
- return this as `ChatDocument.reasoning_content` to the agent, to do anything it wants with it

pchalasani:


pchalasani:
Also addresses Discussion #430

Issue #469, opened 2024-05-14
Add pgvector vectorstore


sammyrulez:
pgvector is a very handy way in mid-size projects to store vector data and embeddings alongside other application data

I would like to use it with longroid
ì

pchalasani:
This can be similar to how `qdrant` is integrated

abab-dev:
@pchalasani, whenever you have a chance, could you take a look at the Weaviate-DB PR?  In the meantime, should I  work on this one?

pchalasani:
Thank you @abab-dev, will review. And yes, adding pgvector would be great, thanks for volunteering!

pchalasani:
Hi @abab-dev I had given this to another contributor several weeks ago and thought it wasn't being worked on, and he just got back to me saying he made progress this week and was testing it. Would it be ok for you to collaborate with him on this?

jaixbhatia:
Draft PR: https://github.com/langroid/langroid/pull/675

@abab-dev sorry for the confusion — I was assigned this a while back but got caught up with finals. I got acquainted with the codebase and started working on it on Monday. The draft PR contains my current TODOs. Feel free to reach out to me at "jaixb" on discord

pchalasani:
addressed in #696

PR #696, opened 2025-01-24
adds postgres-db to vector stores

abab-dev:
addresses #469


pchalasani:
@abab-dev BTW please make sure to run `make check` to lint/type erros

pchalasani:
@abab-dev heads-up: https://github.com/langroid/langroid/releases/tag/0.37.4

abab-dev:
@pchalasani  Please take a look.

langroid/vector_store/postgres.py
```diff
@@ -0,0 +1,402 @@
...
+        else:
+            raise ValueError(
+                "Provide either Docker or Cloud config to connect to the database."
+            )
+
+        return create_engine(connection_string, pool_size=10, max_overflow=20)
+
+    def _setup_table(self) -> None:
+        try:
+            from pgvector.sqlalchemy import Vector
```
pchalasani, line 102: please use LangroidImportError like we did with Weaviate and other extras

langroid/vector_store/postgres.py
```diff
@@ -0,0 +1,402 @@
...
+        self.SessionLocal = sessionmaker(
+            autocommit=False, autoflush=False, bind=self.engine
+        )
+        self.metadata = MetaData()
+        self._setup_table()
+
+    def _create_engine(self) -> Engine:
+        """Creates a SQLAlchemy engine based on the configuration."""
+
+        connection_string = None  # Ensure variable is always defined
```
pchalasani, line 60: insert type `:str`

langroid/vector_store/postgres.py
```diff
@@ -0,0 +1,402 @@
...
+
+        if self.config.docker:
+            username = os.getenv("POSTGRES_USER", "postgres")
+            password = os.getenv("POSTGRES_PASSWORD", "postgres")
+            database = os.getenv("POSTGRES_DB", "langroid")
+
+            if not (
+                self.config.host
+                and self.config.port
+                and username
```
pchalasani, line 70: since we're getting defaults in the lines above, these will never be None, so this check is redundant? Should we _not_ get those defaults?

langroid/vector_store/postgres.py
```diff
@@ -0,0 +1,400 @@
...
+
+        self.embeddings_table = Table(
+            self.config.collection_name,
+            self.metadata,
+            Column("id", String, primary_key=True, nullable=False, unique=True),
+            Column("embedding", Vector(self.embedding_dim)),
+            Column("document", String),
+            Column("cmetadata", JSONB),
+            extend_existing=True,
+        )
```
pchalasani, line 117: Please add a comment regarding `cmetadata` -- is this a required field when using pgvector?

langroid/vector_store/postgres.py
```diff
@@ -0,0 +1,400 @@
...
+
+        self.metadata.create_all(self.engine)
+        self.metadata.reflect(bind=self.engine, only=[self.config.collection_name])
+
+        # Create HNSW index if it doesn't exist
+        index_name = f"hnsw_index_{self.config.collection_name}_embedding"
+        with self.engine.connect() as connection:
```
pchalasani, line 124: Add comment, clarifying - are we initializing the index prior to populating with data?

langroid/vector_store/postgres.py
```diff
@@ -0,0 +1,400 @@
...
+        return bool(result)
+
+    @staticmethod
+    def _create_vector_extension(conn: Engine) -> None:
+        with conn.connect() as connection:
+            with connection.begin():
+                statement = text(
+                    "SELECT pg_advisory_xact_lock(1573678846307946496);"
+                    "CREATE EXTENSION IF NOT EXISTS vector;"
+                )
```
pchalasani, line 155: add comment about this strange number

langroid/vector_store/postgres.py
pchalasani: Also comment - is there an index for each table?

langroid/vector_store/postgres.py
```diff
@@ -0,0 +1,400 @@
...
+            query = session.query(self.embeddings_table)
+
+            # Apply 'where' clause if provided
+            if where:
+                try:
+                    where_json = json.loads(where)
+                    query = query.filter(
+                        self.embeddings_table.c.cmetadata.contains(where_json)
+                    )
+                except json.JSONDecodeError:
```
pchalasani, line 270: is there a test for the `where` feature? If not, please add a separate test for this in `test_vector_stores.py`

langroid/vector_store/postgres.py
```diff
@@ -0,0 +1,400 @@
...
+
+                if new_records:
+                    stmt = insert(self.embeddings_table).values(new_records)
+                    session.execute(stmt)
+                session.commit()
+
+    @staticmethod
+    def _id_to_uuid(id: str, obj: object) -> str:
+        try:
+            doc_id = str(uuid.UUID(id))
```
pchalasani, line 331: maybe could be in the base class, since it may be useful for other vecdbs? 
(Or if this is specific to pgvector, please add comment saying so)

tests/main/test_vector_stores.py
```diff
@@ -276,14 +288,22 @@ class MyDocument(Document):
...
     assert docs_and_scores[0][0].content == "cow"
-
-    coll_names = [f"Test_junk_{i}" for i in range(3)]
-    for coll in coll_names:
-        vecdb.create_collection(collection_name=coll)
-    n_colls = len(
-        [c for c in vecdb.list_collections(empty=True) if c.startswith("Test_junk")]
-    )
-    n_dels = vecdb.clear_all_collections(really=True, prefix="Test_junk")
+    if isinstance(vecdb, WeaviateDB):
```
pchalasani, line 291: lot of repeated code - maybe consolidate, and add short comment indicating what we are doing in this if/then clause.

pchalasani (changes requested):
Thanks @abab-dev ! I added some suggestions. Please add short docs in `notes/pgvector.md` with instructions so I can follow them and test it out.

tests/main/test_vector_stores.py
abab-dev: @pchalasani  We are creating the database with first name capitalized for explicit clarity to the user. 
Tests fail if we do not do this for weaviate but
 if we keep database name like Test_junk for other dbs it gives error
 while creating and deleting for eg. in case of postgres 
or
 I can consolidate the create_collection step since I am validating the name in `weaviatedb.py` 
but I think it is better for clarity overall. I will add the comment surely

langroid/vector_store/postgres.py
abab-dev: This is surely not exclusive to postgres we are doing this in` qdrant`,`weaviate `with its util functions but I had general question . Do you think its good idea to have uniformity and enforce the `metadata id` to be a valid UUID?

langroid/vector_store/postgres.py
abab-dev: Got it! This was oversight. Will definitely remove

langroid/vector_store/postgres.py
pchalasani: >  Do you think its good idea to have uniformity and enforce the `metadata id` to be a valid UUID?
That may depend on the vector-db, so we may address that later (if I understood what you meant).

tests/main/test_vector_stores.py
pchalasani: Yes just add comment please

pchalasani:
Since this seems to require docker, won't it complicate the testing? 

You may need to do something similar to what we do for `test_neo4j_chat_agent.py` where you spin up a docker using docker-compose etc (if I'm not mistaken)

Also we need to add the `postgres` extra to `pytest.yml` so that the tests run

docs/notes/pgvector.md
```diff
@@ -0,0 +1,134 @@
...
+## **Setup PostgreSQL with pgvector using Docker**
+
+To quickly get a PostgreSQL instance with pgvector running, the easiest method is to use Docker. Follow the steps below:
+
+### **1. Create a `.env` file for PostgreSQL credentials**
+
+Add the following environment variables to a `.env` file for configuring your PostgreSQL connection:
+
+```dotenv
+POSTGRES_USER=your_postgres_user
```
pchalasani, line 13: not clear how or where we get these values from

pchalasani: maybe the docker stuff should come first

docs/notes/pgvector.md
```diff
@@ -0,0 +1,134 @@
...
+docker run --name pgvector -e POSTGRES_USER=your_postgres_user -e POSTGRES_PASSWORD=your_postgres_password -e POSTGRES_DB=your_database_name -p 5432:5432 ankane/pgvector
+```
+
+This will pull the `ankane/pgvector` image and run it as a PostgreSQL container on your local machine. The database will be accessible at `localhost:5432` with the credentials provided in your `.env` file.
+
+---
+
+## **Installation**
+
+If you are using `uv` or `pip` for package management, install Langroid with PostgreSQL extra:
```
pchalasani, line 32: `postgres` extra

docs/notes/pgvector.md
```diff
@@ -0,0 +1,134 @@
...
+```bash
+uv add langroid[postgres]  # or
+pip install langroid[postgres]
+```
+
+---
+
+## **Code Example**
+
+Here's an example of how to use Langroid with PostgreSQL:
```
pchalasani, line 43: I'm not too familiar with this -- should we be saying pgvector, i.e. we're talking about the vector-db, not the SQL aspect right?

docs/notes/pgvector.md
pchalasani: never mind, I see we can set these to whatever we want

tests/main/test_vector_stores.py
```diff
@@ -122,6 +123,17 @@ def vecdb(request) -> VectorStore:
         rmdir(cd_dir)
         return
 
+    if request.param == "postgres":
```
pchalasani, line 126: the docker server must be running, for this to work, I assume. As mentioned in the other comment we have to have some preamble code to spin up the docker probably, like we do in `test_neo4j_chat_agent.py`

pchalasani:
@abab-dev I followed the postgres set up instructions in your doc, and ran the test but it failed - 

First I ran this docker - 

```bash
docker run --name pgvector -e POSTGRES_USER=pog_user -e POSTGRES_PASSWORD=pog_pw -e POSTGRES_DB=pog_db -p 5432:5432 ankane/pgvector
```

and then when I ran the `test_vector_stores.py` it failed:

```
       dsn = _ext.make_dsn(dsn, **kwargs)
>       conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
E       sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) connection to server at "127.0.0.1", port 5432 failed: FATAL:  role "postgres-user" does not exist
E
E       (Background on this error at: https://sqlalche.me/e/20/e3q8)

.venv/lib/python3.11/site-packages/psycopg2/__init__.py:122: OperationalError
```

abab-dev:
> ```shell
> docker run --name pgvector -e POSTGRES_USER=pog_user -e POSTGRES_PASSWORD=pog_pw -e POSTGRES_DB=pog_db -p 5432:5432 ankane/pgvector
> ```
> 
> and then when I ran the `test_vector_stores.py` it failed:
> 
> ```
>        dsn = _ext.make_dsn(dsn, **kwargs)
> >       conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
> E       sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) connection to server at "127.0.0.1", port 5432 failed: FATAL:  role "postgres-user" does not exist
> E
> E       (Background on this error at: https://sqlalche.me/e/20/e3q8)
> 
> .venv/lib/python3.11/site-packages/psycopg2/__init__.py:122: OperationalError
> ```

`-e POSTGRES_USER=pog_user -e POSTGRES_PASSWORD=pog_pw -e POSTGRES_DB=pog_db ` environment variables you are setting here should match the variables in your .env variables.
So your environment variables should be 
```
POSTGRES_USER=pog_user
POSTGRES_PASSWORD=pog_pw
POSTGRES_DB=pog_db
```

docs/notes/pgvector.md
abab-dev: No! these should match with the environment variables you have provided while starting the docker container. 
`-e POSTGRES_USER=pog_user -e POSTGRES_PASSWORD=pog_pw -e POSTGRES_DB=pog_db`
this should be in your .env as well

langroid/vector_store/postgres.py
abab-dev: So postgres allows having anything as `id` like a Integer. So what is your suggestion? Do we just keep the `ID` or convert it to `UUID` as I am doing here? particularly in case of postgres? But  catch is  in the tests we are inserting stuff with duplicate `IDs` into the db, so the tests fail if we don't enforce the UUID.

docs/notes/pgvector.md
abab-dev: PostgreSQL is the database and it has pgvector extension. But it has many extensions for specialized tasks like PostGIS adds support for geographic objects. So which name makes sense here?

abab-dev: Easiest route here would be to use aiven (or any other cloud service) with their postgres offering. But I dont not know the reliability of their free tier.

tests/main/test_vector_stores.py
abab-dev: Easiest route here would be to use aiven (or any other cloud service) with their postgres offering. But I dont not know the reliability of their free tier.

abab-dev: I think we can use https://cloud.tembo.io/

langroid/vector_store/postgres.py
pchalasani: Your code is fine, let's stay with this.

pchalasani:
@abab-dev let me know when ready, I think you switched the test to using cloud, which is good, but probably the `docs/notes` needs some mention of how to set that up and get the connection string etc.

abab-dev:
@pchalasani  ready for review!

tests/main/test_vector_stores.py
```diff
@@ -122,6 +126,18 @@ def vecdb(request) -> VectorStore:
...
+        pg_cfg = PostgresDBConfig(
+            collection_name="test_" + embed_cfg.model_type,
+            embedding=embed_cfg,
```
pchalasani, line 132: if this is using tembo's connection string, I am puzzled how the tests passed in GH Actions since you'd have to store the `POSTGRES_CONNECTION_STRING` as a secret.

pchalasani: never mind, I see only the lint was run

pchalasani:
done in PR #718

Issue #698, opened 2025-01-25
Gemini Embeddings in EmbeddingModel


abab-dev:
Would it be a good idea to integrate Google AI Studio Gemini Embeddings directly into Embedding Models? Its API is quite similar to the OpenAI API.

pchalasani:
> Would it be a good idea to integrate Google AI Studio Gemini Embeddings directly into Embedding Models? Its API is quite similar to the OpenAI API.

Are there any particularly good/popular/stand-out embdding models from AI Studio, or would this be primarily for convenience for those in that ecosystem?

abab-dev:
Yes, convenience would be one of the factors and ease of use. Practically anyone with google account can get API key.

pchalasani:
Sounds good to me if you want to take this up, thanks!

abab-dev:
Yes sure, thank you.

pchalasani:
done in PR #706

PR #699, opened 2025-01-25
add gemini embeddings model to EmbeddingModel

abab-dev:
addresses #698


langroid/embedding_models/models.py
```diff
@@ -437,6 +447,60 @@ def embedding_dims(self) -> int:
...
 
+class GeminiEmbeddings(EmbeddingModel):
+    def __init__(self, config: GeminiEmbeddingsConfig = GeminiEmbeddingsConfig()):
+        try:
+            import google.generativeai as genai
+        except ImportError:
+            raise ImportError(
+                """
+                To use Gemini embeddings, you must install 
+                the google-generativeai package,
```
pchalasani, line 458: Please make `google-generativeai` an extra dependency in `pyproject.toml`, and here use the `LangroidImportError()` like we did in weaviate etc.

pchalasani (changes requested):
@abab-dev thanks for this PR.

Could you please add:
- short tests in `test_embedding_models.py` 
- modify `pytest.yml` to include the gemini-related extra dependency
- add a short doc in `notes/gemini.md` about this (doesn't have to be very detailed) and link to it in `mkdocs.yml` `nav` section.
- add an option to use these embeddings in the `examples/docqa/chat.py` (like we show `qdrant` and `lancedb` etc.

pchalasani:
done in PR #706

PR #717, opened 2025-02-04
Fix: cache miss on download_nltk_resource

kagia:
- when using [find](https://www.nltk.org/_modules/nltk/data.html) nltk requires the  `<subdirectory>/<model>` format.
- when downloading nltk requires `<model>` format.

This means that previously we were always getting cache misses since we only passed in the model name.

**Note:** may also help with #407


pchalasani:
Thank you @kagia !

PR #728, opened 2025-02-05
Weaviate local duplicate for tests

abab-dev:


abab-dev:
@pchalasani  Ready for merge.

.github/workflows/pytest.yml
```diff
@@ -9,8 +9,11 @@ on:
     branches:
       - main
   pull_request:
-      branches:
-        - main
+
+    types:
+      - ready_for_review
```
abab-dev, line 14: I am adding trigger with `ready_for_review` option back.

pchalasani: > I am adding trigger with `ready_for_review` option back.

sure, if it works :)

abab-dev: Yesterday that indented block was the hidden culprit. Nothing was getting triggered. The time when you first made change you indented it one level deeper than expected. I triggered this PR using ready for review. Will test few times and update again tomorrow.

PR #730, opened 2025-02-06
add markitdown to check the feasibility

abab-dev:
Get a feel for MarkItDown with these steps:  

1. Run:  
   ```bash
   uv sync --all-groups --all-extras
   ```  
2. Execute the test:  
   ```bash
   uv run langroid/tests/extras/test_markitdown.py
   ```


pchalasani:
thanks @abab-dev !
I got this err:

```
tests/extras/test_markitdown.py:56: in <module>
    test_markitdown()
tests/extras/test_markitdown.py:51: in test_markitdown
    doc = parser.get_doc()
langroid/parsing/document_parser.py:388: in get_doc
    [
langroid/parsing/document_parser.py:388: in <listcomp>
    [
langroid/parsing/document_parser.py:689: in iterate_pages
    result = md.convert_stream(self.doc_bytes, file_extension=file_extension)
.venv/lib/python3.11/site-packages/markitdown/_markitdown.py:1142: in convert_stream
    result = self._convert(temp_path, extensions, **kwargs)
.venv/lib/python3.11/site-packages/markitdown/_markitdown.py:1260: in _convert
    raise UnsupportedFormatException(
E   markitdown._markitdown.UnsupportedFormatException: Could not convert '/var/folders/dx/39xz0fk938zftc3djhbm78bc0000gn/T/tmpd36ucynd' to Markdown. The formats ['.md'] are not supported.
```

tests/extras/test_markitdown.py
```diff
@@ -0,0 +1,56 @@
+import os
+
+from langroid.parsing.document_parser import DocumentParser
+from langroid.parsing.parser import MarkitDownParsingConfig, ParsingConfig
+
+url1 = "tests/main/dummy.pdf"
```
pchalasani, line 6: let's put these in `tests/main/data/`

pyproject.toml
```diff
@@ -56,6 +56,7 @@ dependencies = [
     "trafilatura<2.0.0,>=1.5.0",
     "typer<1.0.0,>=0.9.0",
     "wget<4.0,>=3.2",
+    "xlrd>=2.0.1",
```
pchalasani, line 59: if this is only needed for markitdown we should put it in a suitable extra group

langroid/parsing/document_parser.py
```diff
@@ -627,6 +661,57 @@ def get_document_from_page(self, page: pypdf.PageObject) -> Document:
...
+            )
+            result = md.convert_stream(self.doc_bytes, file_extension=file_extension)
+        else:
+            logger.warning(
+                "Markdown conversion from bytes without file extension "
+                "does not work with markitdown."
+            )
+            return
+
+        yield 1, result.text_content
```
pchalasani, line 697: so we're converting the whole thing to md, then passing it as a single "page" -- the page-wise generation is used downstream in doc-chat agent etc to show page nums in the source, so in this case they would all show up as page 1.

I wonder if there's some hack to make page-wise generation like I did in docling. I think you said there's no way possible, not sure.

tests/extras/test_markitdown.py
```diff
@@ -0,0 +1,56 @@
...
+
+    for url in [url3, url4, url5]:
+        ext = os.path.splitext(url)[1].lower()[1:]
+        parser = DocumentParser.create(
+            url,
+            ParsingConfig(
+                n_neighbor_ids=2,
+                markitdown=MarkitDownParsingConfig(),
+            ),
+            doc_type=ext,
```
pchalasani, line 49: Would this hack work when using `DocChatAgent.ingest_doc_paths(["xyz.md", "blah.xls", ...])` ?
If not we should have some way to populate the `doc_type` arg so that `DocChatAgent.ingest_doc_paths` works seamlessly.

langroid/parsing/document_parser.py
abab-dev: This was to test the usefulness of markitdown. I wanted to show you quality of outputs markitdown generates. Also to generate the pages I have to override the every document parsing class of markitdown. Instead wont it be better to just take the inspiration from markitdown and write our own classes.

tests/extras/test_markitdown.py
abab-dev: This is not fully compatible pr. I thought you would not review.

pchalasani: > This is not fully compatible pr. I thought you will not review.

Ok, I see, got it

abab-dev:
Can I just copy classes from markitdown and adapt them to document_parser. Apart from pptx and xlsx what do you think is important? Markitdown supports rss, Outlook messages and wav mp3 (not relevant for us right now).

pchalasani:
> Can I just copy classes from markitdown and adapt them to document_parser (for pagination etc) . 

sounds fine as long as this doesn’t use internal/unstable parts of markITDown that break with future versions. 

> Apart from pptx and xlsx what do you think is important? Markitdown supports rss, Outlook messages and wav mp3 (not relevant for us right now).

Docx files will be good, we use other libs for this currently but good to have a choice.

abab-dev:
> Docx files will be good, we use other libs for this currently but good to have a choice.
I think will be too convoluted for virtually no advantage. As markitdown plainly converts the docx to html and outputs the plaintext.

pchalasani:
> @pchalasani please review this as well. I have added seperate parsers for pptx, xlsx and xls.

Will do, thanks!

langroid/parsing/document_parser.py
```diff
@@ -857,3 +883,79 @@ def get_document_from_page(self, page: Any) -> Document:
             content=self.fix_text(paragraph.text),
             metadata=DocMetaData(source=self.source),
         )
+
+
+class MarkitdownXLSXParser(DocumentParser):
```
pchalasani, line 888: I see we're converting page by page, is that slow?
I guess there's no way from markitdown to convert all at once and retain page/sheet info?

langroid/parsing/document_parser.py
```diff
@@ -857,3 +883,79 @@ def get_document_from_page(self, page: Any) -> Document:
...
+        for i, slide in enumerate(prs.slides, start=1):
+            new_prs = Presentation()
+            new_slide_layout = new_prs.slide_layouts[5]
+            new_slide = new_prs.slides.add_slide(new_slide_layout)
+
+            for shape in slide.shapes:
+                el = shape._element
+                new_slide.shapes._spTree.insert_element_before(el, "p:extLst")
+
+            with tempfile.NamedTemporaryFile(suffix=".pptx", delete=True) as temp_file:
```
pchalasani, line 940: same comment here about speed of page-by-page conversion (especially the temp file write may be slow, not sure.)

langroid/parsing/document_parser.py
pchalasani: Adding a comment here will be helpful, about why we do page by page.

langroid/parsing/document_parser.py
pchalasani: A comment here would be helpful - why we have to do page by page.

abab-dev:
Stopping for now. We have to push first to langroid to have access to secrets.

abab-dev:
Modified the parsers first to get markdown from markitdown and then split it into the pages (sheets and slides).

PR #741, opened 2025-02-10
use new google-genai client in GeminiEmbeddings

abab-dev:
`google-generativeai` api client has been updated to `google-genai`
You can see changes here.
https://ai.google.dev/gemini-api/docs/migrate


pyproject.toml
```diff
@@ -236,8 +236,8 @@ momento = [
 fastembed = [
     "fastembed<0.4.0,>=0.3.1",
 ]
-google-generativeai = [
-    "google-generativeai>=0.8.4",
+google-genai = [
+    "google-genai>=1.0.0",
```
pchalasani, line 240: also keep the original extra `google-generative` for backward compatibility in case someone is using it

PR #746, opened 2025-02-11
Pg docker

abab-dev:
Introduce postgres docker container in CI


tests/main/docker-compose-pgvector.yml
```diff
@@ -0,0 +1,11 @@
+version: "3.9"
+services:
```
pchalasani, line 2: do we need this file? I couldn't see where it's being used.

pchalasani (commented):
had a comment

tests/main/docker-compose-pgvector.yml
abab-dev: Removing it.

Issue #734, opened 2025-02-06
direct support for gemini api? (to leverage it for document parsing and other)


pchalasani:
We currently use the OpenAI sdk for gemini models, but this misses out on some functionalities specific to gemini models, e.g. the document understanding of `gemini-2.0-flash`

API:

https://ai.google.dev/gemini-api/docs/document-processing?lang=python

abab-dev:
@pchalasani  I am happy to work on this one.

abab-dev:
@pchalasani I was thinking implementation like openai_gpy.py but Gemini throttles api access aggressively on free instance. So is it okay if I do it gradually. First document understanding only then progressively add features.

pchalasani:
> I was thinking implementation like openai_gpt.py

@abab-dev another approach might be -- since the APIs are very similar, we could add some params to `openai_gpt.py` and use constraints like in `model_info.py` to restrict/rename params for Gemini. Maybe this requires some refactoring of `OpenAIGPT` and related classes.

A similar approach could be done for Anthropic API, which incidentally would also be of great interest, since currently we have to use litellm for anthropic models.

abab-dev:
@pchalasani Makes sense. I will try adapting Gemini models to openai_gpt.py with minimal code changes.

PR #748, opened 2025-02-11
feat: add gemini document parser

abab-dev:
partially address #734 
@pchalasani  What are your thoughts?


pchalasani:
> partially address #734 @pchalasani What are your thoughts?

Looks good! I'll have to try it out a bit. E.g. try this doc: https://nguyenduyliemgis.wordpress.com/wp-content/uploads/2014/09/understanding-gps-principles-and-applications-2006.pdf

pchalasani:
@abab-dev  I tried the `docqa/chat.py` with the above 700 page document about GPS, and it only split it into 28 parts, and wasn't able to answer any questions about later parts of the book.

image

langroid/parsing/parser.py
```diff
@@ -71,6 +76,7 @@ class ParsingConfig(BaseSettings):
     pptx: MarkitdownPPTXParsingConfig = MarkitdownPPTXParsingConfig()
     xls: MarkitdownXLSParsingConfig = MarkitdownXLSParsingConfig()
     xlsx: MarkitdownXLSXParsingConfig = MarkitdownXLSXParsingConfig()
+    gemini: GeminiParsingConfig = GeminiParsingConfig()
```
pchalasani, line 79: if the default is a non-None `GeminParsingConfig` then it will become always-enabled right? Or it will depending the order in which the parsers are considered. Maybe it's better to have the default be None?

pchalasani:
@abab-dev I added some comments, maybe this isn't quite ready yet.

abab-dev:
> @abab-dev I added some comments, maybe this isn't quite ready yet.

You can take a look at this minimal reproducible example.
https://github.com/abab-dev/gemini-markdown 
The ingested PDF contains approximately 700,000 tokens, but after parsing, Gemini only outputs around 8,000 tokens. This might be a limitation by Google.

pchalasani:
> but after parsing, Gemini only outputs around 8,000 tokens.

Right I forgot about that, of course makes sense this model has a max output tokens of 8192. 

Another idea - if there's a fast way to split the pdf into pages (preferably using some python lib, rather than calling "subprocess"), then we call the gemini model with a bunch of batched/async calls to get markdown - so this leverages gemini's pdf parsing ability, and also the page-splitting would now be accurate and not left to gemini's whims.

abab-dev:
@pchalasani  consider taking look at this whenever you get time.

pchalasani:
> @pchalasani consider taking look at this whenever you get time.

Thanks! will look

langroid/parsing/document_parser.py
```diff
@@ -952,3 +956,331 @@ def get_document_from_page(self, md_content: str) -> Document:
...
+        super().__init__(source, config)
+        if not config.pdf.gemini_config:
+            raise ValueError(
+                "GeminiPdfParser requires a Gemini-based config in pdf parsing config"
+            )
+        self.model_name = config.pdf.gemini_config.model_name
+
+        # Ensure output directory exists and generate output filename
+        self.OUTPUT_DIR.mkdir(parents=True, exist_ok=True)
+        self.output_filename = (
```
pchalasani, line 999: why not use standard tempfile lib which automatically creates unique temp file/dir ?

langroid/parsing/document_parser.py
```diff
@@ -952,3 +956,331 @@ def get_document_from_page(self, md_content: str) -> Document:
...
+        saves them to a markdown file, and yields page numbers.
+        """
+        import os
+
+        try:
+            load_dotenv()
+            gemini_api_key = os.environ.get("GEMINI_API_KEY")
+            if not gemini_api_key:
+                raise ValueError("GEMINI_API_KEY not found in environment variables.")
+        except ImportError:
```
pchalasani, line 1206: this doesn't make sense here -- there would be no import error from the try block

langroid/parsing/document_parser.py
```diff
@@ -952,3 +956,331 @@ def get_document_from_page(self, md_content: str) -> Document:
             content=self.fix_text(md_content),
             metadata=DocMetaData(source=self.source),
         )
+
+
+class GeminiPdfParser(DocumentParser):
+
```
pchalasani, line 962: Please insert comments here, describing the general strategy of how we'd use gemini for this, e.g. we extract pages using fritz, group them into reasonable size chunks, send them to gemini api, etc. 

Also, I am wondering why not just split into pages, no page-grouping, and send each page to gemini in async/batches - that would simplify the logic but maybe will hit rate-limits?

langroid/parsing/document_parser.py
```diff
@@ -952,3 +956,331 @@ def get_document_from_page(self, md_content: str) -> Document:
...
+        if not self.split_on_page:
+            self.GEMINI_SYSTEM_INSTRUCTION += (
+                " 7. **Pages:** Identify page boundaries using `<!----Page-{n}---->` "
+                "markers added in the PDF at `(10, 10)` coordinates."
+            )
+
+        # Rate limiting parameters
+        import asyncio
+
+        self.requests_per_minute = config.pdf.gemini_config.requests_per_minute or 5
```
pchalasani, line 1015: more comments needed. What is semaphore doing?

pchalasani:
@abab-dev  thanks for doing this. I had some comments. Also, when I ran the same example above on the 700p doc, it was processing for a long time. I wonder if some progress bar or other intermediate output might be helpful to show progress, without being overwhelming.

langroid/parsing/document_parser.py
```diff
@@ -952,3 +956,331 @@ def get_document_from_page(self, md_content: str) -> Document:
...
+        self.model_name = config.pdf.gemini_config.model_name
+
+        # Ensure output directory exists and generate output filename
+        self.OUTPUT_DIR.mkdir(parents=True, exist_ok=True)
+        self.output_filename = (
+            config.pdf.gemini_config.output_filename or self._generate_output_filename()
+        )
+
+        self.max_tokens = config.pdf.gemini_config.max_tokens or self.DEFAULT_MAX_TOKENS
+        self.split_on_page = config.pdf.gemini_config.split_on_page or False
```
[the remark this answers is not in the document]
pchalasani: @abab-dev I now see you've explained it in the `gemini-pdf-parser.md` docs, but a comment here would also be helpful.

pchalasani:
> it was processing for a long time. I wonder if some progress bar or other intermediate output might be helpful to show progress, without being overwhelming.

It finished and produced the expected number of chunks, and doc-qa works fine. I guess the question of showing a progress bar is not specific to this parser (though some parsers do show a nice progress bar), so we can probably ignore that unless it's easy to do.

Overall with some additional comments in the code I'll be confident enough about it to merge the PR

tests/extras/test_gemini_doc_parser.py
```diff
@@ -0,0 +1,44 @@
+from pathlib import Path
+
```
pchalasani, line 2: we can put this test in tests/main, just remember to include the extra dep in the `pytest.yml` workflow,
and I'll add the secret to the CI secrets.

tests/extras/test_gemini_doc_parser.py
```diff
@@ -0,0 +1,44 @@
...
+    doc = gemini_parser.get_doc()
+
+    # Assertions
+    assert isinstance(doc.content, str)
+    assert len(doc.content) > 0  # assuming the PDF is not empty
+    assert doc.metadata.source == str(path)
+
+    docs = gemini_parser.get_doc_chunks()
+    assert len(docs) > 0
+    assert all(d.metadata.is_chunk for d in docs)
```
pchalasani, line 40: please add an assert to confirm that the first few words of the doc are captured at start of first doc-chunk
and last-few words are at and of last chunk (I checked in debugger, and they are indeed captured)

langroid/parsing/document_parser.py
```diff
@@ -952,3 +956,331 @@ def get_document_from_page(self, md_content: str) -> Document:
...
+            if self.split_on_page:
+                pages = full_markdown.split("<!----Chunk-")
+            else:
+                pages = full_markdown.split("<!----Page-")
+
+            if pages and pages[0] == "":
+                pages = pages[1:]
+
+            for i, page in enumerate(pages):
+                page_number = i + 1
```
pchalasani, line 1262: downstream methods (e.g. `get_doc_chunks`) expect a 0-based page index, so no need to add 1 here. You can see in the `test_gemini_doc_parser` that the first doc-chunk's metadata shows page number as 2 instead of 1, due to this.

tests/extras/test_gemini_doc_parser.py
```diff
@@ -0,0 +1,44 @@
+from pathlib import Path
+
+import pytest
+
+from langroid.parsing.document_parser import DocumentParser
+from langroid.parsing.parser import GeminiConfig, ParsingConfig, PdfParsingConfig
+
+
+@pytest.mark.parametrize("pdf_file", ["imagenet.pdf"])
```
pchalasani, line 9: maybe also test other settings of `GeminiConfig` such as `split_on_page = False`, and others if meaningful.

pchalasani:
@abab-dev I pushed some updates to `gemini-docparser` branch, please merge from there.

langroid/parsing/document_parser.py
abab-dev: I'm ensuring that the output directory is deterministic so the user can easily access and debug the results. While using `tempfile` for storage, I still make sure the user can locate the output easily (so added the `Path(source.stem)`.

langroid/parsing/parser.py
```diff
@@ -29,7 +48,18 @@ class PdfParsingConfig(BaseSettings):
         "unstructured",
         "pdf2image",
         "markitdown",
+        "gemini",
     ] = "pymupdf4llm"
```
pchalasani, line 52: Would this work for non-pdf extraction as well, e.g. doc(x) and others? If so, should we consider naming placing in a more general class?

abab-dev: logic will not be same prompt will be different and splitting logic as well. Because we are splitting using fitz per page.

pchalasani: > logic will not be same prompt will be different and splitting logic as well. Because we are splitting using fitz per page.

oh got it, yes, let's leave as is

pchalasani:
merged in #763

PR #754, opened 2025-02-20
fix: AzureConfig: remove deployment_name, model_version, closes #751

### LINKED ISSUE #751: Azure OpenAI client should not require deployment name
**Describe the bug**
Azure OpenAI client should not require deployment name

**To Reproduce**
Attempt to create  `AzureGPT` client without providing `deployment_name` in .env or explicitly

**Expected behavior**
There are two problems here:
1. The `openai.[Async]AzureOpenAI` client does not require `deployment_name` in order to be configured
2. Even if it did, the `AzureGPT` client should not check for `deployment_name` when providing a custom client provider, since this is the responsibility of the provider

**Screenshots**
Image

pchalasani:


pchalasani:
@nilspalumbo can you take a quick look at this PR, since you have some different Azure settings. I removed the requirement for `azure_deployment_config` (and corresponding env var).

pchalasani:
I looked at the Azure OpenAI instructions in llama-index and they also don't use the deployment when they call the client, so we should keep that out of the picture I think. https://docs.llamaindex.ai/en/stable/examples/llm/azure_openai/

pchalasani:
Just pushed another update to main to remove the check for model_name

nilspalumbo:
It seems that that's no longer being used to identify the OpenAI model corresponding to the deployment? In that case, do we want to tell users that we require deployments to be named the same as the string in `OpenAIChatModel`?

pchalasani:
For my own understanding - here are the scenarios: 

Let: 
DN = Deployment name in user's Azure setup; 
CM = `chat_model` in `AzureConfig` (same as `model_name` if they specify it)

CM is used: 
- to determine model parameter constraints
- to pass to the eventual Azure endpoint completion call 

DN is used:
- to pass to the AzureOpenAI client constructor (will fail if it doesn't match user's azure setup)

Scenarios:
1. DN specified in AzureConfig and not same as CM => will fail on completion call
2. DN specified in AzureConfig, same as CM => completion call accepted but:
   2a. DN/CM is not a standard OpenAI model name => completion call may fail due to bad params

Are you talking about 2a? If so, yes I think we should require DN = CM = standard openAI name if they want proper param constraints applied. Or maybe there are other scenarios that I haven't thought of.

nilspalumbo:
Yes, that's what I meant, but 1 isn't necessarily a problem. If DN and CM are set, we can handle nonstandard deployment names and select proper params from CM.

If DN is not specified, the call made by `AzureOpenAI` will use `chat_model` to select the endpoint of the deployment (so there's no need to specify if your deployment names match the model names). If, for example, your have a deployment named "gpt-4o", leaving `chat_model` unset will work, but, if we don't have DN = CM = standard OpenAI name, both DN and either `chat_model` or`model_name` must be specified.

PR #755, opened 2025-02-24
Allow specifying api_key for non-OpenAI models via configuration object

alexagr:
Update OpenAIGPT class's `__init__()` method to allow specifying non-OpenAI model's `api_key` via configuration object.

### DIFF
--- a/langroid/language_models/openai_gpt.py
+++ b/langroid/language_models/openai_gpt.py
@@ -856,6 +856,15 @@ class OpenAIGPT(LanguageModel):
                 tool_deltas += event_tool_deltas
         else:
             event_text = choices[0]["text"]
+
+        finish_reason = choices[0].get("finish_reason", "")
+        if not event_text and finish_reason == "content_filter":
+            filter_names = []
+            for n, r in choices[0].get("content_filter_results", {}).items():
+                if r.get("filtered"):
+                    filter_names.append(n)
+            event_text = 'Content filtered by [' + ', '.join(filter_names) + ']'
+
         if event_text:
             completion += event_text
             sys.stdout.write(Colors().GREEN + event_text)
@@ -891,7 +900,7 @@ class OpenAIGPT(LanguageModel):
                     self.config.streamer(tool_fn_args)
 
         # show this delta in the stream
-        if choices[0].get("finish_reason", "") in [
+        if finish_reason in [
             "stop",
             "function_call",
             "tool_calls",
--- a/langroid/language_models/openai_gpt.py
+++ b/langroid/language_models/openai_gpt.py
@@ -573,12 +573,10 @@ class OpenAIGPT(LanguageModel):
         # Pydantic's BaseSettings will automatically pick it up from the
         # .env file
         # The config.api_key is ignored when not using an OpenAI model
+        self.api_key = config.api_key
         if self.is_openai_completion_model() or self.is_openai_chat_model():
-            self.api_key = config.api_key
             if self.api_key == DUMMY_API_KEY:
                 self.api_key = os.getenv("OPENAI_API_KEY", DUMMY_API_KEY)
-        else:
-            self.api_key = DUMMY_API_KEY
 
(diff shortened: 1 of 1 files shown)


langroid/language_models/openai_gpt.py
```diff
@@ -573,12 +573,10 @@ def __init__(self, config: OpenAIGPTConfig = OpenAIGPTConfig()):
         # Pydantic's BaseSettings will automatically pick it up from the
         # .env file
         # The config.api_key is ignored when not using an OpenAI model
+        self.api_key = config.api_key
```
pchalasani, line 576: Not sure this would work, since we're setting self.api_key in several of the if/else clauses in the code above. So if we unconditionally set this here, it would break those scenarios. And this is not caught in tests since we don't test non-openai llms in the test suite.

pchalasani:
> Update OpenAIGPT class's `__init__()` method to allow specifying non-OpenAI model's `api_key` via configuration object.

Thanks @alexagr  - it would help to know what scenario you're looking to support here. Did you mean, supporting an API key for say gemini or other provider models?

langroid/language_models/openai_gpt.py
alexagr: i've essentially replaced the following code block:
```
        if self.is_openai_completion_model() or self.is_openai_chat_model():
            self.api_key = config.api_key
            if self.api_key == DUMMY_API_KEY:
                self.api_key = os.getenv("OPENAI_API_KEY", DUMMY_API_KEY)
        else:
            self.api_key = DUMMY_API_KEY
```
with this one:
```
        self.api_key = config.api_key
        if self.is_openai_completion_model() or self.is_openai_chat_model():
            if self.api_key == DUMMY_API_KEY:
                self.api_key = os.getenv("OPENAI_API_KEY", DUMMY_API_KEY)
```

The original code block was overwriting the `self.api_key` anyway - hence any assignments done before it were meaningless.

I just used the fact that `config.api_key` defaults to DUMMY_API_KEY - and thus tried to simplify the assignment logic.

alexagr:
> Thanks @alexagr - it would help to know what scenario you're looking to support here. Did you mean, supporting an API key for say gemini or other provider models?

Exactly! ))) I want to be able to pass Gemini api_key via the config object - as I'm spinning up multiple agent instances, and each one has it's own LLM configuration.

langroid/language_models/openai_gpt.py
pchalasani: > hence any assignments done before it were meaningless.

ah right I didn't notice that (I may have to revisit that code to resolve this)

pchalasani:
> I want to be able to pass Gemini `api_key` via the `config` object

wouldn't this be a bad practice? I typically have a bunch of API keys defined in my .env or env vars,
e.g.
```
GEMINI_API_KEY=...
ANTHROPIC_API_KEY=...
```
Having them in code wouldn't be advisable -- and so it's probably good practice for the framework to discourage that.

pchalasani:
> wouldn't this be a bad practice? I typically have a bunch of API keys defined in my .env or env vars, e.g.

Actually, ignore that. I'll look into this, it's a simplification of the code anyway.

alexagr:
> wouldn't this be a bad practice? 

I'm developing multi-tenant platform where each customer has its own agents and corresponding configuration. The data is stored in external databases / keyvaults and agents are created when new chats are started / voice calls are received. Agent's properties (prompts / LLM configuration / tools / documents / ...) are configurable by customers via Web portal and therefore can change at any time.

In such environment I think it makes perfect sense to use configuration object when instantiating the agent and be able to pass all needed data in it...

langroid/language_models/openai_gpt.py
pchalasani: @nilspalumbo can you look at this redunant setting of self.api_key for vllm etc , which is over-written later on? I guess those aren't really needed?

pchalasani:
done in #757

pchalasani:
thanks @alexagr , merged in #757

Issue #753, opened 2025-02-20
Strict recovery should not trigger when LLM response hasn't yet been invoked.


pchalasani:
Scenario - when agent received text that contains potentially json like strings (e.g latex format), the tool_error flag is getting set to True while trying to recognize tools in the incoming message, and this leads to strict recovery being triggered, since it appears as if the LLM "tried" to generate a tool.  However this should not happen since the LLM has not yet responded.

We should add a suitable regression test.

Issue #750, opened 2025-02-17
strict_recovery breaks `Task.run_async()`


alexagr:
When strict recovery is enabled in Agent's config the following code causes endless loop:

```
from datetime import datetime
import json
import asyncio

from langroid.agent.chat_agent import ChatAgent, ChatAgentConfig
from langroid.agent.task import Task
from langroid.language_models import AzureConfig
from langroid.agent.tool_message import ToolMessage
from langroid.agent.chat_document import ChatDocument
from langroid.mytypes import Entity


llm = AzureConfig(
    deployment_name='gpt-4o',
    model_name='gpt-4o',
    model_version='2024-08-06',
    api_base='https://alex-open-ai-1.openai.azure.com',
    api_version='2024-08-01-preview',
    api_key=API_KEY,
    temperature=0.2,
    max_output_tokens=100
)


class GetTimeTool(ToolMessage):
    purpose: str = "Get current time"
    request: str = "get_time"

    def response(self, agent: ChatAgent) -> ChatDocument:
        now = datetime.now()
        return agent.create_agent_response(
            content=json.dumps(
                {
                    'time': now.strftime('%H:%M:%S'),
                    'date': now.strftime('%Y-%m-%d'),
                    'day_of_week': now.strftime('%A'),
                    'week_number': now.strftime('%U'),
                    'tzname': now.tzname()
                }
            ),
            recipient=Entity.LLM
        )

    @classmethod
    def instructions(cls) -> str:
        return ""


config = ChatAgentConfig(
    name='test-agent',
    llm=llm,
    system_message='''
You are a helpful assistant.
Start by calling the get_time tool. Then greet user according to the time of the day.
'''
)

agent = ChatAgent(config)
agent.enable_message(GetTimeTool)
task = Task(agent)
asyncio.run(task.run_async())
```

The loop looks like that:
```
Stats: N_MSG=2, TOKENS: in=30, out=18, max=100, ctx=128000, COST: now=$0.0003, cumul=$0.0003, tot=$0.0003 (gpt-4o)
Agent: {"time": "17:30:00", "date": "2025-02-17", "day_of_week": "Monday", "week_number": "07", "tzname": null}
{"request":"tool_or_function","tool":{"request":"get_time"}}
Stats: N_MSG=4, TOKENS: in=360, out=15, max=100, ctx=128000, COST: now=$0.0011, cumul=$0.0013, tot=$0.0013 (gpt-4o)
[...]

PR #758, opened 2025-02-26
Strict recovery fixes

nilspalumbo:
Fixes the loop seen in #750 by extending the fix in #678 to `Task.response_async()`. Additionally, when `Agent.get_tool_messages()` is called with a `ChatDocument`, the `tool_error` flag is set only when the sender is the LLM. This should correct #753 in most cases, except when `get_tool_messages` is called with a string-typed argument. 

@pchalasani should we add a condition for `ChatAgent`s that the LLM has sent some message in the message history? Or perhaps we could require that the most recent message was sent by the LLM.


pchalasani:
> @pchalasani should we add a condition for `ChatAgent`s that the LLM has sent some message in the message history? Or perhaps we could require that the most recent message was sent by the LLM.

You mean to handle the case where `get_tool_message` is called on a str-type arg?
Yes I think it makes sense to check that the most recent msg is from LLM

Issue #739, opened 2025-02-08
Add Firecrawl to scrape URL content -> LLM-friendly form


pchalasani:
https://www.firecrawl.dev/

It's an API service. We can look into using it in [`url_loader.py`](https://github.com/langroid/langroid/blob/main/langroid/parsing/url_loader.py) in addition to `trafilatura` -- probably add a suitable config option somewhere.

abab-dev:
I am starting to work on this one.

pchalasani:
Great, thanks !

Issue #707, opened 2025-01-31
Add Marker for pdf parsing


pchalasani:
This is a well-regarded pdf -> markdown parser.
https://github.com/VikParuchuri/marker

Add a `MarkerPDFParser` to `document_parser.py` [https://github.com/langroid/langroid/blob/main/langroid/parsing/document_parser.py]

We should pay particular attention to whether we can _efficiently_ get per-page markdown conversion, like we do with `pymupdf4llm`. We are able to get this via `docling` as well but this is done via a workaround that is extremely slow.

abab-dev:
@pchalasani  I am happy to work on this.

pchalasani:
> [@pchalasani](https://github.com/pchalasani) I am happy to work on this.

Awesome, thanks!

pchalasani:
Or #711 which may be even better - up to you.

abab-dev:
I’m open to working on both! If #711 is a higher priority, I can tackle it first.

abab-dev:
Okay, makes sense. Will open PR soon.

pchalasani:
@abab-dev added you to langroid, so you can directly push to branches on this repo and run gh actions etc.

abab-dev:
Great. Thank you!

mahimairaja:
@pchalasani is this issue still open? Would like to work on this issue

pchalasani:
> [@pchalasani](https://github.com/pchalasani) is this issue still open? Would like to work on this issue

Yes please go ahead, thanks !

pchalasani:
@mahimairaja let me know if you're unblocked on this. If you are stuck I can have @abab-dev take this.

mahimairaja:
@pchalasani I am still working on this

pchalasani:
hi @mahimairaja appreciate your offer for help here, but I wanted to check if this is stalled so someone else can tackle it.
Trying to avoid features stuck in limbo. Thanks

abab-dev:
@pchalasani  I am working on this one.

pchalasani:
> [@pchalasani](https://github.com/pchalasani) I am working on this one.

Thank you @abab-dev

abab-dev:
@pchalasani We cannot install docling and marker simultaneously. It leads to conflicting dependencies. Docling (transformers <=4.43) marker (transformers >=4.45)

pchalasani:
@abab-dev in the docling github `pyproject.toml` it shows this -- 

```toml
transformers = [
[...]

PR #762, opened 2025-02-27
Cut down import time, address #407

abab-dev:
@pchalasani please take a look
- Moved imports to the functions where they are used instead of keeping them at the top level.
- Added a script (`visual_log.sh`) to profile and visualize import times.
- Ensure you have `tuna` installed: `pip install tuna` or `uv add tuna`.
- Make the script executable: `chmod +x visual_log.sh`.
- Run the script: `./visual_log.sh`.
- Open your browser at `http://localhost:8000/` to view the results.
- You can edit `visual_log.sh` to profile your python script by default we profile simple import like import langroid.


abab-dev:
> Thanks @abab-dev; is this meant to address #407?

Yes it does.

pchalasani:
> > Thanks @abab-dev; is this meant to address #407?
> 
> Yes it does.

I just informally tested `import langroid as lr` in ipython on main vs this branch and it seems to cut it down by half (12s vs 6s) - which is great! Will profile it with your suggestion later.

abab-dev:
> > > Thanks @abab-dev; is this meant to address #407?
> > 
> > 
> > Yes it does.
> 
> I just informally tested `import langroid as lr` in ipython on main vs this branch and it seems to cut it down by half (12s vs 6s) - which is great! Will profile it with your suggestion later.

Will add few more optimizations once you provide some feedback on this approach. Your general thoughts/criticism.

pchalasani:
> Will add few more optimizations once you provide some feedback on this approach. Your general thoughts/criticism.

Tested with same visual_log.sh in main, and the improvement is 7s vs 3.8s, which is great.

The approach seems fine, we just have to watch out that the various extras and imports behave as expected. Most of it will be caught by the `make check` and tests in CI so I think it's fine.

abab-dev:
@pchalasani  you can test script with visual log like this now ./visual_log "path_to_your_script.py"

pchalasani:
> @pchalasani you can test script with visual log like this now `./visual_log "path_to_your_script.py" `

This is awesome, worth adding to main as a separate PR I think

pchalasani:
> > @pchalasani you can test script with visual log like this now `./visual_log "path_to_your_script.py" `
> 
> This is awesome, worth adding to main as a separate PR I think

No need though since we'll merge this anyway

pchalasani:
@abab-dev once you’re done with your optimizations, please ping me so I can review and merge. Thank you for this !

pchalasani:
> can you think about lazy/late importing `import openai` in `langroid/language_models/utils.py`

I'm fine with it if you can figure out how -- The various error classes are being set as default values in the function signature, so it may be tricky.

abab-dev:
> > can you think about lazy/late importing `import openai` in `langroid/language_models/utils.py`
> 
> I'm fine with it if you can figure out how -- The various error classes are being set as default values in the function signature, so it may be tricky.

What do you think of this. 
```python 
import asyncio
import logging
import random
from typing import Any, Callable, Dict, List

logger = logging.getLogger(__name__)
logger.setLevel(logging.WARNING)


def retry_with_exponential_backoff(
    func: Callable[..., Any],
    initial_delay: float = 1,
    exponential_base: float = 1.3,
    jitter: bool = True,
    max_retries: int = 5,
) -> Callable[..., Any]:
    """Retry a function with exponential backoff."""

    def wrapper(*args: List[Any], **kwargs: Dict[Any, Any]) -> Any:
        import time
        import requests
        import openai
        import aiohttp

        errors = (
            requests.exceptions.RequestException,
            openai.APITimeoutError,
            openai.RateLimitError,
            openai.AuthenticationError,
            openai.APIError,
            aiohttp.ServerTimeoutError,
            asyncio.TimeoutError,
        )

        num_retries = 0
        delay = initial_delay

        while True:
            try:
                return func(*args, **kwargs)

            except (openai.BadRequestError, openai.AuthenticationError, openai.UnprocessableEntityError) as e:
                logger.error(f"OpenAI API request failed with error: {e}.")
                raise e

            except errors as e:
                num_retries += 1
                if num_retries > max_retries:
                    raise Exception(f"Max retries ({max_retries}) exceeded. Last error: {e}.")

                delay *= exponential_base * (1 + jitter * random.random())
                logger.warning(f"OpenAI API request failed: {e}. Retrying in {delay:.2f} seconds...")
                time.sleep(delay)

    return wrapper


def async_retry_with_exponential_backoff(
    func: Callable[..., Any],
    initial_delay: float = 1,
    exponential_base: float = 1.3,
    jitter: bool = True,
    max_retries: int = 5,
) -> Callable[..., Any]:
    """Retry a function with exponential backoff (async version)."""

    async def wrapper(*args: List[Any], **kwargs: Dict[Any, Any]) -> Any:
        import openai
        import aiohttp

        errors = (
            openai.APITimeoutError,
            openai.RateLimitError,
            openai.AuthenticationError,
            openai.APIError,
            aiohttp.ServerTimeoutError,
            asyncio.TimeoutError,
        )

        num_retries = 0
        delay = initial_delay

        while True:
            try:
                return await func(*args, **kwargs)

            except (openai.BadRequestError, openai.AuthenticationError) as e:
                logger.error(f"OpenAI API request failed with error: {e}.")
                raise e

            except errors as e:
                num_retries += 1
                if num_retries > max_retries:
                    raise Exception(f"Max retries ({max_retries}) exceeded. Last error: {e}.")

                delay *= exponential_base * (1 + jitter * random.random())
                logger.warning(f"OpenAI API request failed: {e}. Retrying in {delay:.2f} seconds...")
                await asyncio.sleep(delay)

    return wrapper
```
I GPT'd it. Seems reasonable.

pchalasani:
does that mean every call to the LLM will be slightly slower due to this import?

abab-dev:
> does that mean every call to the LLM will be slightly slower due to this import?

I am not sure. I will check.

abab-dev:
You are right. This would make the calls slower. So I think ,if we have to resort to a few antipatterns (non pythonic ways) for relatively niche modules (vecdbs, niche libraries like nltk,Py GitHub), that's okay but simpler is better.

pchalasani:
Agreed

pchalasani:
I thought Python would cache the imports so if there are repeated llm calls the average impact per call will be small? Or maybe you didn't see that in your testing.

pchalasani:
In other words, if the first call is slightly slower that is OK as long as subsequent calls utilize the cache, so that the average slowdown per call after K calls is very small

abab-dev:
> In other words, if the first call is slightly slower that is OK as long as subsequent calls utilize the cache, so that the average slowdown per call after K calls is very small

I think if we are repeatedly calling the function, the cache lookup overhead is high than one time top level import overhead. Still not sure, will do some more analysis.

abab-dev:
@pchalasani This is ready. I'm being careful with `openai_gpt` and related aspects, so this should suffice for now. 


For more detailed import logs, we can use `pyinstrument`. It provides highly detailed profiling but comes with some overhead, making it less accurate. In contrast, the `visual_log` script gives results with no overhead since it relies on the `importtime` library.

To use `pyinstrument`, install it via:

```sh
pip install pyinstrument
```

Then, create a script for logging imports:

```sh
touch langroid_imports.py
echo "import langroid" >> langroid_imports.py
```

Run the profiler:

```sh
pyinstrument langroid_imports.py
```

To view the generated log at the end, use:

```sh
pyinstrument --load-prev 2025-03-01T22-30-23 (you get this string at the end of profiling) --html
```

pchalasani:
> This is ready. I'm being careful with `openai_gpt` and related aspects, so this should suffice for now.

Thank you! And yes that makes sense. 

If you might be interested in taking up #707  (never heard back from the person who took it), or #739 or something else, please let me know.

pchalasani:
merged in #766

PR #764, opened 2025-02-28
fix incorrect strict recovery from string-typed call to get_tool_messages

### LINKED ISSUE #753: Strict recovery should not trigger when LLM response hasn't yet been invoked.

nilspalumbo:
Extends the fix in #758 to string-typed calls to `get_tool_messages` for `ChatAgents` by ensuring that the recovery flag is sent only when one of the following is true: the most recent message in the history was produced by the LLM or the argument is a `ChatDocument` sent by the LLM.

Resolves #753

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.9.7
+  rev: v0.9.9
   hooks:
     - id: ruff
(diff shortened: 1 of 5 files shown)


pchalasani:
Thanks @nilspalumbo ! Let me know when this is ready.

nilspalumbo:
I'll add some improvements for the case where the sender is the LLM, but this version should be ready otherwise!

PR #766, opened 2025-03-01
improve imports dupe

pchalasani:
- **initial commit**
- **refactor: move imports closer to usage  location**
- **added script for visualizing import times**
- **modified visual log**
- **lazy loading for repo_chat and postgres**
- **clean up redundant check [skip ci]**
- **chainlit update type hints**


pchalasani:
@abab-dev when you get a chance can you please check my changes -- I simplified several import checks  of core dependencies of the form:

```python
try:
    import xyz
except:
    pass
```
with just `import xyz` wherever xyz is a core dependency, since such dependencies should automatically be installed with langroid.

Also inside `weaviate.py` I similarly simplified some of the `try` imports since we would expect the dependency to be already installed if we are using a method of the class (or else the constructor would have failed)

PR #771, opened 2025-03-05
Add basic anthropic sdk module - draft PR

pchalasani:


pchalasani:
@coretado the only general thought I have is -- are the OpenAI and Anthropic really so very different that we can't factor some commonalities and avoid near-duplication? I think you did a bit of this in `base.py` but I was hoping there would be more common parts (I'm not familiar with the Anthropic API at all so my expectation may be misplaced).

pchalasani:
> Is your preference more for `openai_gpt.py` housing additional functionality to handle Anthropic specific code paths, or to lift common functionality and call where appropriate for two classes representing OpenAI and Anthropic inheriting from `LanguageModel`?

Eventually it would be good to lift common functionality, but for now I think the "least disruptive" path would be to continue what you're doing, and _then_ identify commonalities. This way we don't disrupt the current `openai_gpt.py`, and just use your new code for the anthropic paths, which we would have ample time to test.

coretado:
Sounds like a plan, I shall forge ahead in that direction

coretado:
Hello @pchalasani, I hope this message finds you well. Was going to start working on the tests after adding caching patterns to the Anthropic integration, and had two clarifying questions:
- It seems like `test_llm.py` and `test_llm_async.py` are the main test modules, are these the main test files I should look to add Anthropic focused tests to?
- the listed Anthropic models present in `langroid/language_models/model_info.py` were all v3+, which I don't believe supports the `TextCompletion` APIs, and thus I have looked to leverage the `Messages` API for both chat and completion; are we looking to primarily support Claude 3.x onwards?

pchalasani:
> * should look to add Anthropic focused tests to?

Yes



> * for both chat and completion; are we looking to primarily support Claude 3.x onwards?

Don’t worry about text completion. We just need chat completion so 3+

Thanks !

pchalasani:
We would also look to test tools and agents, so you could add tests in the corresponding files as well.

coretado:
Tools and agents, got it, will look for those tests as well

coretado:
Working through testing for tools and agents, plan on adding tests for tool agents and it's async variant, as well as chat agent and it's async variant, then will request first review

coretado:
Hello @pchalasani, would you be able to take a look at this PR when you have some time? Added tests within the llm, chat agent, and tool message modules (and their async counterparts). I'm sure there are other tests we should put the anthropic SDK through, and I am happy to add whatever is needed next. Thought this would be a good stopping point given the LOC. One thought that lingers for me is that the prompting needed for Haiku to match the behavior sought after in the tool messages tests for recovery from errant tool usage was a bit more involved than it's gpt4o counterpart, is that something that is acceptable? Thank you in advance.

pchalasani:
Thank you for all the effort @coretado ! 
It may take me several days to get to review this.

pchalasani:
@coretado apologies for the long delay -- this is a substantial PR that I have to block off at least a day to review carefully, and I haven't had a chance yet.

coretado:
Hello @pchalasani, no worries, makes sense as it is a large LOC. It looks like you've been integrating MCP to Langroid 🚀, will be fun to utilize the rest of the framework with the new paradigm 👀

pchalasani:
> will be fun to utilize the rest of the framework with the new paradigm 👀

@coretado yes I’m accumulating some examples at examples/mcp , open to contributions of interesting examples there !

PR #784, opened 2025-03-16
dupe of Firecrawl PR #780

pchalasani:


pchalasani:
@abab-dev please review the changes I made - most are minor, except for an omission of a code block in your PR when you refactored `url_loader.py`:

```
      if text is None and result is not None and isinstance(result, str):
                        text = result
```
Not sure if there was a reason you skipped it.

PR #791, opened 2025-03-20
generate markdown output in .docx parser

alexagr:
Modify PythonDocxParser parser to generate markdown output.


pchalasani:
Thanks @alexagr. There are a few issues in this PR - there's an O(n^2) complexity due to the nested loop searching for elements. Also it is a bit too ad-hoc, while there is an existing good docx -> md parser from markitdown, so I'm going with this -- https://github.com/langroid/langroid/pull/795

PR #793, opened 2025-03-23
New markdown splitter option in parser.py

alexagr:
New markdown splitter option:

- splits markdown documents based on TOC (headers)
- tries to fit "complete sections" into chunk by using variable chunk size (+/- 30% from `config.chunk_size`)
- adds top-level headers (if such exist) to the beginning of the chunk - thus increasing semantic surface
- uses title / description from metadata section (if such exists) as top-level headings


pchalasani:
Thank you for this @alexagr; closing this as  I did a more extensive release - https://github.com/langroid/langroid/releases/tag/0.50.0

Please let me know if you have any ideas to improve it or concerns etc.

PR #815, opened 2025-04-12
dupe of streaming PR

pchalasani:
- **fix streaming in OpenAIGPT**
- **fix settings.quiet bug**


pchalasani:
@alexagr I made this dupe PR due to a conflict, and also noticed you had forgotten to remove "not silent" condition in the deltas part of `_process_stream_event_async`. But overall looks very good, thanks for digging into this and fixing this -- of course `quiet` mode should not disable `streaming`.

Issue #800, opened 2025-03-29
MCP Support


pchalasani:
Add lightweight adapter to translate between `ToolMessage` and an MCP server.

abab-dev:
@pchalasani I am giving this a go.

pchalasani:
done in PR #832

PR #832, opened 2025-04-27
feat: MCP tools support via FastMCP

pchalasani:
Addresses #800

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.11.6
+  rev: v0.11.7
   hooks:
     - id: ruff
--- a/docs/notes/mcp-tools.md
+++ b/docs/notes/mcp-tools.md
@@ -1,0 +1,233 @@
+# Langroid MCP Integration
+
+Langroid provides seamless integration with Model Context Protocol (MCP) servers via 
+two main approaches:
+
+1. **FastMCPClient** – programmatic creation of `ToolMessage` classes.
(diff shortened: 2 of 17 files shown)


alexagr:
@pchalasani this is awesome!
I didn't have a chance to run the new code yet - just briefly reviewed the new classes - but it looks really good.

One thing that may be a bit troublesome though - in `FastMCPClient::get_langroid_tool()` you populate `fields` with tool parameter names as specified in in the MCP schema. What will happen if some of these parameters are one of `('request', 'purpose', 'id', 'recipient')`? Won't they overwrite "base" attributes of `ToolMessage` class - and thus break the tool handling code?

NB: 'recipient' is not initialized in an attribute of `ToolMessage`, but it's checked in `Agent::_tool_recipient_match()`.

pchalasani:
> What will happen if some of these parameters are one of `('request', 'purpose', 'id', 'recipient')`? Won't they overwrite "base" attributes of `ToolMessage` class - and thus break the tool handling code?

@alexagr thanks for checking it out. Yes this is a good point. There are a ways to deal with it, but the simplest might be to add `"__"` suffix to the MCP schema fields when they clash, and add the extended fields to the Langroid tools. Then LLM will generate these extended field names, and at tool-handling time, in `call_mcp_tool` I can remove these suffixes from `arguments` before calling the MCP tool

pchalasani:
@alexagr fixed in v0.53.2 see the updated tests involving `InfoTool` https://github.com/langroid/langroid/blob/main/tests/main/test_mcp_tools.py

VibeGrind:
TypeError: Client.__init__() got an unexpected keyword argument 'read_timeout_seconds'

PR #841, opened 2025-05-08
Code Injection Mitigation

SCH227:
Added mitigations for most common attack vectors, and warnings so users can use langroid safely.

  - Introduced a `sanitize_command` function in `langroid/utils/pandas_utils.py` to validate and sanitize user-provided expressions before execution. I took the most common methods used for Pandas Dataframes, and stripped the ones that may run code. Using a whitelist approach.
  
  - Updated `pandas_eval` in `TableChatAgent` and `compute_from_docs` in `VectorStore` to use `sanitize_command` for sanitizing expressions before evaluation by default.

 - Added a `full_eval` flag to `TableChatAgentConfig` to toggle input sanitization. By default, it is enabled.

  - And added some tests.

@pchalasani looking forward your thoughts!

### DIFF
--- a/langroid/agent/special/table_chat_agent.py
+++ b/langroid/agent/special/table_chat_agent.py
@@ -121,7 +121,9 @@ class TableChatAgentConfig(ChatAgentConfig):
     cache: bool = True  # cache results
     debug: bool = False
     stream: bool = True  # allow streaming where needed
-    full_eval: bool = False  # runs eval without sanitization. Use only on trusted input!
+    full_eval: bool = (
+        False  # runs eval without sanitization. Use only on trusted input!
+    )
     data: str | pd.DataFrame  # data file, URL, or DataFrame
     separator: None | str = None  # separator for data file
     vecdb: None | VectorStoreConfig = None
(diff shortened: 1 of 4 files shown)


SCH227:
@pchalasani thank you! 
The tests were added for the lower level `pandas_utils.sanitize_command()`, which is imported by both `vector_stores/base.py` and `agent/special/table_chat_agent.py`.
I think adding tests for `VectorStore.compute_from_docs()` wouldn't add anything as `calc` string flows directly to `sanitize_command()` without prior modifications. Am I missing something?

pchalasani:
> @pchalasani thank you! The tests were added for the lower level `pandas_utils.sanitize_command()`, which is imported by both `vector_stores/base.py` and `agent/special/table_chat_agent.py`. I think adding tests for `VectorStore.compute_from_docs()` wouldn't add anything as `calc` string flows directly to `sanitize_command()` without prior modifications. Am I missing something?

sounds good. but please build on top of the PR #850 that includes some fixes

SCH227:
No no, I meant, the tests were already added in the first PR #841

pchalasani:
Ah I see got it. 
We have to see why those other tests are failing. I haven't looked yet

pchalasani:
closing since dupe #850 was merged

PR #843, opened 2025-05-10
Mcp code execution example using deno pyodide runtime

abab-dev:
Address #842
Make sure you have installed deno and then  
Test out using
`uv run examples/mcp/pyodide_code_executor.py`


docs/notes/pyodide_code_executor.md
```diff
@@ -0,0 +1,14 @@
+# Running the pydiode_code_executor.py example
```
pchalasani, line 1: there's probably no need for this doc, if we're not saying anything about what mcp server it's using etc

examples/mcp/pyodide_code_exector.py
```diff
@@ -0,0 +1,71 @@
+from fastmcp.client.transports import StdioTransport
```
pchalasani, line 1: fix typo in file name: exector.py -> executor.py

examples/mcp/pyodide_code_exector.py
```diff
@@ -0,0 +1,71 @@
...
+from fire import Fire
```
pchalasani, line 2: we can put whatever you had in the .md doc file, at the top of this file as a docstring instead, as we normally do in all the examples.

langroid/embedding_models/models.py
```diff
@@ -144,7 +144,7 @@ def __call__(self, input: List[str]) -> Embeddings:
             # Process in batches
             for batch in batched(truncated_texts, self.batch_size):
                 result = self.embed_model.client.embeddings.create(
-                    input=batch, model=self.embed_model.config.model_name  # type: ignore
+                    model=self.embed_model.config.model_name,  # type: ignore
```
pchalasani, line 147: the `input=batch` arg is missing

langroid/embedding_models/models.py
```diff
@@ -532,7 +532,17 @@ def generate_embeddings(self, texts: List[str]) -> List[List[float]]:
                 )
 
             # Extract .values from ContentEmbedding objects
-            all_embeddings.extend([emb.values for emb in result.embeddings])
+            # Use a list comprehension with a filter
+            batch_embeddings: List[List[float]] = [
```
pchalasani, line 536: if we're dopping Nones then the result length will be smaller than the number of strings passed in. 
Are callers of this fn going to be ok with this? I didn't check

pchalasani: maybe best to put an assert in here that we don't expect Nones, that should appease mypy I think, and get rid of the lint err

abab-dev:
Done !
Now I should close the docker_code_executor PR or keep it ..

pchalasani:
> Now I should close the docker_code_executor PR or keep it ..

Please keep it, will look at it later.

The pyodide mcp works great!

pchalasani:
@abab-dev I merged a dupe of this , I removed the embedding mdl changes. Thanks for this!

PR #850, opened 2025-05-17
fixes to code-injection mitigation PR #841

pchalasani:
- **Update SECURITY.md**
- **Add eval commands sanitizer**
- **Sanitizer tests**
- **Mitigation for CVE-2025-46725**
- **Mitigation for CVE-2025-46724**
- **Update langroid/utils/pandas_utils.py**
- **Update langroid/agent/special/table_chat_agent.py**
- **fix: add type annotations and fix bugs in code injection mitigation**


pchalasani:
@SCH227 several tests are failing, related to your fixes

SCH227:
Just checked. 
The tests should be run with `full_eval=True` and bypass sanitization. Tests data is trusted input. That should fix all errors.

### Why this errors?
I added custom `UnsafeCommandError` error messages to know exactly where the `calc` string triggers an unsafe check. For all the tests, the error message is `subscript must be literal`: https://github.com/SCH227/langroid/blob/d99d24c207ef6dab17728e8a571ee1e6ce470f29/langroid/utils/pandas_utils.py#L233-L237

That check is saying _everything inside `df[...]` must be a literal index or slice._ This is to avoid arbitrary code can be run inside the brackets, and to protect from some resource exhaustion attack vectors. 
For example, a malicious command like `df[(df.pipe(lambda d: d.eval("import os; os.system('rm -rf /')")))]` is blocked by this check.
The drawback is that a valid query like `df[(df['gender'] == 'Male')]['income'].mean()` is blocked too, similarly as what happened in table chat agent tests. But this is solvable: the current behavior lets user run what they need by changing a flag, while still keeping safe defaults:
- By default this attack vector is blocked
- If user needs more functionality and runs on trusted input, they can set `full_eval=True`

pchalasani:
@SCH227 merged and released new verison, thank you for your help on this!

SCH227:
Thanks to you! I will update the advisory so it is ready for publishing

PR #857, opened 2025-05-19
docs: update llm-pdf-parser.md to render correctly [skip-ci]

abab-dev:
image

lists are not getting rendered correctly


pchalasani:
@abab-dev with material markdown, you have to have an empty line before starting the bullet lists, otherwise the list doesn't render

PR #861, opened 2025-06-12
MCP updates

nilspalumbo:
Allows `FastMCPServer` to forward image context (enabled by default) and resources (disabled by default) from the MCP server to the LLM.

Adds optional persistence between tool calls to the MCP server connection, needed by certain servers (such as Playwright) to maintain state, and adds a simple example of Wikipedia question answering with Playwright.


langroid/agent/tools/mcp/fastmcp_client.py
```diff
@@ -209,39 +235,58 @@ async def get_tool_async(self, tool_name: str) -> Type[ToolMessage]:
...
             # open a fresh client, call the tool, then close
             async with FastMCPClient(**client_cfg) as client:  # type: ignore
-                return await client.call_mcp_tool(self.request, payload)
+                return await client.call_mcp_tool(itself.request, payload)
 
         tool_model.call_tool_async = call_tool_async  # type: ignore
 
-        if not hasattr(tool_model, "handle_async"):
-            # 3) define an arg-free handle_async() method
+        if not hasattr(tool_model, "response_async"):
```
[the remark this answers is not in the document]
pchalasani: I added the assert

pchalasani:
@nilspalumbo I added a fix to the playwright example since it wasn't working for me initially, plus added some debugging and progress-bar to show the container building. Also added tests.

And I added back the `handle_async` method since we should keep that, in addition to the `response_async`. 

Also added a much simpler playwright example at `examples/mcp/playwright-mcp.py` -- doesn't need docker, and directly uses `NPXStdioTransport`, and it works fine on the wikipedia query.

Can you take a look and let me know if it looks fine?

pchalasani (changes requested):
looks great, added tests, fixed the example, and added tests for the image/resource/blob forwarding, persistent connection, and response_async

pchalasani:
@nilspalumbo I'm realizing it doesn't make sense to have both `handle_async` and `response_async` methods, since langroid will always only use the first one. So I overhauled this to only create a `handle_async` method, and also generalized `handle`, `handle_async` to accept an `agent` arg.

pchalasani:
Also added `examples/mcp/puppeteer-mcp.py` which is exactly same except uses Puppeteer MCP, and also works fine (no docker)

nilspalumbo:
Thanks! The example now requires manual browser installation, as Playwright's script only works on Debian-based distros, but it keeps it simple, the standard version of chrome seems to work for the new version of the MCP server. The Puppeteer example fails for me, but it's likely a platform issue.

nilspalumbo:
The new flexible handler processing is nice! It might fail if users do something non-standard (like using another name for `self`), though, and it might be better to user the non-string annotations when available to better handle agent arguments which don't have `Agent` in their class names.

pchalasani:
> The example now requires manual browser installation, as Playwright's script only works on Debian-based distros, but it keeps it simple

didn't quite understand :)  are you referring to your original playwright mcp example?

nilspalumbo:
I'm referring to the new one (the old docker image did all the setup for system dependencies, browser installation, etc.), the agent can (ignoring authentication) run the scripts for browser installation in the new example, but those do not work on my system.

Not a major issue though, it seems that the MCP server is now compatible with the current version of Chrome, so it's working without Docker when Chrome is installed.

pchalasani:
> it might be better to user the non-string annotations when available to better handle agent arguments which don't have `Agent` in their class names.

@nilspalumbo The problem is that the processing of the `handle_async` is not happening at run time; it's at definition time, so we can't use `instanceof`, and have to rely on strings. Or am I not understanding?

nilspalumbo:
This is happening at run time, so that isn't an issue; I pushed an initial version of the handler generation code which uses subclassing to a new copy of this branch.

examples/mcp/playwright/playwright-wikipedia.py
```diff
@@ -84,14 +84,34 @@ async def main(question: str = "", model: str = ""):
...
 
     try:
         client.images.get(tag)
+        print(f"Using existing Docker image: {tag}")
     except docker.errors.APIError:
+        print(
+            f"Building Docker image {tag}... (this may take several minutes on first run)"
+        )
         path = os.path.dirname(os.path.abspath(__file__))
-        client.images.build(path=path, tag=tag)
```
ondonga254-code, line 94: Tr:
"""Builds runs the playwright-mcp Docker Image with real time log streaming and macOS specific socket handling"""

examples/mcp/playwright/playwright-wikipedia.py
```diff
@@ -100,9 +120,29 @@ async def main(question: str = "", model: str = ""):
...
+        logs = container.logs()
+        print(f"[red]Container logs:\n{logs.decode()}")
+        container.remove()
+        return
+
+    print("Container is running, attempting to connect...")
+
     # Wait for the server to start
     connected = False
-    while not connected:
```
ondonga254-code, line 105: # try connecting to the server with a cap to prevent infinite hanging if the container fails to initialize.

langroid/agent/tools/mcp/fastmcp_client.py
```diff
@@ -281,9 +282,21 @@ async def call_tool_async(itself: ToolMessage) -> Any:
...
+                # For backward compatibility, return just the text content
+                content, files = response
+                return content
+
+            # add the handle_async() method to the tool model
+            tool_model.handle_async = handle_async  # type: ignore
+
         if not hasattr(tool_model, "response_async"):
-            # 3) define an arg-free response_async() method
-            # if the tool model doesn't already have one
```
ondonga254-code, line 286: #Ensure the response includes any necessary file attachments or binary data by returning a ChatDocument with blob resources.

langroid/agent/tools/mcp/fastmcp_client.py
```diff
@@ -387,7 +400,10 @@ def _convert_tool_result(
                     and self.forward_blob_resources
                 ):
                     results_file.append(
-                        FileAttachment.from_io(BytesIO(b64decode(item.resource.blob)))
```
ondonga254-code, line 390: # Ensure the attachment carries its origninal MIME type for proper client-side handling
results_file.append(
      FileAttachment.from_io(
           BytesIO(b64decode(item.resource.blob)),
            mime_type=item.resource.mimeType,

tests/main/test_mcp_tools.py
```diff
@@ -9,7 +9,14 @@
     SamplingParams,
 )
 from fastmcp.client.transports import NpxStdioTransport, UvxStdioTransport
-from mcp.types import ImageContent, TextContent, Tool
```
ondonga254-code, line 12: "Updating mcp.types imports to include BlobResourceContents , EmbeddedResource , and TextResourceContents for enhanced testing of resource based interactions"

examples/mcp/playwright/playwright-wikipedia.py
```diff
@@ -84,14 +84,34 @@ async def main(question: str = "", model: str = ""):
     which is shown first when sorted by year.
     """
     )
-    client = docker.from_env()
```
ondonga254-code, line 87: "Catch DockerException on from_env() to mannually point to the macOS Docker Desktop Socket at ~/.docker/run/docker.sock."

ondonga254-code (commented):
Suggested comments to fx the pending errors.

Issue #867, opened 2025-06-20
Support assignment expressions for data cleaning in pandas_eval (TableChatAgent)


parvk11:
Hello, 

I've been exploring using the `TableChatAgent` for performing data cleaning tasks over datasets. It's an awesome tool for interaction with pandas dataframes, and the ability to generate expressions via `pandas_eval` is incredibly useful.

However, I ran into an issue when trying to perform in-place modifications to the dataframe, such as removing characters from a column:

df['airline'] = df['airline'].str.replace('*', '', regex=False)

This type of expression throws a SyntaxError under the current implementation of pandas_eval, since it uses eval() which cannot handle assignment statements.

**Proposed Solution**

To allow such expressions (especially important for real-world data cleaning), I made a small modification to support an option for full_exec, and then implementing something like this to support assignments and more complex transformations in trusted environments. This allows falling back to exec() when eval() raises a SyntaxError, like so:

 ```
try:
    try:
        if not self.config.full_eval:
            exprn = sanitize_command(exprn)
        code = compile(exprn, "<calc>", "eval")
        eval_result = eval(code, vars, {})
    except SyntaxError as e:
        if self.config.full_exec:
            # If full_exec is enabled, fallback to exec
            code = compile(exprn, "<calc>", "exec")
            exec(code, vars, {})
            eval_result = ""  # exec has no return value
        else:
            raise e  # re-raise if full_eval is not enabled
except Exception as e:
    eval_result = f"ERROR: {type(e).__name__}: {e}"
```

I think this could be useful for using the TableChat agent to also be able to clean/transform data in data analysis workflows. 

Another option might be to build a dedicated DataCleaningAgent for work like this.

Let me know if you'd like a PR for this or if there's a preferred direction to support data mutation.

Thanks for the great work on Langroid!

pchalasani:
Hi @parvk11, thanks for raising this issue!

  After careful consideration of the security implications, we've implemented a solution that addresses your use case while maintaining
  the security model of TableChatAgent.

  What we've done

  We've created PR #869 that enables data cleaning operations through the following changes:

  1. Added df.assign() to the whitelist - This method is now allowed in pandas expressions, which provides a safe way to create modified
  DataFrames without the security risks of arbitrary assignments.
  2. Updated the agent's system message to:
    - Proactively inform the agent that assignment statements are not allowed
    - Guide it to use df.assign() instead when data modifications are needed
    - Provide clear examples of the correct approach

  Why we can't support direct assignment statements

[...]

PR #869, opened 2025-06-21
feat: Enable df.assign() in TableChatAgent for data cleaning operations

pchalasani:
## Summary
- Enables the `df.assign()` method in TableChatAgent to support data cleaning operations
- Updates system message to clarify assignment restrictions and guide proper usage
- Adds test coverage for self-correction behavior

## Context
This PR addresses #867 where users wanted to perform data cleaning operations like removing characters from columns. Previously, assignment statements like `df['col'] = df['col'].str.replace()` would fail with a SyntaxError since `eval()` cannot handle assignments.

## Changes
1. **Added `assign` to the whitelist** (`COMMON_USE_DF_METHODS`) - This method is safe as it returns a new DataFrame without side effects
2. **Updated TableChatAgent system message** to:
   - Proactively explain that assignment statements are not allowed for security
   - Guide users to use `df.assign()` instead
   - Provide clear examples of the correct approach
3. **Added test** to verify the agent self-corrects and successfully uses `df.assign()`

## Security Considerations
The `assign` method is safe to whitelist because:
- It's a pure method that returns a new DataFrame
- It can only create/modify columns within the DataFrame
- It doesn't allow arbitrary code execution, file I/O, or network access
- The expressions passed to `assign` still go through the same sanitization

## Test plan
- [x] Added `test_table_chat_agent_assignment_self_correction` to verify the behavior
- [x] Test passes - agent successfully uses `df.assign()` to clean data
- [x] All existing tests pass
- [x] Linting and type checking pass

### DIFF
--- a/langroid/agent/special/table_chat_agent.py
+++ b/langroid/agent/special/table_chat_agent.py
@@ -50,6 +50,9 @@ Here is a summary of the dataframe:
 Do not assume any columns other than those shown.
 In the expression you submit to the `pandas_eval` tool/function, 
 you are allowed to use the variable 'df' to refer to the dataframe.
+IMPORTANT: You can only use expressions that return a value - assignment statements 
+(like df['col'] = value or temp = df['col']) are NOT allowed for security reasons.
+To modify data, use methods like df.assign() that return a new dataframe.
 
 Sometimes you may not be able to answer the question in a single call to `pandas_eval`,
 so you can use a series of calls to `pandas_eval` to build up the answer. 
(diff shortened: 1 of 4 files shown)


pchalasani:
@SCH227 would you kindly look at this from a security pov -- we've added "assign" to the whitelisted cmds, I think that should be fine?

SCH227:
@pchalasani I added looking into this to my TODO. Will communicate you my findings in this thread or via email

SCH227:
Done, sent you DM via Linkedin

PR #888, opened 2025-07-06
make `max_tokens` and `temperature` in OpenAICallParams optional

alexagr:
**Change description**

Make `max_tokens` and `temperature` parameters in `OpenAICallParams` optional.

**Rationale**

`OpenAICallParams` are intended to be used for configuring additional LLM parameters. For example I'm using them to configure `reasoning_effort` for Gemini 2.5 models. However if `OpenAICallParams` contains numeric values for `max_tokens` and `temperature`, as they currently do by default, these values will override corresponding values from model's `config` element - see `OpenAIGPT._openai_api_call_params()`. This IMO is pretty counter-intuitive and rather limiting - as this means that I need to duplicate values from `llm.config` into the `llm.params`.

### DIFF
--- a/langroid/language_models/openai_gpt.py
+++ b/langroid/language_models/openai_gpt.py
@@ -203,8 +203,8 @@ class OpenAICallParams(BaseModel):
     https://platform.openai.com/docs/api-reference/chat
     """
 
-    max_tokens: int = 1024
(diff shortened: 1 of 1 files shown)


pchalasani:
@alexagr merged

PR #896, opened 2025-07-10
improvement in logic while importing crawl4aiconfig

abab-dev:
Crawl4ai imports in config are contributing to increased import time 
so improved them with little change in logic while import happens


pchalasani:
@abab-dev  merged in #897 with a small fix in test

PR #910, opened 2025-08-24
allow use of reasoning_effort for Gemini-2.5 models

### LINKED ISSUE #909: "reasoning_effort" parameters is ignored for Gemini models
Gemini 2.5 models support `reasoning_effort` parameter in OpenAI compatibility API that controls model's "thinking" capability.

However when I pass it via `OpenAIParams` to `OpenAIGPTConfig` - for example as follows - it never actually gets being sent:
```
    params = OpenAICallParams(reasoning_effort='none')
    llm_cfg = OpenAIGPTConfig(
        chat_model=chat_model,
        api_key=model_config.api_key,
        params=params,
        ...
    )
```

The problem is caused by the `unsupported_params()` function in `OpenAIGPT` - that uses `OpenAI_API_ParamInfo` to remove some parameters passed to the OpenAIGPTConfig(). The later explicitly lists "reasoning_effort" as applicable to `o3-mini` model only:
```
    # model-specific params at top level
    params: Dict[str, List[str]] = dict(
        reasoning_effort=[
            OpenAIChatModel.O3_MINI.value,
        ],
    )
```

We could add Gemini-2.5 models to this list. But I'm not sure whether it makes sense to whitelist OpenAI options at all. Especially, `reasoning_effort` - which is passed via `OpenAICallParams` - i.e. someone needs to EXPLICITLY BOTHER to do it, and then langroid just silently discards it.

And while you are there, can you please make `OpenAICallParams.n` optional and default to `None`? I want to be able to just specify `reasoning_effort` - and not include any other additional parameters in the OpenAI API call.

alexagr:
Fix #909:
* fix names of Gemini models - as per https://ai.google.dev/gemini-api/docs/models
* allow `reasoning_effort` for Gemini 2.5 models - as per https://ai.google.dev/gemini-api/docs/openai
* make `n` parameter optional in `OpenAICallParams`

### DIFF
--- a/langroid/language_models/model_info.py
+++ b/langroid/language_models/model_info.py
@@ -74,10 +74,10 @@ class GeminiModel(ModelName):
     GEMINI_1_5_PRO = "gemini-1.5-pro"
     GEMINI_2_5_PRO = "gemini-2.5-pro"
     GEMINI_2_5_FLASH = "gemini-2.5-flash"
-    GEMINI_2_5_FLASH_LITE_PREVIEW = "gemini-2.5-flash-lite-preview-06-17"
+    GEMINI_2_5_FLASH_LITE = "gemini-2.5-flash-lite"
     GEMINI_2_PRO = "gemini-2.0-pro-exp-02-05"
     GEMINI_2_FLASH = "gemini-2.0-flash"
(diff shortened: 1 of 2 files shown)


pchalasani:
thanks @alexagr 
will merge as soon as tests pass

PR #915, opened 2025-08-31
Allow dropping messages from the middle of call history

alexagr:
In some scenarios there is a need to drop messages - e.g. tools calls / responses - from the middle of message history.
`ChatAgent.clear_history()` method does this properly - dropping tool call id's and deleting the ChatDocuments - but it currently only supports "truncating" the message history.
This PR adds support for `end` parameter - that enables use of `clear_history()` method for this scenario too, while preserving backwards compatibility.


pchalasani:
done in PR #917

PR #916, opened 2025-08-31
support reasoning_effort parameter for OpenAI thinking models

alexagr:
`reasoning_effort` parameter is used to control thinking level of various OpenAI models.
However current implementation "whitelists" it only for `o3-mini` model - and not, for example, for `gpt-5` models.
This PR fixes the issue.


pchalasani:
done in PR #917

PR #920, opened 2025-09-27
adhere to llama.cpp's /embedding endpoint data format

sidietz:
Llama.cpp's /embedding endpoint supports a batch mode to process more than one embedding at once. Hence, an array of embedding objects is returned. To support llama.cpp's /embedding properly, this has to be taken into account.

### DIFF
--- a/langroid/embedding_models/models.py
+++ b/langroid/embedding_models/models.py
@@ -463,7 +463,7 @@ class LlamaCppServerEmbeddings(EmbeddingModel):
         response = requests.post(self.embedding_url, json=data)
 
         if response.status_code == 200:
-            embeddings = response.json()["embedding"]
+            embeddings = response.json()[0]["embedding"][0]
             if not (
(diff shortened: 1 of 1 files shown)


pchalasani:
thanks @sidietz I'm handling it in this PR #922

pchalasani:
Thanks for identifying this issue! However, there are a few concerns with this approach:

1. The double indexing `[0]["embedding"][0]` extracts a single float, not the full embedding vector
2. This would fail the validation on line 468 which expects a list of floats
3. llama.cpp can return multiple different response formats depending on the endpoint and configuration

We're addressing this more comprehensively in #922 with a robust solution that:
- Handles 5 different llama.cpp response formats automatically
- Includes comprehensive unit tests
- Is backwards compatible with existing deployments

Feel free to check out #922 and test it with your setup!

pchalasani:
closing, see https://github.com/langroid/langroid/releases/tag/0.59.10

pchalasani:
@codex review

PR #939, opened 2026-01-08
fix chat_doc "leak" in llm_response_messages

alexagr:
refactor `llm_response_messages` / `llm_response_messages_async` so that they don't create unneeded temporary `ChatDocument` instances just for the sake of calling callbacks and "polluting" ObjectRegistry.

### DIFF
--- a/langroid/agent/chat_agent.py
+++ b/langroid/agent/chat_agent.py
@@ -1858,17 +1858,7 @@ class ChatAgent(Agent):
                 function_call=fun_call,
                 response_format=output_format,
             )
-        if self.llm.get_stream():
-            self.callbacks.finish_llm_stream(
-                content=str(response),
-                is_tool=self.has_tool_message_attempt(
-                    ChatDocument.from_LLMResponse(response, displayed=True),
-                ),
-            )
-        self.llm.config.streamer = noop_fn
-        if response.cached:
-            self.callbacks.cancel_llm_stream()
-        self._render_llm_response(response)
+
         self.update_token_usage(
             response,  # .usage attrib is updated!
             messages,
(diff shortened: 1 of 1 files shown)


pchalasani:
@alexagr Thanks for this fix! The ChatDocument leak in `llm_response_messages` was a real issue - having temporary objects accumulate in the `ObjectRegistry` is definitely something that needed addressing.

I've created PR #949 that builds on top of your changes with a couple of additions:

1. **Bug fix**: Found a small issue where `_render_llm_response` would print `"None"` when called without an `LLMResponse` (e.g., from `doc_chat_agent.py`). Changed `str(response)` to `str(response or chat_doc)` to handle this case.

2. **Regression test**: Added `test_llm_response_messages_no_registry_leak` that verifies exactly one `ChatDocument` is created per `llm_response_messages` call, which should help prevent this type of leak from being reintroduced in the future.

Once your PR is merged, we can rebase #949 on top of it (or merge them together).

pchalasani:
@alexagr Thanks again for identifying this issue! After reviewing the Codex feedback, we've decided to take a different approach.

### Responding to Codex Review

The Codex review raised a valid concern: by reusing the same `chat_doc` for the `has_tool_message_attempt()` callback, we introduce a mutation side effect. The `get_tool_messages()` method caches parsed tools in `chat_doc.tool_messages`, which can cause issues in multi-agent scenarios where the document is forwarded to other agents - they'd see the cached (and potentially incomplete) tool list instead of re-parsing.

### Alternative Approach

Instead of restructuring the code to reuse objects, PR #950 keeps the original code flow but explicitly cleans up the temporary `ChatDocument` objects:

```python
temp_doc = ChatDocument.from_LLMResponse(response, displayed=True)
self.callbacks.finish_llm_stream(
    content=str(response),
    is_tool=self.has_tool_message_attempt(temp_doc),
)
ObjectRegistry.remove(temp_doc.id())  # Explicit cleanup
```

This approach:
- Fixes the registry leak (your original goal)
- Preserves the original behavior (no mutation side effects)
- Is explicit about the cleanup intent

**We'll be closing this PR in favor of #950.** Your contribution in identifying the leak was valuable - thank you!

PR #940, opened 2026-01-08
improve tool parsing

alexagr:
Two improvements in tool parsing logic inside `Agent.get_tool_messages()`:

- `msg.all_tool_messages` is used for indication that we already parsed specific message regardless of whether tool calls exist in it or not
  - we set it to None in `__init__` and then change it to `List` - either empty or with some tools upon the first call to `get_tool_messages`
  - this prevents continuous parsing attempts for the same message, as different parts call, for example, to `has_tool_message_attempt()`

- new configuration parameter `search_for_tools_everywhere` in `ChatAgentConfig`
  - it's set to `False` for backwards compatibility
  - when set to `True` - tools will be searched only in the parts of the message that match agent configuration
    - this prevents unneeded (and potentially dangerous) parsing of message content for OpenAI models that use Tools or Function APIs

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.14.6
+  rev: v0.14.10
   hooks:
     - id: ruff
--- a/langroid/agent/base.py
+++ b/langroid/agent/base.py
@@ -7,6 +7,7 @@ import re
 from abc import ABC
 from collections import OrderedDict
 from contextlib import ExitStack
+from enum import Enum
 from types import SimpleNamespace
 from typing import (
     Any,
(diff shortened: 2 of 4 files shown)


pchalasani:
Thanks for the explanation @alexagr. You're right that in the common case, the agent that invokes the LLM will be the first to parse the response.

However, Langroid explicitly supports multi-agent message passing through orchestration tools like `ForwardTool`, `SendTool`, and `PassTool`. In these scenarios, `ChatDocument` objects can flow between agents with different tool registries. For example:

1. Agent A (with ToolX enabled) receives an LLM response containing ToolY JSON
2. Agent A parses it, finds no tools it handles → caches `all_tool_messages = []`
3. Agent A forwards the message to Agent B (which has ToolY enabled)
4. Agent B sees the cached empty list and skips parsing entirely
5. Agent B misses ToolY that it could have handled

Requiring users to manually reset `all_tool_messages = None` before forwarding is error-prone and leaks implementation details.

We've addressed this in a follow-up PR by adding an `all_tool_messages_agent_id` field that tracks which agent populated the cache. The cache is only used if the same agent is parsing again - different agents will re-parse with their own tool registry. This preserves your performance optimization for the common case while ensuring correctness in multi-agent orchestration.

pchalasani:
superseded by #948

PR #945, opened 2026-01-11
adjust shape of show_llm_response / finish_llm_stream callbacks

alexagr:
### Problem description

`show_llm_response` and `finish_llm_stream` callbacks currently include `content=str(response)` which mixes text `message` generated by model with serialization of `functions / tools`. While this may be fine for a simple `chainlit` integration, this is not "good enough" for applications that use these callbacks to handle model responses. For example, I use these callbacks in my application to convert the text generated by model to voice - and if model generates some invalid function / tools call, `content` gets populated with JSON string, while `is_tool` is False - and I'm "playing" to user JSON string - definitely not an expected behavior...

### Proposed solution

Change shape of `show_llm_response` and `finish_llm_stream` to include separate `content` and `tools_content` parameters.
The former is always populated with `response.message`. The later is populated with serialization of `functions / tools` if such exist; or empty string otherwise.

This gives enough information for applications that need to know THE EXACT information that was present in the LLM response. While still keeping `chainlit` integration logic very straightforward - see the diff in `chainlit.py` file.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.14.6
+  rev: v0.14.11
   hooks:
     - id: ruff
--- a/langroid/agent/callbacks/chainlit.py
+++ b/langroid/agent/callbacks/chainlit.py
@@ -356,7 +356,7 @@ class ChainlitAgentCallbacks:
             run_sync(self.stream.remove())  # type: ignore
         else:
             run_sync(self.stream.update())  # type: ignore
-        stream_id = self.stream.id if content else None
+        stream_id = self.stream.id if tools_content or content else None
         step = cl.Message(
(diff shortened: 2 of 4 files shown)


pchalasani:
@alexagr Thanks for this excellent contribution! The separation of `content` and `tools_content` in callbacks is a great improvement - it solves a real problem for applications that need to distinguish between text messages and tool call serializations.

We've merged PR #952 which includes all your changes plus some additional fixes discovered during review:

1. **Tests for `tools_content()`** - Added comprehensive test coverage for the new method
2. **Stream removal fix** - Fixed the logic to check both `content` and `tools_content` before removing the stream (tool-call responses often have empty message content)
3. **Type handling fix** - Fixed `_render_llm_response` to handle both `LLMResponse` and `ChatDocument` types (it was calling `.message` and `.tools_content()` on `ChatDocument` which doesn't have those attributes)

This is now released in **v0.59.26**.

Closing this PR in favor of #952. Thanks again for the contribution!

PR #946, opened 2026-01-12
add gpt-5.2 and gemini-3-flash-preview models

alexagr:
add `ModelInfo` for latest models from OpenAI and Google:
- gpt-5.2
- gpt-5.2-pro
- gemini-3-flash-preview

fix `has_tools` configuration for gpt-5 models (the code currently doesn't really use it - but let's not confuse ourselves)

### DIFF
--- a/langroid/language_models/model_info.py
+++ b/langroid/language_models/model_info.py
@@ -43,6 +43,8 @@ class OpenAIChatModel(ModelName):
     GPT5_1 = "gpt-5.1"
     GPT5_1_CODEX = "gpt-5.1-codex"
     GPT5_1_CODEX_MINI = "gpt-5.1-codex-mini"
+    GPT5_2 = "gpt-5.2"
+    GPT5_2_PRO = "gpt-5.2-pro"
     GPT_OSS_120b = "gpt-oss-120b"
     GPT_OSS_20b = "gpt-oss-20b"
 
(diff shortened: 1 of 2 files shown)


pchalasani:
@claude review this

alexagr:
@pchalasani  please note that I pushed another fix - adding missing info for latest Claude and OpenAI OSS models.

I also removed `OpenAI_API_ParamInfo()` - and instead populated `unsupported_params` in relevant models. Rationale is very simple - use of `OpenAI_API_ParamInfo()` makes it impossible for application to easily add `ModelInfo` for a model that supports "reasoning_effort" parameter; for example for Mistral models I had to add parameter as "custom_reasoning_effort" in my code and then map it back to "reasoning_effort" via `model_info.rename_params` - pretty convoluted way to do it.

Finally, I removed `rename_params={"max_tokens": "max_completion_tokens"}` for OpenAI models - this is a legacy code from the time when `OpenAIGPT::_prep_chat_completion` used `max_tokens` in `args` - but it for a while been using `max_completion_tokens` so the rename is no longer relevant.

pchalasani:
Thanks for identifying the issue with the allowlist approach @alexagr!

We've taken a slightly different approach in PR #951 that achieves the same goal while maintaining backward compatibility:

Instead of removing `OpenAI_API_ParamInfo`, we made the allowlist checks **lenient for unknown models**. The logic now checks if the model is in `MODEL_INFO` before applying the allowlist restriction:

- **Known models**: Existing behavior (allowlist enforced)
- **Unknown/custom models**: All params allowed by default

This way:
1. Users with custom models can use `reasoning_effort` and other params without modifying library code
2. `OpenAI_API_ParamInfo` remains available for anyone who was importing/using it
3. No breaking changes for existing users

See PR #951 for details.

pchalasani:
**Update:** We've now incorporated the model additions from this PR into #951:

- GPT-5.2, GPT-5.2-Pro, GPT-OSS models
- Gemini 3 Flash/Pro
- Claude 4 family
- Fixes for `has_tools` and `rename_params` on GPT-5 models

PR #951 now combines the lenient allowlist fix with all your model updates, while keeping `OpenAI_API_ParamInfo` for backward compatibility.

pchalasani:
closing in favor of #951

PR #954, opened 2026-01-21
support Vertex AI for Gemini models

alexagr:
Google Vertex AI uses project-specific URLs for OpenAI Compatibility layer (which is different from fixed URL used by Google API) - see https://docs.cloud.google.com/vertex-ai/generative-ai/docs/start/openai for details.

This simple PR adds support for Vertex AI by allowing specification of `api_base` via `OpenAIGPTConfig` for Gemini models.


pchalasani:
@alexagr closing in favor of #960 which has a fix for what Codex flagged above. Let me know your thoughts there

PR #955, opened 2026-01-21
replace deprecated parseString

alexagr:
replace deprecated `parser.parseString` with `parse_string`


pchalasani:
included in another PR

PR #958, opened 2026-01-25
configuration for content based routing of LLM responses

alexagr:
Add `content_based_routing` attribute to `ChatAgentConfig` that controls whether we look for "routing instructions" - e.g. "TO: destination" in textual part of LLM response. For scenarios where we use proper orchestration tools, such parsing can be unneeded and even dangerous - as user may trick LLM into generating these pre-defined strings and thus affecting task orchestration logic.


pchalasani:
@alexagr thank you for this -- I had minor mods on top of this, see PR #961

PR #959, opened 2026-01-25
Though signatures in message history

alexagr:
Many reasoning LLMs include "thought signatures" in textual response, for example:
- Amazon Nova and Nova 2 models (when configured for extended thinking via `reasoning_effort`)
- Gemini 3 Flash (with `thought_tag_marker` configured via `extra_body`)

Typical response from Gemini 3 Flash looks like this:
```
<thinking>
**Assessing My Capacities**

I've been mapping out the appropriate response to the user's question, aiming for brevity and clarity. The core is a succinct outline of my abilities, like answering questions and giving info. I'm keeping the tone polite, and also mentally noting when I should consider using the `transfer_call` feature.
</thinking>

I can answer your questions, provide information, and help with various tasks. I also have the ability to transfer your call to a specialist if you need further assistance.
```

Keeping "thought signatures" in message context is crucial for maintaining model's ability to reason / make function calls as the conversation progresses across turns.

Current `langroid` implementation extracts `message` and `reasoning` from the actual LLM response - using `get_reasoning_final()` - and populates corresponding fields in `LLMResponse` instance. These two fields get copied into the `ChatDocument` - in `from_LLMResponse` method. But conversion to `LLMMessage` - in `to_LLMMessage` method - uses only `content`. This essentially means that "though signatures" get lost and never "make it" into the message history.

My solution is fairly straightforward:
* add `message_with_reasoning` field to `LLMResponse` and populate it with **original** text received from the LLM if we "parse" it into "message" and "reasoning"
* add similar `content_with_reasoning` field to `ChatDocument` and copy to it data from `LLMResponse` - in `from_LLMResponse` method
* finally, use `content_with_reasoning` - if present - when constructing the `LLMMessage` - in `to_LLMMessage` method

It adds minimal amount of memory - as `message_with_reasoning` stays `None`, unless we **actually** modify original LLM response. All existing APIs of `LLMResponse` and `ChatDocument` remain unchanged. And we keep original LLM response - with "thought signatures" - in message history.

### DIFF
--- a/langroid/agent/chat_document.py
+++ b/langroid/agent/chat_document.py
@@ -120,7 +120,7 @@ class ChatDocument(Document):
 
     reasoning: str = ""  # reasoning produced by a reasoning LLM
     content_any: Any = None  # to hold arbitrary data returned by responders
-    content_with_reasoning: Optional[str] = None # original content including reasoning
+    content_with_reasoning: Optional[str] = None  # original content including reasoning
     files: List[FileAttachment] = []  # list of file attachments
     oai_tool_calls: Optional[List[OpenAIToolCall]] = None
     oai_tool_id2result: Optional[OrderedDict[str, str]] = None
@@ -415,8 +415,12 @@ class ChatDocument(Document):
         sender_role = Role.USER
         if isinstance(message, str):
             message = ChatDocument.from_str(message)
-        content = message.content_with_reasoning or \
-            message.content or to_string(message.content_any) or ""
+        content = (
+            message.content_with_reasoning
+            or message.content
+            or to_string(message.content_any)
+            or ""
+        )
         fun_call = message.function_call
         oai_tool_calls = message.oai_tool_calls
         if message.metadata.sender == Entity.USER and fun_call is not None:
--- a/langroid/agent/chat_document.py
+++ b/langroid/agent/chat_document.py
@@ -120,6 +120,7 @@ class ChatDocument(Document):
 
     reasoning: str = ""  # reasoning produced by a reasoning LLM
     content_any: Any = None  # to hold arbitrary data returned by responders
+    content_with_reasoning: Optional[str] = None # original content including reasoning
     files: List[FileAttachment] = []  # list of file attachments
     oai_tool_calls: Optional[List[OpenAIToolCall]] = None
     oai_tool_id2result: Optional[OrderedDict[str, str]] = None
@@ -316,6 +317,7 @@ class ChatDocument(Document):
         return ChatDocument(
             content=message,
             reasoning=response.reasoning,
+            content_with_reasoning=response.message_with_reasoning,
             content_any=message,
             oai_tool_calls=response.oai_tool_calls,
             function_call=response.function_call,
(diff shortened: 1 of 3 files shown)


pchalasani:
## PR Review

**Purpose**: Preserve inline reasoning/thought tags (e.g., `<thinking>...</thinking>`) in the LLM message history so models like Gemini 3 Flash and Amazon Nova don't lose their reasoning context across turns.

### Approach

The approach is sound and minimal:

1. Add `message_with_reasoning` to `LLMResponse` — stores the original LLM text before stripping thought tags
2. Add `content_with_reasoning` to `ChatDocument` — carries the value forward
3. In `to_LLMMessage`, prioritize `content_with_reasoning` when constructing history messages

This preserves backward compatibility: existing APIs remain unchanged, and the new fields are `None` unless reasoning was actually extracted from inline tags.

### Issues Found

**1. `message_with_reasoning` set unnecessarily for separate-reasoning-field models**

In both the streaming path (`openai_gpt.py:1535`) and non-streaming path (`openai_gpt.py:2218`), the condition is:

```python
message_with_reasoning=completion if reasoning else None,
```

This triggers whenever `reasoning` is non-empty, including models that provide `reasoning_content` as a **separate API field** (DeepSeek, o1, etc.). For those models, `content`/`completion` doesn't contain thought signatures — it's already the final answer. So `message_with_reasoning` gets set to the same value as `message`, which is wasteful and somewhat misleading.

Better condition structure:

```python
# streaming path
message_with_reasoning = None
if reasoning == "":
    reasoning, message = self.get_reasoning_final(completion)
    if reasoning:
        message_with_reasoning = completion
else:
    message = completion
```

This way `message_with_reasoning` is only set when inline tags were actually extracted.

**2. `content_with_reasoning` unconditionally wins in `to_LLMMessage`**

At `chat_document.py:418`:

```python
content = (
    message.content_with_reasoning
    or message.content
    or to_string(message.content_any)
    or ""
)
```

If any downstream code modifies `ChatDocument.content` after creation (e.g., agent processing, tool result appending), the modification is silently ignored in the message history because `content_with_reasoning` takes priority. Combined with issue #1, this affects **all** reasoning model responses, not just ones with embedded thought tags.

**3. Non-streaming path: `content` not updated on function-call parse failure**

At `openai_gpt.py:2184-2186`, when function call parsing fails:

```python
msg = msg_str + args_str  # msg overridden with content + args
```

But `content` (used for `message_with_reasoning`) is not updated to include the appended `args_str`. Edge case, but an inconsistency.

**4. Typo in PR title**: "Though signatures" should be "Thought signatures".

**5. No tests**: The PR has no test coverage for the new behavior. A test verifying that `message_with_reasoning` roundtrips through `LLMResponse -> ChatDocument -> LLMMessage` would be valuable.

### What's Good

- The core use case (inline `<thinking>` tags preserved in history) is correctly handled
- Minimal memory overhead — fields are `None` unless reasoning is extracted
- All existing APIs preserved, no breaking changes
- The streaming path cleanup (adding the missing `else: message = completion` branch) makes previously implicit variable reuse explicit

### Recommendation

The PR solves a real problem with a reasonable approach. I'd suggest addressing issues #1 and #2 before merging — together they create a subtle risk where `content` modifications are silently dropped for all reasoning model responses, not just the intended inline-tag case. Issue #1 is a one-line fix, and issue #2 could be mitigated by the same fix (if `content_with_reasoning` is only set for inline-tag models, the blast radius is much smaller). Adding at least a basic unit test would also strengthen confidence.

pchalasani:
@alexagr I am adding some changes on top of this, to address the above, in #962

PR #963, opened 2026-01-28
add update_message_history parameter to llm_response / llm_response_async

alexagr:
Add `update_message_history` parameter to `ChatAgent.llm_response / llm_response_async` methods to allow their use for scenarios where LLM response NOT NECESSARILY needs to be recorded in message history. 

For example, in my voice agent use-case I want to be able to generate "tentative LLM responses" while user is speaking (using STT "hypothesis" as user input). When final recognition arrives, I compare it to the last hypothesis - and if it's "close enough" use the already generated LLM response - thus reducing the end-to-end latency.

### DIFF
--- a/langroid/agent/chat_agent.py
+++ b/langroid/agent/chat_agent.py
@@ -1354,7 +1354,9 @@ class ChatAgent(Agent):
                 )
 
     def llm_response(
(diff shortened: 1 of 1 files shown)


alexagr:
please ignore this PR - I'll be better off calling LLM instance methods directly

PR #964, opened 2026-01-29
add reasoning parameter to show_llm_response / finish_llm_stream callbacks

alexagr:
Add `reasoning` parameter to `show_llm_response` / `finish_llm_stream` callbacks, so that they can properly display COT


pchalasani:
Thanks @alexagr , adding some mods on top of this in #965

PR #975, opened 2026-02-10
remove traceback from OpenAI API error logs

alexagr:
Current code logs traceback for every exception that happens in OpenAI SDK.
For API errors - e.g. Authentication or BadRequest errors - this is not very useful, as the error is coming from the OpenAI servers and has nothing to do with langroid/SDK code - so we are just flooding logs with unneeded tracebacks.
To overcome the problem I added special handling for `openai.APIError` in corresponding `OpenAIGPT` class methods.


pchalasani:
thanks @alexagr , covered in #979

PR #989, opened 2026-02-27
support creation of async httpx client by http_client_factory

alexagr:
**Problem description**

Current implementation of `http_client_factory` in `OpenAIGPT` supports only returning sync client - thus making it useless for scenarios where async APIs are used.

**Solution**

Support two shapes for `http_client_factory` return:
* single sync client - for backwards compatibility
* tuple of `(sync_client, async_client)` - for scenarios where async client is needed


pchalasani:
Thanks @alexagr . Covered in #990

PR #1012, opened 2026-04-08
fix: normalize Gemini model names with multi-suffix/date variants

nuthalapativarun:
## Summary

Fixes #995.

`_normalize_gemini_model_name` failed to normalize Gemini model variants whose name contains a date suffix appended *after* a keyword like `-exp` — for example `gemini-2.0-flash-thinking-exp-01-21`. The canonical name for this model is `gemini-2.0-flash-thinking-exp`, so stripping a keyword suffix alone (the previous logic) would give `gemini-2.0-flash-thinking`, which is not in `GEMINI_CANONICAL_MODEL_NAMES`.

### Root cause

The suffix-stripping loop tried each keyword (`-preview`, `-exp`, `-experimental`, `-latest`) and returned the stripped form only if it was a canonical name. It did not handle the case where a trailing `MM-DD` date component sits *after* the keyword.

### Fix

Two-pass normalization is applied after the initial direct-match check:

1. **Date-only pass** — strip a trailing `-MM-DD` component and check canonical. This handles variants of canonical names that already end with `-exp` (e.g. `gemini-2.0-flash-thinking-exp-01-21` → `gemini-2.0-flash-thinking-exp` ✓).
2. **Keyword + optional date pass** — existing keyword-stripping logic, extended to also try stripping a trailing date from the keyword-stripped result (for forward-compatibility with patterns where the date precedes the keyword suffix).

### Tests

`tests/main/test_model_info.py` (new file) covers:
- All canonical `GeminiModel` values normalize to themselves
- Provider-prefixed variants (`google/`, `vertex_ai/`)
- Simple keyword suffixes (`-preview`, `-latest`, `-experimental`)
- Dated preview variants (`gemini-2.5-flash-lite-preview-06-17` → `gemini-2.5-flash-lite`)
- Date-only suffix on canonical names containing `-exp` (`gemini-2.0-flash-thinking-exp-01-21` → `gemini-2.0-flash-thinking-exp`)
- Non-Gemini and completely unknown Gemini variants return `None`

## Test plan

- [x] `pytest tests/main/test_model_info.py` — 20 tests, all pass
- [x] `black` formatting applied

### DIFF
--- a/langroid/language_models/model_info.py
+++ b/langroid/language_models/model_info.py
@@ -819,19 +819,51 @@ def _normalize_model_names(models: List[str | ModelName]) -> List[str]:
 
 
 def _normalize_gemini_model_name(model: str) -> str | None:
+    """Normalize a Gemini model name to its canonical form.
+
+    Handles dated variants (e.g. ``gemini-2.0-flash-thinking-exp-01-21``),
+    preview/exp/experimental/latest suffixes, and combinations thereof.
+    """
+    import re
+
     base_model = model.rsplit("/", 1)[-1]
     if base_model in GEMINI_CANONICAL_MODEL_NAMES:
         return base_model
     if not base_model.startswith("gemini-"):
         return None
 
-    # Try stripping known suffixes to find a canonical name.
+    # Pattern for trailing date components like "-01-21" or "-06-17".
+    _DATE_RE = re.compile(r"-\d{2}-\d{2}$")
+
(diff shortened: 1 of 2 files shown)


pchalasani:
Thanks @nuthalapativarun — this fix was taken over in #1020, which implements the same date-suffix normalization with a superset of your test cases. Closing in favor of #1020; appreciate the contribution!

PR #1018, opened 2026-04-16
fix: avoid constructing extra httpx transports on client cache hits

### LINKED ISSUE #1017: client_cache: cache hit with http_client_config still constructs an unused httpx.Client
## Describe the bug

When `get_openai_client(..., http_client_config=...)` or
`get_async_openai_client(..., http_client_config=...)` hits the cache, Langroid
still constructs a fresh `httpx.Client` / `httpx.AsyncClient` before returning
the cached OpenAI client.

That extra transport is never attached to the returned client, so it becomes an
unused allocation on every cache hit for that config.

This seems to conflict with Langroid's stated goal that cached clients help
prevent resource exhaustion.

## To reproduce

A minimal regression test can reproduce this without network access by
monkeypatching `httpx.Client`:

```python
import httpx

from langroid.language_models.client_cache import _clear_cache, get_openai_client


def test_cache_hit_does_not_create_extra_httpx_client(monkeypatch):
    _clear_cache()
    created_clients = []
    real_client = httpx.Client

    class TrackingClient(real_client):
        def __init__(self, *args, **kwargs):
            super().__init__(*args, **kwargs)
            created_clients.append(self)

    monkeypatch.setattr(httpx, "Client", TrackingClient)

    client1 = get_openai_client(
        api_key="test-key",
        http_client_config={"timeout": 1.0},
    )
    client2 = get_openai_client(
        api_key="test-key",
        http_client_config={"timeout": 1.0},
    )

    assert client1 is client2
    assert len(created_clients) == 1
```

On current `main`, the final assertion fails because the second cache hit still
creates another `httpx.Client`.

The same pattern exists in `get_async_openai_client(...)` with
`httpx.AsyncClient`.

## Expected behavior

If the cache already contains a matching OpenAI client, Langroid should return
that cached client **without constructing a new transport first**.

## Relevant code paths

- `langroid/language_models/client_cache.py`
  - `get_openai_client()` creates `Client(**http_client_config)` before cache lookup
  - `get_async_openai_client()` creates `AsyncClient(**http_client_config)` before cache lookup

## Why this matters

For long-running processes that repeatedly instantiate clients with the same
`http_client_config`, each cache hit does extra work and can allocate an unused
transport object. That undermines the resource-saving goal of the cache.

## Suggested fix

Move the cache lookup ahead of `Client(**http_client_config)` /
`AsyncClient(**http_client_config)` construction, so transports are only created
on cache misses.

shaun0927:
Fixes #1017

## Summary

Avoid constructing extra `httpx` transports when `http_client_config` is used
with a cache hit in `client_cache.py`.

## Problem

`get_openai_client(..., http_client_config=...)` and
`get_async_openai_client(..., http_client_config=...)` were creating a fresh
`httpx.Client` / `httpx.AsyncClient` before checking the cache.

When the cache already contained a matching OpenAI client, Langroid returned the
cached client but the newly created transport was never used. That means every
cache hit for the same config could allocate an extra transport object.

This is at odds with the cache's goal of reducing resource churn.

## Fix

- compute the cache key first
- check the cache under the existing lock
- only construct `httpx.Client` / `httpx.AsyncClient` after confirming a cache miss

This keeps the existing cache and `prune_cache()` semantics unchanged while
removing the wasted transport allocation.

## Tests

Added regression coverage for both sync and async paths:
- `test_openai_client_cache_hit_does_not_create_extra_httpx_client`
- `test_async_openai_client_cache_hit_does_not_create_extra_httpx_client`

Local validation:
- `ruff check langroid/language_models/client_cache.py tests/main/test_openai_gpt_client_cache.py`
- `pytest -q tests/main/test_openai_gpt_client_cache.py -k 'not concurrent_async_achat' -q`

### DIFF
--- a/langroid/language_models/client_cache.py
+++ b/langroid/language_models/client_cache.py
@@ -111,19 +111,6 @@ def get_openai_client(
         _all_clients.add(client)
         return client
 
-    # If http_client_config is provided, create client from config and cache
-    created_http_client = None
-    if http_client_config is not None:
-        try:
-            from httpx import Client
-
-            created_http_client = Client(**http_client_config)
-        except ImportError:
-            raise ValueError(
-                "httpx is required to use http_client_config. "
-                "Install it with: pip install httpx"
-            )
-
     cache_key = _get_cache_key(
         "openai",
         api_key=api_key,
@@ -138,6 +125,18 @@ def get_openai_client(
         cached_client = _get_cached_client(cache_key)
         if cached_client is not None:
             return cast(OpenAI, cached_client)
+
+        created_http_client = None
+        if http_client_config is not None:
+            try:
+                from httpx import Client
+
+                created_http_client = Client(**http_client_config)
+            except ImportError:
+                raise ValueError(
+                    "httpx is required to use http_client_config. "
+                    "Install it with: pip install httpx"
+                )
 
         client = OpenAI(
             api_key=api_key,
@@ -192,19 +191,6 @@ def get_async_openai_client(
         _all_clients.add(client)
         return client
 
-    # If http_client_config is provided, create async client from config and cache
-    created_http_client = None
-    if http_client_config is not None:
-        try:
-            from httpx import AsyncClient
-
-            created_http_client = AsyncClient(**http_client_config)
-        except ImportError:
-            raise ValueError(
-                "httpx is required to use http_client_config. "
-                "Install it with: pip install httpx"
-            )
-
     cache_key = _get_cache_key(
         "async_openai",
         api_key=api_key,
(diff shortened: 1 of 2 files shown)


shaun0927:
I re-triggered the PR workflows by toggling the PR back through draft -> ready for review so both configured workflows are now attached to this PR.

At the moment, both runs are in `action_required`:
- Lint
- Pytest

I also ran the relevant checks locally before opening the PR:
- `ruff check langroid/language_models/client_cache.py tests/main/test_openai_gpt_client_cache.py`
- `pytest -q tests/main/test_openai_gpt_client_cache.py -k 'not concurrent_async_achat' -q`

If this is just the usual fork-PR workflow approval gate, I think the remaining step is a maintainer approval for Actions to run on this PR.

shaun0927:
@pchalasani I took another pass over the patch and there are still no review comments or code-level action items on my side.

The current blocker seems to be the fork-PR workflow approval gate (`Lint` and `Pytest` are both in `action_required`). I also re-ran the relevant checks locally and they passed:

- `ruff check langroid/language_models/client_cache.py tests/main/test_openai_gpt_client_cache.py`
- `pytest -q tests/main/test_openai_gpt_client_cache.py -k 'not concurrent_async_achat' -q`

If the fix direction looks good, I think the next step is maintainer approval for the PR workflows so the normal checks can run.

pchalasani:
thanks @shaun0927 I haven't had a chance to look at it. I am on holiday thru Monday, so will look after

pchalasani:
Thanks @shaun0927! This change was taken over and extended in https://github.com/langroid/langroid/pull/1048 (review fixes plus added tests/docs), which supersedes this PR - closing in its favor, with your original work credited there.

PR #1020, opened 2026-05-01
Fix Gemini model name normalization for date-suffixed exp variants

pchalasani:
## Summary
Fixes a bug in `_normalize_gemini_model_name()` where Gemini model names with both "-exp" suffix and a trailing date (e.g., "gemini-2.0-flash-thinking-exp-01-21") were not being normalized correctly to their canonical forms.

## Key Changes
- Added regex pattern `_GEMINI_DATE_SUFFIX` to match and strip trailing "-MM-DD" date stamps from Gemini model names
- Updated `_normalize_gemini_model_name()` to first attempt stripping the date suffix before checking for keyword suffixes like "-preview", "-exp", "-experimental", and "-latest"
- This ensures that canonical names already ending in "-exp" (like "gemini-2.0-flash-thinking-exp") can be correctly identified even when followed by a date stamp
- Added comprehensive test coverage in `test_model_info.py` with 20+ test cases covering:
  - Canonical names that round-trip correctly
  - Provider-prefixed names
  - Various keyword suffixes (preview, latest, experimental)
  - Keyword + date combinations
  - The specific bug case: date-only suffix on "-exp" names
  - Non-Gemini and unknown variant rejection
  - Validation that all canonical GeminiModel values normalize to themselves

## Implementation Details
The fix uses a two-stage normalization approach:
1. First strip any trailing "-MM-DD" date pattern
2. Then attempt to strip known keyword suffixes from the result

This ordering is critical because some canonical Gemini names already end with "-exp", and we need to handle the case where a date is appended to such names.

https://claude.ai/code/session_01NHqoho4CxDh94xd9XC1Tkx


pchalasani:
Validated and hardened: normalization no longer guesses for bare-dated unknown Gemini variants (they return None); merged latest main; model-info tests pass; make check clean.

PR #1021, opened 2026-05-05
fix memory leak in Task::init()

alexagr:
### Problem
`Task::init()` constructs a `ChatDocument` instance purely to feed `log_message(Entity.SYSTEM, ...)`.
This doc is registered in the class-level `ObjectRegistry` by `ChatDocument.__init__` but never removed.
In long-running services that spawn many tasks, this creates pretty massive memory leak.

### Solution
- explicitly remove document from `ObjectRegistry` by calling `ChatDocument.delete_id()`
- rename the variable to make it clear that it's temporary

### DIFF
--- a/langroid/agent/task.py
+++ b/langroid/agent/task.py
@@ -327,8 +327,8 @@ class Task:
         except Exception:
             logger.warning(
                 """
-                Failed to deep-copy Agent config during task creation, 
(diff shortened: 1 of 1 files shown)


pchalasani:
thanks @alexagr , added some mods in #1022 , closing this

Issue #1026, opened 2026-05-28
Remove dependency on pymupdf4llm/pymupdf


alexagr:
`langroid` library is licensed under MIT license.
However it has a **default** dependency on `pymupdf4llm/pymupdf` that are licensed under AGPL 3.0.
@pchalasani Is it possible to remove this **default** dependency, please? (and, for example, keep it only as part of `doc-chat`).

pchalasani:
@alexagr is there a non-AGPL licensed pdf parser that would be a good substitute? Not sure if you're actually using the pdf parser. I'm looking into pypdf

pchalasani:
Also pypdfium2 seems like a good candidate

PR #1027, opened 2026-05-28
remove pymupdf4llm from default dependencies

alexagr:
remove `pymupdf4llm` from default dependencies to mitigate #1026

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,6 @@ dependencies = [
     "pydantic-settings<3.0.0,>=2.0.0",
     "pygithub<2.0.0,>=1.58.1",
     "pygments<3.0.0,>=2.15.1",
-    "pymupdf4llm<0.1.0,>=0.0.17",
     "pyparsing<4.0.0,>=3.0.9",
     "pytest-rerunfailures<16.0,>=15.0",
     "python-dotenv>=1.0.0,<2.0.0",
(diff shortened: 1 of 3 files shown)


pchalasani:
Thanks for flagging the AGPL issue and putting up this fix, @alexagr! 🙏

We're going to supersede this with #1028, which builds on the same idea but also closes a gap: simply removing `pymupdf4llm` from the default deps would leave the **default** PDF parser (`pymupdf4llm`) uninstalled, so a bare `pip install langroid` would then break at PDF-parse time with an import error.

#1028 does the following:
- Removes `pymupdf4llm` (and its transitive AGPL `pymupdf`) from default deps — same as here. They remain available via the `doc-chat` / `all` / `pdf-parsers` / `pymupdf4llm` extras.
- Adds **`pypdfium2`** (Apache-2.0 / BSD-3-Clause — permissive, MIT-compatible) to core deps and makes it the **new default PDF parser**, so out-of-the-box PDF parsing keeps working with no AGPL code. Per the py-pdf benchmarks, `pypdfium2` matches/beats `pymupdf` on raw text-extraction accuracy; for richer Markdown-structured output you can still opt into `pymupdf4llm` via an extra.
- Adds a `PyPDFium2Parser` plus unit tests, and includes the same lazy-`import fitz` fix in `tests/main/test_llm.py` that you have here.

Closing this in favor of #1028 — but thank you again for the report and the PR, it's appreciated!

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1034, opened 2026-06-10
fix: restore agent-to-agent tool handoffs broken by GHSA-gjgq USER-origin filter

pchalasani:
## Problem

The `GHSA-gjgq-w2m6-wr5q` fix (released in **0.65.3**) added
`_filter_user_origin_tools`, which drops *handle-only* tools from
`USER`-origin messages — so a raw human input containing tool JSON can't
invoke a handler directly, bypassing the LLM.

But langroid `Task`s **relabel inter-agent messages' sender to `USER`**
(to the receiving sub-task, the caller's message looks like its "user"
input). So the new filter also dropped **legitimate agent-to-agent
handoffs** — one agent's LLM emits a tool, another agent handles it —
silently breaking the core multi-agent pattern.

**This is a regression on `main` since 0.65.3.** It affects
`recipient_tool` and the orchestration tools (`PassTool`, `ForwardTool`,
`SendTool`, `DoneTool`).

## Fix — distinguish agent/LLM-produced tools from raw human input

| Where | Change |
|---|---|
| `ChatDocMetaData` | New `tools_from_agent: bool`, set at construction (model_validator) whenever `sender ∈ {LLM, AGENT}`. **Only ever set, never cleared** — survives the USER relabeling, relays, and deepcopies. |
| `Task` result-wrapping | Preserves `tools_from_agent` when a sub-task result is relabeled to `USER` for the parent (its tools came from the sub-task's agent). |
| `_filter_user_origin_tools` | Returns tools unchanged when `tools_from_agent` is set; still drops handle-only tools from genuine raw `USER` input (flag stays `False`). |

A raw human input is **never** marked (it enters as `USER` and no agent
produced it), so the GHSA-gjgq protection is fully preserved.

## Verification

- **Regression (was failing on `main` since 0.65.3, now passes):**
  `tests/main/test_tool_orchestration.py` — **13 deterministic (MockLM)
  cases** across PassTool/ForwardTool/SendTool/DoneTool went 13-failed →
  21-passed. Also `test_recipient_tool.py` / `_async`.
- **Security still holds:** `tests/main/test_handle_message_security.py`
  **7/7** — raw USER-origin tool JSON is still filtered.
- `mypy -p langroid`, `black --check`, `ruff` all clean.

## Notes

`test_tool_orchestration.py::test_send_tools` shows occasional flakiness
in the *full* suite (a `ValidationError` from pre-existing test
inter-dependence — it passes 4/4 in isolation and is unrelated to the
filter); CI rerun-on-failure covers it.

### DIFF
--- a/langroid/agent/base.py
+++ b/langroid/agent/base.py
@@ -628,7 +628,7 @@ class Agent(ABC):
         recipient: str = "",
     ) -> ChatDocument:
         """Template for agent_response."""
-        return self.response_template(
+        doc = self.response_template(
             Entity.AGENT,
             content=content,
             files=files,
@@ -640,6 +640,15 @@ class Agent(ABC):
             function_call=function_call,
             recipient=recipient,
         )
+        if tool_messages:
+            # Tools explicitly produced as structured ToolMessages by this
+            # agent/handler (e.g. a handler returning AgentSendTool(tools=[...])
+            # or a tool object) are trusted like LLM output -- unlike content
+            # echoed from untrusted input, which carries no structured tools.
+            # Mark so they survive _filter_user_origin_tools after a Task
+            # relabels the sender to USER (GHSA-gjgq-w2m6-wr5q).
+            doc.metadata.tools_from_agent = True
+        return doc
 
     def render_agent_response(
         self,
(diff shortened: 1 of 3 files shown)


pchalasani:
@codex review

pchalasani:
@codex review

pchalasani:
@codex review

pchalasani:
@codex review

PR #1053, opened 2026-07-07
perf(mcp): build all MCP tools with a single list_tools() round-trip

alexagr:
FastMCPClient.get_tools_async() issued 1 + N tools/list round-trips to build N tools: one initial list_tools(), then one more inside get_tool_async -> get_mcp_tool_async for every tool, each re-fetching the entire tool list only to pick one entry by name. Because fastmcp's Client.list_tools() is uncached, every call is a real round-trip, so tool init was O(N) network calls -- the bulk of init latency on slow/remote servers with many tools.

Split the network fetch from the (pure, synchronous, network-free) schema-to-ToolMessage conversion:

- Add public FastMCPClient.tool_model_from_mcp_tool(target): builds a ToolMessage subclass from an already-fetched mcp.types.Tool with no round-trip. Exposing it lets subset/allow-list consumers convert a filtered list_tools() snapshot without re-listing once per tool.
- get_tool_async() fetches the target, then delegates to the builder (single-tool path keeps its unavoidable 1-list floor).
- get_tools_async() lists once and maps the builder over the result, taking it from 1 + N to 1 round-trip.

Behavior is preserved: same Tool objects, same generated models, no public signature changes. It also removes the latent inconsistency of re-listing a server whose tool list could change between the N calls.

Add tests that count real list_tools() round-trips: get_tools_async must issue exactly one regardless of tool count, and the public builder must add zero round-trips when building a filtered subset from one snapshot.

### DIFF
--- a/langroid/agent/tools/mcp/fastmcp_client.py
+++ b/langroid/agent/tools/mcp/fastmcp_client.py
@@ -321,6 +321,24 @@ class FastMCPClient:
         target = await self.get_mcp_tool_async(tool_name)
         if target is None:
             raise ValueError(f"No tool named {tool_name}")
+        return self.tool_model_from_mcp_tool(target)
+
+    def tool_model_from_mcp_tool(self, target: Tool) -> Type[ToolMessage]:
+        """
+        Build a Langroid ToolMessage subclass from an already-fetched MCP
+        `Tool` object.
+
+        This is a pure, synchronous, network-free conversion: it performs no
+        ``list_tools()`` round-trip. Pair it with a single ``list_tools()`` call
+        to build many tools without re-listing the server once per tool
+        (see :meth:`get_tools_async`).
+
+        Args:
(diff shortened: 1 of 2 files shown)


pchalasani:
Thanks a lot for this, @alexagr — nice catch on the `1 + N` `list_tools()`
round-trips, and a clean split of the network fetch from the pure,
network-free schema→`ToolMessage` conversion.

I've carried it forward in #1054, which **keeps your original commit** (and the
perf work intact) and layers on top:

- hardening of the now-public `tool_model_from_mcp_tool()` / `_schema_to_field()`
  against malformed or missing MCP `Tool` metadata (null/non-dict `inputSchema`,
  missing attributes, null nested `properties`/`items`, non-string `required`);
- a behavior-**equivalence** test (bulk `get_tools_async` vs per-tool
  `get_tool_async` model generation), alongside your round-trip counters;
- a short note in `docs/notes/mcp-tools.md`.

Closing this in favor of #1054 — full credit to you there. 🙏

PR #1057, opened 2026-07-11
fix: fail fast on non-retryable 4xx LLM errors

alexagr:
Native OpenAI-SDK errors have str(e) like 'Error code: 404 - ...', which does NOT contain 'NotFoundError', so the string-based guard in the retry decorators missed them and retried a permanent error with exponential backoff.

Add a type-based guard (before the generic retryable-errors handler) in both retry_with_exponential_backoff and its async variant, catching BadRequestError (400), AuthenticationError (401), PermissionDeniedError (403), NotFoundError (404) and UnprocessableEntityError (422). This also closes two latent gaps: the async path was missing 422 handling, and neither path handled 403. Retryable cases (429, 5xx, timeouts) are unaffected; the string guard is kept as a fallback for LiteLLM-proxy errors.

Add tests/main/test_retry_utils.py covering fail-fast for each 4xx type (sync + async) and a regression check that 429 still retries.

### DIFF
--- a/langroid/language_models/utils.py
+++ b/langroid/language_models/utils.py
@@ -43,17 +43,17 @@ def retry_with_exponential_backoff(
             try:
                 return func(*args, **kwargs)
 
-            except openai.BadRequestError as e:
-                # do not retry when the request itself is invalid,
-                # e.g. when context is too long
-                logger.error(f"OpenAI API request failed with error: {e}.")
-                raise e
-            except openai.AuthenticationError as e:
-                # do not retry when there's an auth error
-                logger.error(f"OpenAI API request failed with error: {e}.")
-                raise e
-
-            except openai.UnprocessableEntityError as e:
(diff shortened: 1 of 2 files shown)


pchalasani:
Thanks so much for this fix, @alexagr — great catch. Native OpenAI-SDK errors stringify to `Error code: 404 - ...` (no class name), so they slipped right past the substring-based guard and got retried with backoff; the 403/404 retry-storm against a retired model on an OpenAI-compatible endpoint (e.g. Gemini) is a real problem.

I've folded your commit (authorship preserved) into #1058, which adds a few regression tests on top — a proxy/LiteLLM-wrapped error whose message contains `NotFoundError` still fails fast via the retained substring fallback, and native 5xx still retries to exhaustion — plus a clarifying comment on the now-redundant `AuthenticationError` in the retry tuple.

Closing this in favor of #1058. Much appreciated! 🙏

PR #1061, opened 2026-07-22
fix: carry LLM "no content" faithfully as None (fixes Gemini 3.x tool calls)

alexagr:
### Overview

This PR fixes `langroid` behavior for new Gemini 3.5 Flash Lite and Gemini 3.6 Flash models on Vertex AI (Gemini Enterprise Agents Platform).

Before the change any tool call response triggered 400 error from the API - see [this post](https://discuss.ai.google.dev/t/tool-calling-doesnt-work-for-new-gemini-3-6-flash-and-gemini-3-5-flash-lite-models-on-vertex-ai-enterprise-agents-platform-via-openai-compatibility-api/175801/2) for details. The root cause was coercion of LLM response with missing content field into a LLM message with content == " ".

I tried to minimize the "blast radius" - and therefore kept `ChatDocument.content` as a mandatory string, and instead introduced a new `content_is_none` attribute that "carries the signal" along the LLMResponse > ChatDocument > LLMMessage parsing "pipeline".

I explicitly verified the change against both new Gemini models in my Vertex AI account.
Old code produced 400 error. New code produced a valid response.

### Detailed changes description

An assistant turn that is only a tool/function call has no text content. ChatDocument.to_LLMMessage padded such empty content with a single space (" ") "because some LLM APIs don't like an empty msg", so every stored tool-call turn carried content=" ". Gemini 3.x (e.g. gemini-3.5-flash-lite, gemini-3.6-flash) rejects an assistant turn that has BOTH a tool/function call AND non-empty text content with 400 INVALID_ARGUMENT, breaking every conversation on the turn that sends a tool result back to the model.

Carry the response shape faithfully instead of fabricating content:

- LLMResponse.message is now Optional[str]; a tool-only response keeps message=None (distinct from "") rather than being coerced to "".
- ChatDocument gains an explicit `content_is_none` flag (content itself stays a mandatory str). content_any cannot serve this role since it defaults to None on every ChatDocument.
- ChatDocument.to_LLMMessage reproduces "no content" as LLMMessage.content=None (now Optional[str]) instead of " ".
- LLMMessage.api_dict drops None content from the payload, and pads a lone space only when a message would otherwise be entirely empty (no content AND no tool/function call) - the case older Gemini rejects.

This distinguishes "missing" (None, omitted on the wire) from "present-but-empty" ("") end to end. None-safety guards added for the LLMMessage.content / LLMResponse.message consumers (token counting, truncation, display callbacks). Regression tests added.

### DIFF
--- a/langroid/agent/base.py
+++ b/langroid/agent/base.py
@@ -1186,7 +1186,7 @@ class Agent(ABC):
             # streaming was enabled, AND we did not find a cached response.
             # If we are here, it means the response has not yet been displayed.
             cached = f"[red]{self.indent}(cached)[/red]" if response.cached else ""
-            print(cached + "[green]" + escape(response.message))
+            print(cached + "[green]" + escape(response.message or ""))
         async with self.lock:
             self.update_token_usage(
                 response,
@@ -1261,7 +1261,7 @@ class Agent(ABC):
             # If we are here, it means the response has not yet been displayed.
             cached = "[red](cached)[/red]" if response.cached else ""
             console.print(f"[green]{self.indent}", end="")
-            print(cached + "[green]" + escape(response.message))
+            print(cached + "[green]" + escape(response.message or ""))
         self.update_token_usage(
             response,
             prompt,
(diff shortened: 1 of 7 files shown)


pchalasani:
Thanks @alexagr — great diagnosis of the Gemini 3.x tool-call 400s and a well-scoped fix. I've taken this over in #1062, which carries both of your commits (authorship preserved) rebased onto latest main, plus hardening from an adversarial review (making `content_is_none` authoritative on serialization round-trips, None-safety across consumers/providers, expanded regression tests, and a docs note). Closing this in favor of #1062.

PR #1064, opened 2026-07-24
fix: keep tool/function-result messages in history even when empty

### LINKED ISSUE #1063: History filter drops empty tool/function-result messages, leaving pending tool calls unanswered
### Summary

The message-history filter in `langroid/agent/chat_agent.py` (around line
1646–1651, `_prep_llm_messages` history filtering) drops tool/function-result
messages whose textual content is empty. As a result, when a pending
tool/function call's handler legitimately returns an empty string (`""`), the
`Role.TOOL` / `Role.FUNCTION` result message is removed from the prepared
history, the pending call is never answered, and `llm_response()` can return
`None` instead of sending the required tool response back to the model.

This is a **pre-existing** bug on `main`, NOT a regression from #1062 — see
adjudication below.

### Details

The filter predicate keeps a message only if it has non-whitespace text
content or carries a call:

```python
(m.content or "").strip() != "" or m.function_call is not None or bool(m.tool_calls)
```

A tool **result** message carries neither `function_call` nor `tool_calls`
(those live on the assistant turn that *made* the call), so an empty result
is filtered out. `done_tools` (the bookkeeping that clears pending calls) is
computed only **after** this filter, so the dropped result can never satisfy
its pending call.

Note this was equally broken before #1062: the old code padded empty result
content to `" "` (`content = content or " "`), but the old predicate was
`m.content.strip() != ""`, which drops the padded `" "` result too. Same
blast radius on both revisions.

First flagged by the codex review bot on #1062 (adjudicated pre-existing):
https://github.com/langroid/langroid/pull/1062#discussion_r3638079581 and
reply with evidence:
https://github.com/langroid/langroid/pull/1062#discussion_r3638271997

### Proposed fix

In the history filter, retain messages that answer a pending call regardless
of empty textual content:

- keep `Role.TOOL` messages (they carry `tool_call_id`), and
- keep `Role.FUNCTION` / function-result messages,

while preserving the existing filter behavior for other roles. Also verify
`LLMMessage.api_dict()` emits a valid `content` field for an empty tool
result (OpenAI requires `content` on tool messages; `""` or the existing
entirely-empty `" "` padding path is acceptable — after #1062 the padding
path already covers messages with no content and no calls).

### Regression test

A pending tool call whose handler returns `""` must:

- survive history preparation (result message retained),
- serialize as a tool-result message answering the call (valid `content`
  key present),
- clear the pending call so `llm_response()` proceeds normally.

Natural home: `tests/main/test_prep_llm_message.py`.

### Related (also pre-existing, noted on #1062)

Token accounting (`_message_num_tokens`) ignores serialized
`tool_calls`/`function_call` payload sizes, so a call with large arguments
counts as ~zero tokens in context-limit calculations. Could be fixed in the
same pass or separately.

AmirF194:
## Root cause

`_prep_llm_messages` (`langroid/agent/chat_agent.py:1646-1652`) filters the
messages built from an incoming turn to those with non-empty content, a
`function_call`, or `tool_calls`, then computes `done_tools` (the pending
tool-call ids being answered) from whatever survives that filter.

A tool/function-result message never carries `function_call` or
`tool_calls`: those live on the assistant turn that made the call, not on
the message answering it. So a legitimately-empty tool result (a tool that
succeeds with no output) fails all three conditions and is dropped before
`done_tools` is computed. Its `tool_call_id` never reaches `done_tools`, so
`self.oai_tool_calls` keeps treating that call as pending. When the result
was the only message being prepped, `llm_msgs` ends up empty and
`llm_response()` returns `None` instead of sending the tool response back
to the model.

## Fix

Keep a message when its role is `FUNCTION` or `TOOL`, alongside the
existing content/function_call/tool_calls checks, matching the same
"still required in history" reasoning the comment above it already states
for call-only assistant turns.

## Verification

Ran in a clean `python:3.11` Docker container this session, on this
branch (`uv sync`, then pytest). `python -c 'import
langroid.agent.chat_agent as m; print(m.__file__)'` resolved to the
in-repo copy, confirming provenance.

- Added `test_prep_keeps_legitimately_empty_tool_result`
  (`tests/main/test_prep_llm_message.py`): sets a single pending tool call,
  preps an empty-content tool-result `ChatDocument` answering it, and
  asserts the `TOOL` message survives in history and
  `agent.oai_tool_calls` clears. Fails on main (0 tool messages survive,
  `AssertionError: 0 == 1`), passes with the fix. Confirmed both ways by
  toggling the source change with the test left in place.
- Full `tests/main/test_prep_llm_message.py`: 22 passed.
- `tests/main/test_openai_gpt_none_content.py` and
  `tests/main/test_llm_response.py` (touch the same content/tool-call
  handling, run offline): 22 passed.
- Repo's PR-blocking Lint gate (confirmed from #1062's checks: `lint
  (3.11)`, `lint (3.13)`), run locally on the changed files: `black
  --check`, `ruff check`, `mypy -p langroid` all clean.

Not verified: the broader test suite gated behind `ready_for_review` (the
`Pytest` workflow) and any test requiring an OpenAI API key, since those
need live credentials this session does not have.

Fixes #1063

### DIFF
--- a/langroid/agent/chat_agent.py
+++ b/langroid/agent/chat_agent.py
@@ -1641,14 +1641,20 @@ class ChatAgent(Agent):
                 # either the message is a str, or it is a fresh ChatDocument
                 # different from the last message in the history
                 llm_msgs = ChatDocument.to_LLMMessage(message, self.oai_tool_calls)
-                # Keep messages that have text or a function/tool call. A call-only
-                # assistant turn has no content but is still required in history.
+                # Keep messages that have text, a function/tool call, or ARE a
+                # function/tool call result. A call-only assistant turn has no
+                # content but is still required in history, and so is a
+                # legitimately-empty tool/function result: neither carries
+                # function_call/tool_calls (those live on the assistant turn
+                # that made the call), so dropping it here would silently
+                # un-answer the pending call it responds to (issue #1063).
                 llm_msgs = [
                     m
                     for m in llm_msgs
                     if (m.content or "").strip() != ""
                     or m.function_call is not None
                     or bool(m.tool_calls)
+                    or m.role in (Role.FUNCTION, Role.TOOL)
                 ]
                 if len(llm_msgs) == 0:
                     return [], 0
--- a/tests/main/test_prep_llm_message.py
+++ b/tests/main/test_prep_llm_message.py
@@ -591,4 +591,27 @@ def test_content_is_none_overrides_stale_text_after_serialization():
 
     assert msg.content is None
     assert msg.tool_calls == [_tool_call()]
-    assert "content" not in msg.api_dict(MODEL)
+
+
+def test_prep_keeps_legitimately_empty_tool_result(agent: ChatAgent) -> None:
+    """A tool result with genuinely empty content (e.g. a tool that succeeds
+    with no output) must stay in history and clear the pending tool call from
+    agent.oai_tool_calls, not get silently dropped (issue #1063). Before the
+    fix, the history filter kept a message only if it had non-empty content, a
+    function_call, or tool_calls -- none of which a TOOL-role result carries,
+    so an empty-content result was dropped and its call_id never reached
+    `done_tools`, leaving the call marked pending forever."""
+    agent.oai_tool_calls = [_tool_call()]
+    doc = ChatDocument(
+        content="",
+        metadata=ChatDocMetaData(
+            sender=Entity.AGENT, source=Entity.AGENT, oai_tool_id="call_1"
+        ),
+    )
+
+    hist, _ = agent._prep_llm_messages(doc)
+
+    tool_msgs = [m for m in hist if m.role == Role.TOOL]
+    assert len(tool_msgs) == 1
+    assert tool_msgs[0].tool_call_id == "call_1"
+    assert agent.oai_tool_calls == []


pchalasani:
Thanks @AmirF194 — your root-cause analysis here was exactly right, and it was
a genuinely subtle one: the observation that a tool result carries neither
`function_call` nor `tool_calls` (because those live on the assistant turn that
made the call) is the whole bug, and the write-up made it easy to verify.

I've opened #1070, which builds directly on your fix and extends it:

- covers the legacy `Role.FUNCTION` path your patch described but didn't test
- requires the correlation identifier before keeping an empty result, so
  malformed results that `main` previously dropped aren't newly retained and
  sent to providers
- treats whitespace-only results as empty and canonicalizes them, so a large
  blank tool payload can't become un-truncatable
- restores `assert "content" not in msg.api_dict(MODEL)` in
  `test_content_is_none_overrides_stale_text_after_serialization`, which the
  diff here dropped — it guards the #1062 serialization behavior
- adds coverage for malformed identifiers, partial completion of multiple
  pending calls, truncation across roles, and an end-to-end `llm_response()`
  test for the symptom reported in #1063

Closing in favor of #1070. Thanks again for finding and diagnosing this.

PR #1067, opened 2026-07-25
fix: avoid constructing unused httpx client on cache hit

### LINKED ISSUE #1017: client_cache: cache hit with http_client_config still constructs an unused httpx.Client

Sanjays2402:
Closes #1017

When `http_client_config` is passed, the httpx transport was built *before* the cache lookup, so every cache hit allocated an `httpx.Client`/`AsyncClient` that was immediately discarded. Moving the construction inside the cache-miss branch fixes it for both `get_openai_client` and `get_async_openai_client`.

Regression tests count httpx client constructions across two identical calls: 2 before the fix, 1 after.

### DIFF
--- a/langroid/language_models/client_cache.py
+++ b/langroid/language_models/client_cache.py
@@ -111,19 +111,6 @@ def get_openai_client(
         _all_clients.add(client)
         return client
 
-    # If http_client_config is provided, create client from config and cache
-    created_http_client = None
-    if http_client_config is not None:
-        try:
-            from httpx import Client
-
-            created_http_client = Client(**http_client_config)
-        except ImportError:
-            raise ValueError(
-                "httpx is required to use http_client_config. "
-                "Install it with: pip install httpx"
-            )
-
     cache_key = _get_cache_key(
         "openai",
         api_key=api_key,
@@ -138,6 +125,20 @@ def get_openai_client(
         cached_client = _get_cached_client(cache_key)
         if cached_client is not None:
             return cast(OpenAI, cached_client)
+
+        # Only create the transport on a cache miss, so cache hits do not
+        # allocate an httpx client that is never used.
+        created_http_client = None
+        if http_client_config is not None:
+            try:
+                from httpx import Client
+
+                created_http_client = Client(**http_client_config)
+            except ImportError:
+                raise ValueError(
+                    "httpx is required to use http_client_config. "
+                    "Install it with: pip install httpx"
+                )
 
         client = OpenAI(
             api_key=api_key,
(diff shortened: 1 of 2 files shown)


pchalasani:
Thank you for this, @Sanjays2402 — your diagnosis is exactly right, and the fix
is the correct one.

It turns out this overlapped with #1048, which was already open and covers the
same change (delaying the `httpx.Client` / `AsyncClient` construction until
after the cache lookup, in both the sync and async getters) along with a fuller
set of regression tests. While reconciling the two I also picked up a related
issue in the same block: the `try` wrapped both the `from httpx import Client`
and the constructor call, so an `ImportError` from httpx's own constructor
(e.g. a `socks://` proxy without `socksio`) was misreported as "httpx is
required" — that's now fixed too, and #1048 credits both you and @shaun0927.

Closing this as superseded by #1048. Sorry for the duplicated effort — and
thanks for taking the time to write the regression tests rather than just the
one-line move; that's the right instinct. Contributions very welcome.

PR #1069, opened 2026-07-25
fix(batch): wait for first non-None result

### LINKED ISSUE #1068: `stop_on_first_result` cancels pending tasks after a `None` result
**Describe the bug**

`_process_batch_async(..., stop_on_first_result=True)` cancels all pending tasks after the first completed task set, even when every completed task maps to `None`.

The public batch API documents that it should continue until a non-`None` result is obtained. As a result, a fast invalid result can prevent a slower valid result from ever being returned.

**To reproduce**

```python
import asyncio

from langroid.agent.batch import ExceptionHandling, _process_batch_async


async def task(value: str, index: int) -> str | None:
    delays = {"invalid": 0.01, "valid": 0.2, "slow": 1.0}
    await asyncio.sleep(delays[value])
    return None if value == "invalid" else f"Processed {value}"


results = asyncio.run(
    _process_batch_async(
        ["invalid", "valid", "slow"],
        task,
        stop_on_first_result=True,
        handle_exceptions=ExceptionHandling.RAISE,
    )
)
print(results)
```

Current result:

```python
[None, None, None]
```

**Expected behavior**

The batch should keep waiting after the first `None`, return the first non-`None` result in its original position, and cancel only the tasks still pending at that point:

```python
[None, "Processed valid", None]
```

**Root cause**

The `finally` block that cancels `pending` tasks runs immediately after the first `asyncio.wait(..., FIRST_COMPLETED)` call. Waiting on the remaining tasks needs to continue until a mapped result is non-`None` or no tasks remain.

KXHXK:
## Summary

- keep waiting when the first completed batch task maps to `None`
- stop once a non-`None` result is available, preserving its original input position
- cancel only the tasks that are still pending at that point
- add a deterministic event-based regression test

## Root cause

The `stop_on_first_result` branch called `asyncio.wait(..., FIRST_COMPLETED)` only once. Its `finally` block then cancelled every pending task even when all completed tasks had produced `None`, contrary to the documented behavior of continuing until a valid result is obtained.

The branch now waits repeatedly on the remaining task set. Existing exception handling and cleanup stay in the same paths.

Fixes #1068

## Validation

- `pytest tests/main/test_batch.py -k "process_batch_async" -q` (`9 passed`)
- `ruff check langroid/agent/batch.py tests/main/test_batch.py`
- `git diff --check`

### DIFF
--- a/langroid/agent/batch.py
+++ b/langroid/agent/batch.py
@@ -118,21 +118,23 @@ async def _process_batch_async(
             task_indices = {task: i for i, task in enumerate(tasks)}
             results = [None] * len(tasks)
 
-            done, pending = await asyncio.wait(
-                tasks, return_when=asyncio.FIRST_COMPLETED
-            )
+            pending = set(tasks)
+            while pending:
+                done, pending = await asyncio.wait(
+                    pending, return_when=asyncio.FIRST_COMPLETED
+                )
 
-            # Process completed tasks
-            for task in done:
-                index = task_indices[task]
-                try:
-                    result = await task
-                    results[index] = output_map(result)
-                except BaseException as e:
-                    results[index] = handle_error(e)
+                # Process completed tasks
+                for task in done:
+                    index = task_indices[task]
+                    try:
+                        result = await task
+                        results[index] = output_map(result)
+                    except BaseException as e:
+                        results[index] = handle_error(e)
 
-            if any(r is not None for r in results):
-                return results
+                if any(r is not None for r in results):
+                    return results
         finally:
             for task in pending:
                 task.cancel()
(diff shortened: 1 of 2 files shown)


pchalasani:
Thanks a lot for this, @KXHXK — clean diagnosis in #1068 and a correct fix. I've taken it forward in #1072, which keeps your commit and authorship and adds some hardening on top: a deterministic version of your regression test (the valid task is now released from inside `output_map`, so it can't race into the first `asyncio.wait` round), an all-`None` termination test, a test for the exception path under `RETURN_NONE`, precise `stop_on_first_result` docstrings on the public batch helpers, and a new `docs/notes/batch-processing.md`.

Closing this one in favor of #1072 — the fix and the credit are yours. Much appreciated!

PR #1074, opened 2026-07-25
fix(sql): reject writes nested below a SELECT top-level node

pchalasani:
Fixes GHSA-3gpx-vwr3-xvwx (and the earlier, identical GHSA-wc83-4cvx-p8xc).

## The bug

`SQLChatAgent._validate_query` classified a statement by `isinstance` on the
**top-level** parse node only. Two shapes perform writes without a write node
at the top:

- a data-modifying CTE — `WITH x AS (DELETE FROM t RETURNING *) SELECT * FROM x`
- `SELECT ... INTO tbl`, which creates and populates a table

Both parse with a `Select` top node, so both were classified `SELECT`, passed
the default SELECT-only allowlist, and were executed and committed while the
database performed the write. Verified against sqlglot:

```
writable CTE   top=Select  nested_writes=['Delete']
SELECT INTO    top=Select  into=True
```

This defeats the specific guarantee the allowlist exists to provide — that the
default config is safe even when the agent's DB role can write. It needs only
ordinary write privileges, not superuser, so it is a strictly lower bar than
CVE-2026-25879.

## The fix

New `_nested_write_kinds()` walks the parsed statement for
`Insert`/`Update`/`Delete`/`Merge`/`Create`/`Drop`/`Alter`/`TruncateTable`
nodes below the top level, and separately flags the `SELECT ... INTO` form
(carried on the `Select`'s `into` arg, not as a nested `Create`). Any nested
kind outside `allowed_statement_types` is rejected with a message that names
what it found.

Deliberately **not** a denylist patch. Statement kinds are a closed set, so
walking the tree closes the whole class rather than naming one more construct —
which is why this is worth fixing where the function-denylist reports are not.

Behavior preserved:

- read-only CTEs still pass (the check keys on the embedded statement type, not
  on `WITH`)
- an operator who extends `allowed_statement_types` still gets exactly the
  writes they authorized, nested or not
- `allow_dangerous_operations=True` still bypasses everything

## Scope correction in SECURITY.md

I previously listed "statement shapes such as writable CTEs" as out of scope
and closed GHSA-wc83-4cvx-p8xc on that basis. That was wrong: it grouped a
bounded classification bug in the **allowlist** with the unbounded **denylist**
problem. The allowlist decides what a statement *does* over a closed set of
kinds, so a misclassification there is fixable and in scope. SECURITY.md now
says so explicitly, and the writable-CTE exclusion is removed.

## Tests

10 new cases in `tests/main/sql_chat/test_sql_chat_security.py`: three CTE
write forms, a doubly-nested CTE, `SELECT ... INTO`, read-only CTEs that must
still pass, the extended-allowlist path, and the `allow_dangerous_operations`
bypass.

**6 of them fail against unfixed source**; all 70 tests in the file pass with
the fix. black, ruff, and `mypy -p langroid` clean.

Two failures in `test_sql_chat_agent.py::test_sql_schema_tools` are pre-existing
LLM-dependent flakes — they fail with the fix stashed too, on a different
parametrization each run.

## Credit

Reported by [@manus-use](https://github.com/manus-use) (GHSA-3gpx-vwr3-xvwx),
and earlier by [@LHMisme420](https://github.com/LHMisme420)
(GHSA-wc83-4cvx-p8xc).

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -90,6 +90,12 @@ We aim to acknowledge in-scope reports within 7 days.
   precedence over this category.
 - **A code path that skips a documented safety gate entirely** — that is, a
   *missing call* to a validator, not a *bypass* of one.
+- **A statement the `allowed_statement_types` allowlist misclassifies.** The
+  allowlist is a different kind of control from the denylists above: it decides
+  what a statement *does* over a closed set of statement kinds, so a query that
+  performs a write the allowlist did not authorize — for example by nesting it
+  where the classifier does not look — is a bounded, fixable defect and is in
+  scope.
 - Vulnerable dependencies with a demonstrated impact on Langroid.
 
 ### Out of scope
@@ -98,8 +104,8 @@ The following will be closed without a fix, an advisory, or a CVE request:
 
 - New bypasses of the SQL, Cypher, or AQL denylists: previously unlisted
   functions, alternate quoting or escaping forms (backticks, Unicode escapes,
-  schema qualification), dialect-specific syntax, or statement shapes such as
-  writable CTEs. See "What is, and is not, a security boundary" above.
+  schema qualification), or dialect-specific syntax. See "What is, and is not,
+  a security boundary" above.
 - Sandbox escapes from the pandas `eval()` path when `full_eval=True`,
   including object-graph traversal via dunder attributes. `full_eval=True`
   disables the AST validator by design.
--- a/langroid/agent/special/sql/sql_chat_agent.py
+++ b/langroid/agent/special/sql/sql_chat_agent.py
@@ -231,6 +231,11 @@ def _nested_write_kinds(stmt: Any, kind_map: Dict[Any, str]) -> Set[str]:
         stmt: A parsed sqlglot statement.
         kind_map: Mapping of sqlglot expression type to statement-kind name.
 
+    Only independently executable nested statements count. A ``MERGE``'s
+    ``WHEN ... THEN UPDATE/INSERT/DELETE`` actions parse as child write nodes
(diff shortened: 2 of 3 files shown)


pchalasani:
Good catch — confirmed and fixed in cc0655a0.

sqlglot does parse a MERGE's actions as child write nodes:

```
MERGE INTO target t USING source s ON t.id = s.id
  WHEN MATCHED THEN UPDATE SET t.name = s.name
  WHEN NOT MATCHED THEN INSERT (id, name) VALUES (s.id, s.name)

top: Merge
  nested Update   ancestors=['When', 'Whens', 'Merge']
  nested Insert   ancestors=['When', 'Whens', 'Merge']
```

So with `allowed_statement_types=["MERGE"]` the walk returned
`{"UPDATE", "INSERT"}`, neither in the allowlist, and a perfectly valid MERGE
was rejected — forcing the operator to grant standalone UPDATE/INSERT, which is
broader than what they asked for.

Fixed by skipping write nodes that have a `When` ancestor. Keying on the `When`
wrapper rather than on "the top-level statement is a MERGE" preserves the
original intent: a MERGE nested inside a CTE is still reported, as `MERGE`.

```
top-level MERGE      top=Merge    nested=[]
MERGE inside CTE     top=Select   nested=['MERGE']
CTE delete           top=Select   nested=['DELETE']
SELECT INTO          top=Select   nested=['CREATE']
plain select         top=Select   nested=[]
```

Three regression tests added: a top-level MERGE under
`allowed_statement_types=["MERGE"]` passes, the same MERGE is still blocked
under the SELECT-only default, and a MERGE nested in a CTE is still caught. The
first fails without the exemption; all 73 tests in the file pass with it.

PR #1075, opened 2026-07-25
fix: route google/-prefixed Gemini models to the Gemini API

Sanjays2402:
Closes #995

`is_gemini_model()` matched only the `gemini/` prefix, so `google/gemini-3-flash-preview` was not detected as Gemini — `api_base` stayed unset and `GEMINI_API_KEY` was never applied — even though `get_model_info()` already normalizes `google/`-prefixed names to their canonical Gemini entry. That inconsistency is what the issue's follow-up comment reports.

Both prefixes now match via a shared `GEMINI_MODEL_PREFIXES` constant, and whichever prefix was used is stripped when deriving the provider-side model name.

Regression test `test_gemini_google_prefix` asserts both prefixes give `is_gemini`, the stripped model name, and `GEMINI_BASE_URL`; it fails on `main` (`assert llm.is_gemini` -> False) and passes with the fix.

### DIFF
--- a/langroid/language_models/openai_gpt.py
+++ b/langroid/language_models/openai_gpt.py
@@ -96,6 +96,8 @@ else:
 DEEPSEEK_BASE_URL = "https://api.deepseek.com/v1"
 OPENROUTER_BASE_URL = "https://openrouter.ai/api/v1"
 GEMINI_BASE_URL = "https://generativelanguage.googleapis.com/v1beta/openai"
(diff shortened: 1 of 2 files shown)


pchalasani:
Thanks @Sanjays2402! This change was taken over and extended in https://github.com/langroid/langroid/pull/1102 (review fixes plus added tests/docs), which supersedes this PR — closing in its favor, with your original work credited there.

PR #1076, opened 2026-07-26
fix(neo4j): match procedure namespaces through back-ticked identifiers

pchalasani:
Fixes GHSA-v5gj-pcf6-jjh7 (reported three times: also GHSA-h75w-m9jv-2c6v and
GHSA-8c3r-3hgh-c8hv).

## The bug

`_strip_literals_and_comments` blanks back-ticked identifiers so labels and
property keys named after keywords don't trip the write-clause checks. But that
also erases the text the *procedure* patterns need:

```
CALL `apoc`.load.json('u')   →   scrubs to   "CALL       .load.json(  )"
```

`\bapoc\.` matches nothing, so it was allowed — while the unescaped
`CALL apoc.load.json('u')` was correctly blocked. Neo4j resolves `` `apoc` ``
and `apoc` identically (openCypher `EscapedSymbolicName`), so the backtick was
a pure evasion of the `apoc.*`, `dbms.*` and `CALL db.*` checks. Verified before
the fix:

```
ALLOWED  CALL `apoc`.load.json('http://x/e')
ALLOWED  CALL `dbms`.security.changePassword('p')
ALLOWED  CALL `db`.schema.visualization()
BLOCKED  CALL apoc.load.json('http://x/e')
```

## The fix

`_unescape_backticked_names()` is a second normalization that keeps back-ticked
contents instead of blanking them, left-padded to the original span so offsets
are preserved and the name sits flush against the `.` that follows. The
dangerous-procedure patterns now match against both normalizations.

**The write-clause checks deliberately stay on the fully-scrubbed text.** A
back-ticked keyword is an identifier, not a clause — `` `CREATE` (n:X) `` is a
syntax error to Neo4j, not a sneaky write — so matching there would only cause
false positives on labels and properties legitimately named `CREATE`, `DELETE`,
`SET`. The reporter listed that as part of the bypass; it isn't one.

## Why this is in scope when denylist gaps are not

Not a missing name on a list — a **normalization** bug. The grammar defines two
spellings for one identifier and the check saw only one. That is finite and
completely fixable, unlike "add DuckDB `read_text` to the denylist", which has
no terminating fix. SECURITY.md now draws that line explicitly, with the test
being: **does the fix terminate?**

Same reasoning that put allowlist misclassification in scope in #1074. It also
removes a real asymmetry: the SQL side already received exactly this treatment
in CVE-2026-54760, where quoted identifiers evaded the regex and the fix was
AST-level normalization. The Cypher validator never got the equivalent, which
is why this keeps getting reported.

## Tests

10 new cases in `tests/main/test_neo4j_cypher_validation.py`: back-ticked
`apoc`/`dbms`/`db` namespaces on both read and write paths, partially and fully
back-ticked namespaces, offset preservation, the `allow_dangerous` bypass, and
six false-positive guards (back-ticked `CREATE`/`DELETE` labels, `` n.`set` ``
properties, dangerous names inside string literals and comments).

**8 fail with the fix reverted**; all 66 in the file pass with it. black, ruff,
`mypy -p langroid` clean; the AQL and CSV-KG validator suites are green too.

## Not included

`aql_validator.py` blanks backticks the same way, and its one dangerous pattern
is the UDF form `namespace::func`. I could not verify whether ArangoDB actually
accepts back-ticked UDF namespace segments, and applying the same normalization
blindly would risk false positives on attribute names containing `::`. Left
alone rather than fixed speculatively — worth a look if anyone has an ArangoDB
instance to confirm against.

## Credit

Reported by [@arpitjain099](https://github.com/arpitjain099), and earlier by
[@kaimandalic](https://github.com/kaimandalic) and
[@AI-Risk-Management](https://github.com/AI-Risk-Management).

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -96,16 +96,27 @@ We aim to acknowledge in-scope reports within 7 days.
   performs a write the allowlist did not authorize — for example by nesting it
   where the classifier does not look — is a bounded, fixable defect and is in
   scope.
+- **A *normalization* bug that lets a language-defined escaping mechanism
+  defeat a check.** Distinct from a gap in the denylist's contents: if the
+  grammar says two spellings mean the same thing (Cypher's back-ticked
+  identifiers, SQL's quoted or schema-qualified names) and our check sees only
+  one of them, that is a finite, completely fixable defect, and in scope.
 - Vulnerable dependencies with a demonstrated impact on Langroid.
+
+The distinction that decides scope, in one question: **does the fix
+terminate?** "Add another name to a list" does not. "This control fails to
+normalize a form the grammar defines, or misclassifies a closed set of
+statement kinds" does.
 
 ### Out of scope
 
 The following will be closed without a fix, an advisory, or a CVE request:
 
-- New bypasses of the SQL, Cypher, or AQL denylists: previously unlisted
-  functions, alternate quoting or escaping forms (backticks, Unicode escapes,
-  schema qualification), or dialect-specific syntax. See "What is, and is not,
-  a security boundary" above.
+- **Gaps in the *contents* of the SQL, Cypher, or AQL denylists**: a function,
+  procedure, or dialect-specific construct that is not on the list. There is no
+  terminating fix for this — the lists cover an unbounded grammar — so adding
+  one more name is not a security fix. See "What is, and is not, a security
+  boundary" above.
 - Sandbox escapes from the pandas `eval()` path when `full_eval=True`,
   including object-graph traversal via dunder attributes. `full_eval=True`
   disables the AST validator by design.
--- a/langroid/agent/special/neo4j/cypher_validator.py
+++ b/langroid/agent/special/neo4j/cypher_validator.py
@@ -182,8 +182,18 @@ def _unescape_backticked_names(query: str) -> str:
                 inner.append(query[j])
                 j += 1
             j = min(j + 1, n)
-            name = "".join(inner)
-            out.append(name.rjust(j - i))
+            # A back-ticked name that FOLLOWS a dot is a property/map key
+            # (``n.`apoc`.name``), never the head of a procedure namespace.
+            # Keep blanking those, exactly as before, or an ordinary map access
+            # named `apoc` or `dbms` would be rejected. A namespace head is not
+            # dot-preceded (`` `apoc`.load ``), and later segments of a
+            # namespace stay blanked without weakening the check, since the
+            # head plus its trailing dot is what the patterns match on.
+            prev = "".join(out).rstrip()
+            if prev.endswith("."):
+                out.append(" " * (j - i))
+            else:
+                out.append("".join(inner).rjust(j - i))
             i = j
         else:
             out.append(query[i])
--- a/langroid/agent/special/neo4j/cypher_validator.py
+++ b/langroid/agent/special/neo4j/cypher_validator.py
@@ -121,6 +121,76 @@ def _strip_literals_and_comments(query: str) -> str:
     return "".join(out)
 
 
+def _unescape_backticked_names(query: str) -> str:
+    """Like :func:`_strip_literals_and_comments`, but keep back-ticked *names*.
+
+    Blanking back-ticked identifiers stops false positives on labels and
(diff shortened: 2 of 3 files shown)


pchalasani:
Converting to draft. A four-lens cold-diff review found that this approach does not work, and I verified every claim by running both the base and branch validators side by side.

**It introduces false positives** — queries the base accepts and this branch rejects:

```
base=ALLOW head=BLOCK   WITH {x: 1} AS `apoc` RETURN `apoc`.x
base=ALLOW head=BLOCK   MATCH (n:`apoc.foo`) RETURN n
base=ALLOW head=BLOCK   RETURN `safe-apoc`.value
base=ALLOW head=BLOCK   MATCH (`apoc`:Person) RETURN `apoc`.name
```

**And it still does not close the hole** — whitespace or a comment between the namespace head and its dot walks straight through:

```
base=ALLOW head=ALLOW   CALL `apoc` .load.json('u')
base=ALLOW head=ALLOW   CALL `apoc`/*gap*/.load.json('u')
```

Plus: `\uXXXX` escapes inside back-ticked names are still unhandled, the `prev = "".join(out)` guard is quadratic (3.6s at 20k identifiers), and one of my own tests was vacuous — `CALL apoc.`load`.json(...)` has an unescaped head, so the base already blocked it.

**Why I am not going to iterate on it.** The blocker is `WITH ... AS `apoc` RETURN `apoc`.x`: distinguishing a back-ticked *variable* followed by property access from a back-ticked *namespace head* requires knowing what `apoc` was bound to. That is scope tracking, i.e. a parser. Cypher has no lightweight AST parser in our dependency set — the module docstring says exactly this — so every regex heuristic I add trades one false positive for another. Three rounds in, that is what has happened each time.

This is the test SECURITY.md now applies: **does the fix terminate?** I reclassified this from "denylist content gap" (out of scope) to "bounded normalization bug" (in scope) on the theory that backtick escaping is one finite rule. The evidence says I was wrong — position disambiguation, inter-token whitespace, and unicode escapes each need real grammar knowledge, and it does not terminate.

So the honest resolution is the policy one: the Cypher procedure denylist is best-effort hardening, not a security boundary. The boundary is the Neo4j role's privileges and the container. I would rather leave the guard imperfect and honest than ship a version that rejects valid user queries while still being bypassable.

Leaving this in draft. Closing GHSA-v5gj-pcf6-jjh7 as out of scope, with this reasoning on the advisory.

pchalasani:
Closing. Keeping this open as a draft would only suggest it's a work in progress, and it isn't — the approach is a dead end, not an unfinished patch.

Recap of why, so the reasoning survives with the branch:

- It **broke valid queries** that `main` accepts — `WITH {x: 1} AS \`apoc\` RETURN \`apoc\`.x`, ``MATCH (n:`apoc.foo`) RETURN n``, ``RETURN `safe-apoc`.value``, ``MATCH (`apoc`:Person) RETURN `apoc`.name``.
- It **still didn't close the hole** — ``CALL `apoc` .load.json('u')`` and ``CALL `apoc`/*gap*/.load.json('u')`` walk straight through, and `\uXXXX` escapes inside back-ticked names are unhandled.
- The blocker is structural: telling a back-ticked *variable* followed by property access apart from a back-ticked *namespace head* requires knowing what the name was bound to. That's scope tracking, i.e. a parser, and Cypher has no lightweight AST parser in our dependency set — which is why this validator is regex-based to begin with.

GHSA-v5gj-pcf6-jjh7 is closed as out of scope, with this evidence and a full explanation on the advisory. The Cypher procedure denylist stays what [SECURITY.md](https://github.com/langroid/langroid/blob/main/SECURITY.md) says it is: best-effort hardening, not a security boundary. The boundary is the privileges of the Neo4j role handed to the agent, plus the container — a least-privilege role without APOC or `dbms.*` closes this whole family regardless of what our regexes catch.

If anyone revisits this, the useful residue is here in the branch: `_unescape_backticked_names` plus 10 tests covering back-ticked namespace heads, offset preservation, and the false-positive guards. With a real tokenizer they'd become a reasonable starting point. Without one, the honest position is an imperfect guard rather than one that rejects users' valid queries while remaining bypassable.

Branch left in place at `fix/cypher-backtick-normalization`.

PR #1082, opened 2026-07-27
fix: correct _append_metadata_source logic and align docling version

fix2015:
## Summary

- Fixed logic bug in `_append_metadata_source()` in `doc_chat_agent.py` where duplicate sources were concatenated instead of deduplicated, and empty sources were not handled cleanly
- Aligned the `docling` version constraint in the `all` extra to match the stricter `doc-chat` requirement (`>=2.20.0`)

## Changes

**`langroid/agent/special/doc_chat_agent.py`** (lines 296-299)

The previous implementation had two issues:
1. When `orig_source == source` (both non-empty), the else branch concatenated them (`orig + source`), producing duplicated text like `"file.pdffile.pdf"`
2. When one source was empty, it concatenated both stripped values instead of returning only the non-empty one

New logic:
- Both non-empty and different → `"orig; source"` (semicolon-separated)
- Both non-empty and equal → return one copy
- One empty → return the non-empty one
- Both empty → return empty string

**`pyproject.toml`** (line 105)

Changed `"docling<3.0.0,>=2.16.0"` → `"docling<3.0.0,>=2.20.0"` in the `all` extra to match the `doc-chat` extra's requirement. Previously, `pip install langroid[all]` could install docling 2.16.x which is too old for doc-chat features.

## Test plan

- [x] Verified the fix handles all four source combination cases correctly
- [x] Verified docling version now matches between `doc-chat` and `all` extras


pchalasani:
Thanks @fix2015! This change was taken over and extended in https://github.com/langroid/langroid/pull/1097 (review fixes plus added tests/docs), which supersedes this PR — closing in its favor, with your original work credited there.

PR #1086, opened 2026-08-14
fix: stop sharing a mutable default config across no-arg instances

### LINKED ISSUE #1079: All vector-store providers share one mutable default config object across no-arg instances
## Summary

Every vector-store provider takes its config as a **mutable default argument**, which
Python evaluates once at import time. All of them then **mutate that config in place**.
So state set on one no-argument instance leaks into the next no-argument instance of
the same class, for the lifetime of the process.

All eight providers share the pattern:

```python
langroid/vector_store/chromadb.py:31      def __init__(self, config: ChromaDBConfig = ChromaDBConfig()):
langroid/vector_store/lancedb.py:56       def __init__(self, config: LanceDBConfig = LanceDBConfig()):
langroid/vector_store/meilisearch.py:40   def __init__(self, config: MeiliSearchConfig = MeiliSearchConfig()):
langroid/vector_store/milvusdb.py:41      def __init__(self, config: MilvusDBConfig = MilvusDBConfig()):
langroid/vector_store/pineconedb.py:67    def __init__(self, config: PineconeDBConfig = PineconeDBConfig()):
langroid/vector_store/postgres.py:64      def __init__(self, config: PostgresDBConfig = PostgresDBConfig()):
langroid/vector_store/qdrantdb.py:64      def __init__(self, config: QdrantDBConfig = QdrantDBConfig()):
langroid/vector_store/weaviatedb.py:42    def __init__(self, config: WeaviateDBConfig = WeaviateDBConfig()):
```

And all of them mutate it. `VectorStore.set_collection()` in `base.py:190-191` writes
`self.config.collection_name` and `self.config.replace_collection`; every provider's
`create_collection()` writes `self.config.collection_name` as well.

## Reproduction

```console
$ python -c "
import inspect
from langroid.vector_store.qdrantdb import QdrantDB
from langroid.vector_store.lancedb import LanceDB
for cls in (QdrantDB, LanceDB):
    d = inspect.signature(cls.__init__).parameters['config'].default
    print(cls.__name__, 'before:', d.collection_name, d.replace_collection)
    d.collection_name = 'foo'; d.replace_collection = True   # what set_collection() does
    d2 = inspect.signature(cls.__init__).parameters['config'].default
    print(cls.__name__, 'after: ', d2.collection_name, d2.replace_collection, 'same object:', d is d2)
"
QdrantDB before: temp False
QdrantDB after:  foo True same object: True
LanceDB before: temp False
LanceDB after:  foo True same object: True
```

## The damaging path

```python
a = QdrantDB()                          # uses the shared default config
a.set_collection("important", replace=True)   # mutates the shared default

b = QdrantDB()                          # no arguments, fresh object -- or so it looks
# b.config.collection_name == "important" and b.config.replace_collection is True,
# so b's constructor drops and recreates "important".
```

A caller who never mentioned `important` and passed no config destroys it. The window
is narrow (it needs no-argument construction) but the failure is silent data loss, and
the surprising part is that `b` looks entirely independent of `a`.

Provider `__init__`s pollute the shared object in quieter ways too: Milvus resolves
`uri`/`token`/`db_name` from the environment and upper-cases `metric_type` onto the
config, so the *first* instance permanently fixes those values for every later
no-argument instance in the process — including after the environment changes.

## Proposed fix

The standard remedy, applied uniformly across all eight providers:

```python
def __init__(self, config: QdrantDBConfig | None = None):
    config = config or QdrantDBConfig()
```

Worth pairing with a decision about whether `create_collection()` and
`set_collection()` should be mutating the caller's config at all — they are the reason
the aliasing is observable. That is a larger design question; making construction stop
sharing one object is the minimal fix and is independently correct.

Surfaced by the GitHub Codex reviewer on #1077, which fixes the Milvus instance
(its `set_collection()` override and env-resolving `__init__` make the aliasing most
reachable there). The remaining seven providers are unchanged.

Related: #1078 (the other latent `VectorStoreConfig` footgun).

Terminator666666:
Closes #1079

## Summary
Each vector-store provider used a mutable default argument for its config,
e.g. `def __init__(self, config: QdrantDBConfig = QdrantDBConfig())`.
Python evaluates that default once at import time, so every no-arg instance
received the same config object. Providers mutate that shared config in
place, so state set on one no-arg instance leaked into the next no-arg
instance of the same class.

## Change
Replace the shared mutable default with
`config: XxxDBConfig | None = None` plus `config = config or XxxDBConfig()`
in all 7 providers, so each no-arg instance builds a fresh config object.
Callers that pass an explicit config are unaffected.

Affected providers:
- `ChromaDB`, `LanceDB`, `MeiliSearch`, `PineconeDB`, `PostgresDB`,
  `QdrantDB`, `WeaviateDB`

## Tests
Added `tests/main/test_no_shared_default_config.py` with regression tests
asserting:
1. each provider's `__init__` default for `config` is `None` (not a shared
   mutable instance), and
2. two no-arg instances get distinct config objects.

14 tests pass locally; `black` and `ruff` are clean.

## Notes
A secondary design question raised in #1079 — whether
`set_collection`/`create_collection` should mutate the caller's config at
all — is left for a follow-up, as it is a larger API change with broad
call-site impact.

### DIFF
--- a/langroid/vector_store/chromadb.py
+++ b/langroid/vector_store/chromadb.py
@@ -28,7 +28,8 @@ class ChromaDBConfig(VectorStoreConfig):
 
 
 class ChromaDB(VectorStore):
-    def __init__(self, config: ChromaDBConfig = ChromaDBConfig()):
+    def __init__(self, config: ChromaDBConfig | None = None):
+        config = config or ChromaDBConfig()
         super().__init__(config)
         try:
             import chromadb
--- a/langroid/vector_store/lancedb.py
+++ b/langroid/vector_store/lancedb.py
@@ -53,7 +53,8 @@ class LanceDBConfig(VectorStoreConfig):
 
 
 class LanceDB(VectorStore):
-    def __init__(self, config: LanceDBConfig = LanceDBConfig()):
+    def __init__(self, config: LanceDBConfig | None = None):
+        config = config or LanceDBConfig()
         super().__init__(config)
         if not has_lancedb:
             raise LangroidImportError("lancedb", "lancedb")
--- a/langroid/vector_store/meilisearch.py
+++ b/langroid/vector_store/meilisearch.py
@@ -37,7 +37,8 @@ class MeiliSearchConfig(VectorStoreConfig):
 
 
 class MeiliSearch(VectorStore):
-    def __init__(self, config: MeiliSearchConfig = MeiliSearchConfig()):
+    def __init__(self, config: MeiliSearchConfig | None = None):
+        config = config or MeiliSearchConfig()
         super().__init__(config)
         try:
             import meilisearch_python_sdk as meilisearch
--- a/langroid/vector_store/pineconedb.py
+++ b/langroid/vector_store/pineconedb.py
@@ -64,7 +64,8 @@ class PineconeDBConfig(VectorStoreConfig):
 
 
 class PineconeDB(VectorStore):
-    def __init__(self, config: PineconeDBConfig = PineconeDBConfig()):
+    def __init__(self, config: PineconeDBConfig | None = None):
+        config = config or PineconeDBConfig()
         super().__init__(config)
         if not has_pinecone:
             raise LangroidImportError("pinecone", "pinecone")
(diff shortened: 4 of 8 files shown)


pchalasani:
Thanks @Terminator666666! This change was taken over and extended in https://github.com/langroid/langroid/pull/1099 (review fixes plus added tests/docs), which supersedes this PR — closing in its favor, with your original work credited there.

PR #1088, opened 2026-08-17
Serialize video attachments as `video_url` content-parts

octo-patch:
Reason: Video attachments were serialized as generic `file` content-parts, so chat-completions APIs that accept video input never received them as video.

## Changes

- `langroid/parsing/file_attachment.py`
  - `FileAttachment.to_dict()` now emits a `video_url` content-part for attachments whose MIME type starts with `video/`, mirroring the existing `image_url` part used for `image/*`.
  - As with images, a full `http`/`https` URL is passed through as-is, and otherwise the base64 data URI is used.
  - Extracted that shared "remote URL, else data URI" logic into a small `_content_url()` helper now used by both the image and video branches.
  - Images, PDFs and every other MIME type keep their current serialization: the new branch is only reached when the existing image branch does not match, so no existing payload changes.

- `tests/main/test_file_attachment.py`
  - New unit tests: video bytes produce a `video_url` part carrying the data URI, and a remote video URL is passed through unchanged.
  - Added regression tests pinning the existing `image_url` part for images and the generic `file` part for other files, which were previously only asserted to be non-`None`.

- `docs/notes/file-input.md`
  - Note that video files and URLs are sent as `video_url` content-parts.

## Checks

- `ruff check langroid/parsing/file_attachment.py tests/main/test_file_attachment.py` — all checks passed
- `ruff format --diff` on both changed Python files — already formatted
- Unit tests in `tests/main/test_file_attachment.py` — 13 passed

Model registry entries are out of scope here: the dispatch is keyed only off the attachment MIME type, so it stays model-agnostic and applies to any model whose API accepts video input.

### DIFF
--- a/docs/notes/file-input.md
+++ b/docs/notes/file-input.md
@@ -58,6 +58,9 @@ Currently the OpenAI-API supports:
 - PDF files (including image-based PDFs)
 - image files and URLs
 
+Video files and URLs are serialized as `video_url` content-parts, and work with
+any chat-completions API that accepts video input.
+
 
 ## Creating Attachments
 
--- a/langroid/parsing/file_attachment.py
+++ b/langroid/parsing/file_attachment.py
@@ -201,6 +201,21 @@ class FileAttachment(BaseModel):
         base64_content = self.to_base64()
         return f"data:{self.mime_type};base64,{base64_content}"
 
+    def _content_url(self) -> str:
+        """URL to send to the API for this attachment.
+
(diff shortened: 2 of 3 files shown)


pchalasani:
Thanks @octo-patch! This change was taken over and extended in https://github.com/langroid/langroid/pull/1101 (review fixes plus added tests/docs), which supersedes this PR — closing in its favor, with your original work credited there.

PR #1098, opened 2026-08-22
fix(security): add per-provider env_prefix to all vector-store configs (#1078)

### LINKED ISSUE #1078: VectorStoreConfig has no env_prefix, so bare HOST/PORT/FULL_EVAL env vars silently override every vector-store config
## Summary

`VectorStoreConfig` extends `pydantic_settings.BaseSettings` but never sets an
`env_prefix`. In pydantic-settings v2 that means **every field name is itself an
environment variable name**, matched case-insensitively. So a bare `HOST`, `PORT`,
`CLOUD`, `TIMEOUT`, `BATCH_SIZE`, `TYPE`, `STORAGE_PATH`, `COLLECTION_NAME` or
`FULL_EVAL` sitting in the process environment for any unrelated reason silently
overrides the default for **every** vector-store config.

## Reproduction

```console
$ HOST=evil.example.com PORT=9999 CLOUD=true FULL_EVAL=true \
  STORAGE_PATH=/tmp/pwned COLLECTION_NAME=hijacked \
  python -c "
from langroid.vector_store.qdrantdb import QdrantDBConfig
from langroid.vector_store.lancedb import LanceDBConfig
for C in (QdrantDBConfig, LanceDBConfig):
    c = C()
    print(C.__name__, c.host, c.port, c.cloud, c.full_eval, c.storage_path, c.collection_name)
"
QdrantDBConfig evil.example.com 9999 True True /tmp/pwned hijacked
LanceDBConfig  evil.example.com 9999 True True /tmp/pwned hijacked
```

`HOST` and `PORT` in particular are set routinely by shells, container runtimes,
process managers and CI systems.

## Why this matters most for `full_eval`

`full_eval` is the code-injection guard, documented as defaulting to `False` for
security and intended to be enabled only in trusted environments. A bare
`FULL_EVAL=true` anywhere in the environment disables sanitization for every vector
store, with no opt-in from the caller and nothing in the code or config to indicate it
happened.

## Why this is a vector_store-specific gap

The rest of Langroid already gets this right and does so consistently:

- `OPENAI_` and `AZURE_OPENAI_` (`embedding_models/models.py`,
  `language_models/openai_gpt.py`, `language_models/azure_openai.py`)
- `LITELLM_`, `LANGDB_`, `PORTKEY_`, `FORMAT_` (`language_models/`)
- `NEO4J_`, `ARANGO_` (`agent/special/`)
- `FIRECRAWL_`, `EXA_` (`parsing/url_loader.py`)

`langroid/vector_store/` is the one package that never adopted the pattern.
`MilvusDBConfig` (added in #1077) sets `env_prefix="MILVUS_"` because its fields are
named `uri` / `token` / `db_name` and the bleed was actively dangerous there — bare
`TOKEN` and `URI` were shadowing `MILVUS_TOKEN` and `MILVUS_URI`. That fix was
deliberately scoped to Milvus; this issue is the general case.

## Proposed fix

Give `VectorStoreConfig` a prefix (e.g. `VECDB_`) and each provider config its own
(`QDRANT_`, `LANCEDB_`, `CHROMA_`, `PINECONE_`, `WEAVIATE_`, `POSTGRES_`,
`MEILISEARCH_`), mirroring the pattern used everywhere else in the codebase.
Explicit constructor arguments already take priority over environment values in
pydantic-settings, so that precedence is unchanged.

## Compatibility note

This is a behavior change: anyone currently relying on bare `HOST` / `PORT` /
`COLLECTION_NAME` to configure a vector store would lose it. That behavior is
undocumented and looks accidental rather than intended, so the risk seems low — but it
warrants its own PR and a release note rather than riding along with a feature change.

Related: #1079 (the other latent `VectorStoreConfig` footgun).

pchalasani:
Closes #1078.

`VectorStoreConfig` extended `pydantic_settings.BaseSettings` with no `env_prefix`, so every field name was a bare, case-insensitive env var: an ambient `HOST`, `PORT`, `CLOUD`, `STORAGE_PATH`, `COLLECTION_NAME` — or worst, `FULL_EVAL=true`, which disables the code-injection guard — silently overrode the defaults of every vector-store config.

## Changes

- Base `VectorStoreConfig` now uses prefix `VECDB_`; each provider gets its own, mirroring the convention used everywhere else in the codebase: `QDRANT_`, `CHROMA_`, `LANCEDB_`, `PINECONE_`, `POSTGRES_`, `MEILISEARCH_`, `WEAVIATE_`.
- Kubernetes/Docker service-link guard: in a pod with a service named e.g. `qdrant`, the auto-injected `QDRANT_PORT=tcp://10.0.0.8:6333` would now feed the integer `port` field and crash config construction. A `mode="before"` validator on the base config detects the `tcp://` docker-link format, logs a warning, and falls back to the provider's default port. Any other invalid value still raises `ValidationError`.
- `docs/notes/vecstore-env-prefix.md`: new prefixes, migration note (this is a behavior change for anyone relying on the undocumented bare names — a repo-wide search found no such reliance), the service-link guard, and the inheritance caveat for third-party subclasses that don't set their own `env_prefix`.

## Verification

- `tests/main/test_vecstore_env_prefix.py` (36 tests, written TDD): bare vars no longer leak into any of the 8 provider configs; prefixed vars are honored; constructor kwargs beat env; no cross-provider prefix leakage; service-link `tcp://` values construct with default port + warning while other junk still fails validation. The suite clears ambient env first, so it passes even on machines with `POSTGRES_HOST` etc. set.
- The issue's exact repro now prints correct defaults.
- `make check` clean; `llms*.txt` bundles regenerated.

### DIFF
--- a/docs/notes/vecstore-env-prefix.md
+++ b/docs/notes/vecstore-env-prefix.md
@@ -1,0 +1,80 @@
+# Vector-Store Config Env-Var Prefixes
+
+Vector-store configs are `pydantic-settings` classes, which means their
+fields can be set via environment variables. Prior to this change (see
+issue #1078), `VectorStoreConfig` and its subclasses declared **no
+`env_prefix`**, so every field name was itself a case-insensitive
+environment variable. A bare `HOST`, `PORT`, `CLOUD`, `TIMEOUT`,
+`BATCH_SIZE`, `STORAGE_PATH`, `COLLECTION_NAME`, or `FULL_EVAL` in the
+environment silently overrode the default for *every* vector-store
+config:
+
+```bash
+HOST=evil.example.com PORT=9999 FULL_EVAL=true \
+  python -c "from langroid.vector_store.qdrantdb import QdrantDBConfig; \
+             c = QdrantDBConfig(); print(c.host, c.port, c.full_eval)"
+# old behavior: evil.example.com 9999 True
+```
+
+This was dangerous for two reasons:
+
+- `HOST` and `PORT` are routinely set by shells, containers, and CI
+  systems, so vector-store clients could silently point at the wrong
+  server.
+- `FULL_EVAL=true` disabled the code-injection guard (see
+  [code-injection-protection](code-injection-protection.md)) with no
+  explicit opt-in anywhere in the user's code.
+
+## New behavior
+
+Each vector-store config now has its own env prefix, consistent with
+the rest of the codebase (`OPENAI_`, `AZURE_OPENAI_`, `NEO4J_`, ...):
+
+| Config class        | Env prefix     |
+|---------------------|----------------|
+| `VectorStoreConfig` | `VECDB_`       |
+| `QdrantDBConfig`    | `QDRANT_`      |
+| `ChromaDBConfig`    | `CHROMA_`      |
+| `LanceDBConfig`     | `LANCEDB_`     |
+| `PineconeDBConfig`  | `PINECONE_`    |
+| `PostgresDBConfig`  | `POSTGRES_`    |
+| `MeiliSearchConfig` | `MEILISEARCH_` |
+| `WeaviateDBConfig`  | `WEAVIATE_`    |
+
+Bare env vars (`HOST`, `FULL_EVAL`, ...) are now ignored by all of
+these configs. To set a field via the environment, use the prefix of
+the specific config class, e.g. `QDRANT_HOST=my.qdrant.example` or
+`QDRANT_FULL_EVAL=true` for `QdrantDBConfig`. Every provider subclass
+in langroid overrides the prefix with its own, so within langroid
+`VECDB_*` vars affect only direct `VectorStoreConfig` instances.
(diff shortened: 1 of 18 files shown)


pchalasani:
Addressed the codex P2 in the follow-up commit: the `tcp://` service-link exception now lives in the environment settings source only (constructor / `model_validate` inputs fail validation normally), and only the exact service-link shape `tcp://<host>:<port>` is forgiven — malformed `tcp://` values now raise `ValidationError`. Also made `set_env` prefix-aware so it writes the env names the prefixed configs actually read, and tightened the regex anchor (`\Z`) so trailing-newline values fail loudly.

Known pre-existing limitation surfaced while testing (deferred, not introduced here): `set_env` serializes complex fields (e.g. `embedding`) as Python reprs that pydantic-settings cannot parse back, so full config roundtrips through `set_env` were already broken for such fields — orthogonal to this PR; worth its own issue if anyone relies on `set_env`.

PR #1099, opened 2026-08-22
fix: don't share mutable default configs across vector-store instances

### LINKED ISSUE #1079: All vector-store providers share one mutable default config object across no-arg instances

pchalasani:
## Summary
This takes over and extends #1086 by @Terminator666666. It prevents no-argument
vector-store instances from sharing mutable default configuration objects across
all seven concrete providers while preserving explicitly supplied configs.
Fixes #1079.

## Beyond #1086

- Fixed review findings around falsey explicit configs and explicit `None`.
- Extended the fix consistently across all seven concrete vector-store providers.
- Added constructor-signature, identity, explicit-config, and real Qdrant tests.
- Regenerated and committed all `llms*.txt` documentation artifacts.

## Testing
`uv run pytest -xvs tests/main/test_no_shared_default_config.py`: 22 passed.
`make check`: Black, Ruff, mypy, and generated documentation all clean.

Supersedes #1086 — thanks @Terminator666666 for the original contribution.


pchalasani:
Good catch — the llms*.txt bundles were regenerated before the pre-commit autoupdate revert, so they embedded ruff v0.16.4 while the committed config pins v0.16.0. Regenerated from the final tree in f14ab607; all six bundles now match .pre-commit-config.yaml. The same ordering issue affected several sibling PRs and has been fixed on those branches as well.

PR #1102, opened 2026-08-22
fix: route google/-prefixed Gemini models to the Gemini API

pchalasani:
## Summary

This takes over and extends #1075 by @Sanjays2402. It routes
`google/gemini-*` models to the Gemini API while retaining existing
`gemini/*` behavior. It also preserves provider-qualified model IDs for
custom and Vertex AI endpoints.

## Beyond #1075

- Fixed review findings around API-base provenance and Vertex model IDs.
- Kept explicit, copied, assigned, and provider-specific API bases intact.
- Added regression coverage for aliases, non-Gemini Google models, and Vertex.
- Regenerated and committed all `llms*.txt` documentation indexes.

## Testing

```text
uv run pytest -xvs \
  tests/main/test_llm.py::test_gemini_api_base \
  tests/main/test_llm.py::test_gemini_dynamic_config_preserves_api_base \
  tests/main/test_llm.py::test_gemini_google_prefix \
  tests/main/test_llm.py::test_gemini_alias_preserves_explicit_api_base \
  tests/main/test_llm.py::test_gemini_model_copy_preserves_explicit_api_base \
  tests/main/test_llm.py::test_gemini_assignment_preserves_explicit_api_base \
  tests/main/test_llm.py::test_non_gemini_google_model_is_not_rerouted
# 13 passed

make check
# clean
```

Supersedes #1075 — thanks @Sanjays2402 for the original contribution.


pchalasani:
Addressed in 77208af9 and 0ebdf624: api_base provenance is now value-based rather than keyword-based — a non-None api_base that does not come from the ambient environment default (including subclass class-level defaults) is preserved for Gemini-routed models, and ambient-env detection uses the same last-wins case normalization as pydantic-settings. Regression tests cover the subclass-default case and mixed-case env duplicates through the public constructor. Bundles regenerated from the final tree.

PR #1104, opened 2026-08-22
fix(security): generalize taint propagation across tool re-emission paths (#1035 Step A)

### LINKED ISSUE #1035: Harden GHSA-gjgq tool-origin filter against content laundering (taint propagation)
## Background

The `GHSA-gjgq-w2m6-wr5q` fix added `ChatAgent._filter_user_origin_tools`,
which drops *handle-only* tools from `USER`-origin messages so raw human
input cannot invoke a handler directly (bypassing the LLM).

That filter broke legitimate multi-agent tool handoffs (a Task relabels
inter-agent messages to `USER`), so PR #1034 restored them by marking
tools that were produced by an LLM or explicitly constructed by an
agent/handler (`ChatDocMetaData.tools_from_agent`); the filter passes those
through.

## Problem — the origin marking can be laundered

`tools_from_agent` is a *message-origin* signal, and langroid represents
tools as **JSON inside message content**. An agent that **forwards or
repackages untrusted USER content into structured `tool_messages`** can
launder user-supplied tool JSON into a "trusted" agent message, re-opening
the bypass. Known laundering paths:

1. **Echo** — a custom `agent_response` that returns `msg.content`
   (untrusted text containing tool JSON). *(Mitigated in #1034: a
   content-only response carries no structured `tool_messages`, so it is
   not marked.)*
2. **Parse-and-repackage** — `DonePassTool`/`AgentDoneTool` extract tools
   from a passed message via `agent.get_tool_messages(...)` (which parses
   tool JSON out of content, possibly from a raw `USER` message) and wrap
   them as `AGENT` `tool_messages`. `response_template` then marks them
   trusted, and `Task.result` preserves the flag through the `USER` relabel.
   **Not mitigated** — this is the developer-forwarding case #1034 leaves
   out of scope.

The general issue: once user-supplied tool JSON is parsed and repackaged
into an `AGENT` message's `tool_messages`, it is indistinguishable from an
LLM-emitted tool. Any per-message "is this agent-produced?" flag is leaky
against agents that repackage content.

## Proposed robust fix — taint propagation

Track *untrusted-origin* rather than *agent-origin*:

- Mark **external user input** as tainted at the entry point
  (`Task.run`/interactive input).
- **Propagate** the taint through derivations: `ChatDocument.deepcopy`,
  content echoes, `get_tool_messages` parsing, and
  `DonePassTool`/`AgentDoneTool`/`ForwardTool`/`PassTool` repackaging.
- `_filter_user_origin_tools` drops handle-only tools from **tainted**
  messages regardless of sender relabeling.
- LLM emissions and tools an agent constructs directly are never tainted,
  so legitimate handoffs keep working.

This makes the origin signal robust against the laundering paths above.

## Scope

PR #1034 ships the common-case fix (LLM emissions + agent/handler
explicitly-constructed structured tools) and treats "an agent forwarding
untrusted user content into a tool-handler" as the developer's
responsibility. This issue tracks the robust taint-propagation redesign.

Refs: PR #1034, GHSA-gjgq-w2m6-wr5q.

pchalasani:
Closes #1035 (Step A — Step B landed in #1038/#1065).

Generalizes untrusted-origin (taint) tracking so it survives every mechanical derivation path, closing the remaining content-laundering routes around the GHSA-gjgq tool-origin filter.

## Design

- `_tainted` moves from `AgentDoneTool` onto the `ToolMessage` base as a pydantic private attribute: invisible to all LLM-facing schemas, unspoofable (an LLM emitting a literal `"_tainted"` key lands in pydantic extra fields, never the private attr), and it survives `deepcopy`.
- Parse-time stamping: `Agent.get_tool_messages` marks every tool parsed out of (or attached to) a tainted document, so taint rides the tool object through any repackaging.
- Tool-level veto: `_filter_user_origin_tools` drops a `_tainted` handle-only tool regardless of which document carries it — laundering a stamped tool into a fresh trusted-looking message cannot wash it.
- Copy-on-stamp: the framework never mutates a tool object it didn't just create (a config-held reusable instance can't go sticky-tainted).

## Newly threaded paths

Handler results (sync + async, str/object/ToolMessage, including the recursive dispatch of handler-returned tools — a tainted handle-only returned tool is refused execution), `handle_message_fallback`-returned tools, `NonToolAction.DONE` repackage, `SendTool`/`AgentSendTool` re-emission, `TaskTool` sub-task seeding (carrier-doc taint OR the tool's own mark), and `Role.USER` history rehydration. The full audit (already-threaded / newly-threaded / not-applicable with reasons) is in `docs/notes/tool-origin-taint.md`.

## Explicit boundaries (documented, unchanged rulings)

- Taint cannot flow through an LLM generation — the trust boundary GHSA-gjgq's model explicitly trusts.
- A handler that constructs and returns its own `ChatDocument`, or hand-copies fields between tools, is trusted to label its output; taint tracking cannot follow arbitrary Python dataflow.
- Doc-chat corpus extracts and MCP results are separate external-taint *sources*, deliberately deferred to a follow-up.

## Verification

49 taint tests (test_tool_origin_taint 34 + test_tool_taint_laundering 15), each laundering test paired with a legitimate-flow control; fast orchestration/multi-agent batch 340 passed; live-LLM suites green; `make check` clean; `llms*.txt` regenerated.

### DIFF
--- a/docs/notes/tool-origin-taint.md
+++ b/docs/notes/tool-origin-taint.md
@@ -1,0 +1,126 @@
+# Tool-Origin Taint Tracking
+
+Langroid marks external user input as *tainted* and propagates that mark through
+every mechanical derivation of a message, so that untrusted content can never be
+"laundered" into a trusted-looking message that triggers handle-only tools.
+
+## Threat model
+
+`enable_message(MyTool, use=False, handle=True)` registers a tool that an LLM (this
+agent's or another agent's) may trigger, but that raw user input must not. The
+original fix (GHSA-gjgq-w2m6-wr5q) dropped handle-only tools from USER-origin
+messages, and PR #1034 added `ChatDocMetaData.tools_from_agent` so legitimate
+multi-agent handoffs (where a Task relabels an agent's output to `sender=USER`)
+keep working.
+
+Issue #1035 identified the remaining *laundering* hole: orchestration code that
+mechanically re-emits or repackages message content can make USER-derived data look
+like trusted AGENT/LLM output. Examples:
+
+- `DonePassTool` parses tool JSON out of the passed message and repackages it into
+  an `AgentDoneTool`
+- `RewindTool` / `RecipientTool` re-emit attacker-controlled `content` as a new
+  `sender=LLM` document
+- a tool handler echoes tainted content into its string result, which becomes an
+  AGENT-sender document
+- `handle_llm_no_tool=DONE` repackages `msg.content` and `msg.tool_messages` into
+  an `AgentDoneTool`
+
+## Mechanism
+
+Two marks, both set-only (nothing ever clears taint):
+
+- `ChatDocMetaData.tainted: bool` on every `ChatDocument`: True when the document
+  is external user input or was mechanically derived from a tainted document.
+- `ToolMessage._tainted: bool` (a private attribute on the `ToolMessage` base):
+  True when the tool *object* was parsed out of tainted content, or repackages such
+  a tool. Private attributes never appear in LLM-facing schemas or serialized
+  JSON, and survive `copy.deepcopy` (hence `ChatDocument.deepcopy`).
+
+Taint sources (documents born tainted):
+
+- `ChatDocument.from_str` (raw string input)
+- `Agent.to_ChatDocument` of a USER-authored string, `create_user_response`, and
+  the interactive-input path `_user_response_final` (USER sender; SYSTEM input is
+  operator-trusted and not tainted)
+- `Task.init` / `Task.run` USER-string entry points, and a pre-built USER
+  `ChatDocument` handed to a root task
+- `ChatDocument.from_LLMMessage` of a `Role.USER` history message
+
+Propagation (mechanical derivations that carry the mark):
+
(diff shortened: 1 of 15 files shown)


pchalasani:
Addressed the codex P2 in the follow-up commit: `_taint_tool_result` no longer force-taints a handler-returned `ChatDocument` — the handler's declared `metadata.tainted` label is honored, per the documented trust rule (docs/notes/tool-origin-taint.md). The mechanical cases keep their taint: plain str/object handler results still inherit the dispatched tool's taint, returned `ToolMessage`s still flow as stamped copies, and deepcopy-derived documents inherit automatically. Also fixed in the same commit: the strict-recovery `AnyTool` shim now propagates its `_tainted` mark onto the tool it reconstructs (a re-emission path that silently dropped taint), with sync+async tests.

Documented residual boundary (deliberate, per the issue's scope): a handler that constructs a fresh `ChatDocument` and copies tainted field values into it is trusted to label it — taint tracking cannot follow arbitrary Python dataflow inside handler code. The built-in MCP result handlers fall under the separately-deferred item of treating MCP/corpus results as external-taint sources in their own right (follow-up issue candidate).

PR #1105, opened 2026-08-22
feat: provider-neutral pre-execution tool policy hook (#1095)

### LINKED ISSUE #1095: feat: add a provider-neutral pre-execution tool policy hook
## Motivation

Langroid has a central tool-dispatch path in `Agent`: parsed `ToolMessage` types are mapped to handlers and wrapped before the handler is invoked. That is a useful framework boundary for authorization, approvals, DLP, command policy, and other security controls, but there is not currently a documented provider-neutral callback immediately before the selected tool handler executes.

I would like to contribute the smallest generic interception point rather than a product-specific dependency in core.

## Proposed shape

Add a sync/async pre-tool callback at the common handler-dispatch boundary, naming/API up to maintainers. It should receive the final tool identity / `ToolMessage` plus bounded agent/context metadata and be able to return a framework-native rejection without calling the underlying handler.

For a first slice, I would keep semantics narrow:

- no callback configured -> current behavior unchanged;
- allow/proceed -> invoke the selected handler exactly once;
- reject -> invoke the selected handler zero times;
- callback failure semantics are explicit rather than silently bypassing a configured policy boundary;
- no argument transformation unless maintainers want that in the contract.

## Acceptance contract

A focused implementation should prove:

1. sync and async handlers preserve existing behavior when no policy hook is configured;
2. allow executes the real handler exactly once;
3. deny executes the real handler zero times;
4. the hook sees the final parsed tool identity and structured payload before execution;
5. failures do not leak raw tool arguments in framework diagnostics;
6. existing `ToolMessage.handle` / agent handler dispatch remains backward compatible.

## Concrete consumer

I reached this seam while evaluating a local HOL Guard adapter for Langroid. HOL Guard would remain optional and externally maintained; it is only one consumer of the generic hook. The same primitive would support application authorization, approval services, DLP, and other policy engines.

If this direction fits Langroid's architecture, I can contribute the small generic hook + contract tests first and keep the HOL Guard adapter outside core unless you prefer an example/docs placement.

pchalasani:
Closes #1095.

Adds the smallest generic interception point at the central tool-dispatch boundary, so external policy engines (authorization, approvals, DLP, command policy) can allow or deny tool executions. No product-specific code in core.

## API

`AgentConfig.tool_policy: Optional[Callable[..., Any]] = None` (matches the existing callable-config idiom — `api_key_provider`, `streamer`). Called as `tool_policy(tool, agent)`, or with a `chat_doc=` kwarg when the callable declares it (mirrors the tool-handler convention). Sync or async callables both accepted.

Decision protocol: `True`/`None` = allow; `False` = reject (generic reason); `str` = reject with that reason; any other type = reject (fail-closed).

## Semantics

- No hook configured → behavior unchanged (negligible overhead: one None-check per dispatch).
- Allow → the selected handler runs exactly once.
- Reject → handler runs zero times; a framework-native rejection flows back to the LLM like other tool-handling errors.
- **Fail-closed:** a hook exception, or a tool payload that cannot be copied for evaluation, blocks the tool. The LLM-visible message names only the tool and exception class — never the tool arguments (those go to the operator log).
- The hook receives deep copies of the tool and chat_doc, so it cannot mutate what the handler executes.
- Enforcement lives ABOVE `handle_tool_message(_async)` (restored to their pristine bodies), in wrappers at every framework call site, so it composes with subclass overrides. Async policies are awaited on the caller's loop.
- The policy is shared by reference across `clone()` / `Task` / batch copies (a stateful budget policy stays one budget), via a memo-seeded deepcopy that never mutates the live config.
- The strict-recovery `AnyTool` parsing shim is structurally exempt (class-level marker, immune to LLM-injected instance fields); the recovered real tool is policy-checked exactly once, with the dispatching `chat_doc` threaded through.

## Verification

45 tests in `tests/main/test_tool_policy_hook.py` covering all six acceptance-contract points (sync + async), backward-compat across existing tool/task suites, and composition with the `_filter_user_origin_tools` security filter (vetoed tools never reach the policy). `make check` clean; docs note `docs/notes/tool-policy-hook.md`; runnable `examples/basic/tool-policy-hook.py`; `llms*.txt` regenerated.

## Follow-up (out of scope here)

A cold review noted that core dispatch resolves a tool's `_handler` from an instance attribute (`getattr(tool, "_handler", ...)`), which — given `ToolMessage`'s `extra="allow"` — an LLM could in principle inject. This is **pre-existing behavior in `main`'s `handle_tool_message`**, unrelated to this hook (the policy would in fact gate such a call), and the same instance-vs-class trust question as the `_tainted`/`_tool_policy_exempt` markers. It warrants its own hardening issue against core dispatch rather than riding this PR.

### DIFF
--- a/docs/notes/tool-policy-hook.md
+++ b/docs/notes/tool-policy-hook.md
@@ -1,0 +1,150 @@
+# Pre-execution Tool Policy Hook
+
+When Langroid agents are embedded in larger systems, operators often need a
+central place to decide whether a tool call the LLM just made should actually
+run: authorization checks, human-approval gates, DLP scanning, command
+policies, budget limits, and so on. The `AgentConfig.tool_policy` hook
+provides the smallest possible interception point for this: a single optional
+callback, consulted immediately before the selected tool handler executes.
+No policy engine, rules, or registries live in Langroid itself — you plug in
+whatever external engine you like.
+
+## API
+
+```python
+import langroid as lr
+
+class RunCommandTool(lr.ToolMessage):
+    request: str = "run_command"
+    purpose: str = "To run a shell <command>"
+    command: str
+
+def my_policy(tool: lr.ToolMessage, agent: lr.Agent) -> bool | str | None:
+    if isinstance(tool, RunCommandTool) and "rm" in tool.command:
+        return "destructive commands are not allowed"
+    return True  # allow
+
+config = lr.ChatAgentConfig(
+    # ... other params ...
+    tool_policy=my_policy,
+)
+```
+
+The callback is invoked as `tool_policy(tool, agent)` — or as
+`tool_policy(tool, agent, chat_doc=...)` if the callable's signature has a
+`chat_doc` parameter (mirroring the convention for tool handlers). It
+receives:
+
+- `tool`: the final parsed `ToolMessage` instance about to be handled — its
+  `request` field is the tool name, and its other fields are the fully
+  validated arguments
+- `agent`: the agent that is about to run the handler
+- `chat_doc` (optional): the `ChatDocument` containing the tool call, when
+  available
+
+The return value decides what happens:
+
+- `True` or `None` — allow: the selected handler runs exactly once, as usual
+- `False` — reject: the handler is NOT run; the LLM receives a rejection
+  message naming the tool
+- a `str` — reject, with that string included as the reason in the
+  LLM-visible rejection message
+
+The hook cannot modify the tool: argument transformation is deliberately out
+of scope, and the hook receives deep copies of both the `ToolMessage` and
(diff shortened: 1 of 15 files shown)


pchalasani:
Closing this in favor of the approach recommended in #1095: wrapping the innermost dispatch chokepoint (`Agent.handle_tool_message` / `handle_tool_message_async`) externally, rather than a built-in config hook. A core hook has to compose with every tool-dispatch path (sync/async, subclass overrides, AnyTool strict-recovery, handler-returned-tool recursion, clone/Task/batch, and parallel OpenAI tool-call accounting) or it silently leaves a gap — a poor property for a security control. Wrapping the innermost method covers all paths at once with no changes to Langroid. See the detailed writeup in #1095.

PR #1119, opened 2026-08-30
fix(parsing): tolerate and log malformed segment specs in parse_number_range_list

pchalasani:
Supersedes #1117 — thanks @pacocartones for finding this and for the original
fix, which this PR builds on.

### Problem

`parse_number_range_list` (`langroid/parsing/utils.py`) is documented to
tolerate weak-LLM output, but it raised `ValueError` on common messy input:

- a trailing comma: `"1,2,3,"` → `int("")`
- an empty entry: `"1,,2"`
- a lone hyphen or a malformed range: `"-"`, `"1-"`, `"-5"`, `"1-2-3"`

Its only production caller, `extract_numbered_segments` via
`RelevanceExtractorAgent.extract_segments`, wraps the call in
`except Exception: return DONE + NO_ANSWER`, so the crash was swallowed and
**every** segment the model had identified was silently discarded in the
DocChat RAG relevance path.

### Fix

Skip digit-less fragments and malformed ranges instead of raising, so the
well-formed part of the spec is still honored — and log the skipped fragments
at `WARNING`, so the partial result is not silent. (The previous behavior
dropped everything with no log at all.) A malformed range such as `"1-2-3"` is
deliberately **not** guessed at.

Well-formed input is unchanged: a differential run over 10,129 generated
well-formed specs found 0 differences from `main`.

### Tests

- 13 parse cases: every well-formed case preserved, every messy case tolerated.
- Warning-emitted / not-emitted cases.

The log assertions use a plain `logging.Handler` rather than `caplog`, since
CI runs `tests/main` with `-p no:logging` (see #1115).

Verified RED against `main` (13 failures) and GREEN on this branch (18 passed),
each run with an in-process source assertion so the import could not silently
resolve to the wrong tree. `black --check`, `ruff check`, and `mypy` on the
changed source are clean.

### DIFF
--- a/langroid/parsing/utils.py
+++ b/langroid/parsing/utils.py
@@ -216,6 +216,11 @@ def parse_number_range_list(specs: str) -> List[int]:
         specs (str): A string containing segment numbers and/or ranges
                      (e.g., "3,5,7-10").
 
+    Weak LLMs sometimes emit messy specs (a trailing comma, an empty entry, a
+    lone hyphen, or a malformed range like "1-" or "1-2-3"). Such fragments are
+    skipped -- and logged at WARNING -- rather than raised on, so the
+    well-formed part of the spec is still honored.
+
     Returns:
         List[int]: List of segment numbers.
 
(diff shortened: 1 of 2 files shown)


pchalasani:
Thanks — the P2 is valid and is fixed in 25825ed2.

You're right that the docstring promised more than the code delivered: an empty fragment (trailing comma, or `1,,2` — the two motivating cases) was skipped without being recorded.

I sharpened the rule rather than adding noise, since a trailing comma loses nothing:

- a fragment with **no digits at all** (``, `-`) refers to no segment → skipped quietly;
- a fragment that **does contain digits** but isn't a valid spec (`1-`, `-5`, `1-2-3`) drops a segment reference the model meant to make → skipped **and** logged at WARNING.

The docstring now states exactly this, and the tests cover both halves (`1,-,2` no longer warns; `1-2-3,5` still does). An exhaustive differential over 5,482,928 well-formed spec combinations shows 0 differences from main.

@codex review

pchalasani:
Valid again — fixed in c074549f.

`--` and `-foo-` both filter down to `--`, which is neither `""` nor `"-"`, so they fell through to the range branch and got logged as a dropped segment despite carrying no digits. The check now tests for the **absence of digits** rather than the two literal values, which is what the documented rule actually says. Added `1,--,2`, `1,-foo-,2` and `--` to both the parse cases and the no-warning cases; counter-verified RED against the previous check.

@codex review

PR #1120, opened 2026-08-30
fix(utils): expand '~' in read_file and close the resulting file-tool sandbox escape

pchalasani:
Supersedes #1118 — thanks @pacocartones for finding this and for the original
fix, which this PR builds on.

### Problem

`read_file` (`langroid/utils/system.py`) supports `~` home paths via
`expanduser()`, but checked existence on the **un-expanded** path first:

```python
if not Path(path).exists():
    raise FileNotFoundError(f"File not found: {path}")
file = Path(path).expanduser()
```

So `read_file("~/existing_file")` raised `FileNotFoundError` even when
`$HOME/existing_file` exists. Reachable through `ReadFileTool`.

### Why the one-line fix alone is not safe

`ReadFileTool` gates on `safe_resolve_path(dir, self.file_path)`, which
resolved `"~/secret.txt"` as a **literal `~` directory** under `curr_dir` — so
the guard passes, and an expanding `read_file` then reads the real home
directory. Expanding only in `read_file` turns a blocked read into a working
sandbox escape (`~/.ssh/id_rsa` and friends). Reproduced against a fake
`$HOME`:

```
----- before this PR, with only read_file expanding -----
safe_resolve_path ALLOWED -> /tmp/escape/base/~/secret.txt
read_file result: 'SECRET-HOME-FILE\n'
```

### Fix

1. `read_file` expands once, then checks and reads the expanded path.
2. `safe_resolve_path` expands `~` too, so a `~/...` path is checked as the
   home path it actually refers to and is rejected as outside the allowed
   directory.
3. `Path.expanduser()` raises `RuntimeError` for an unexpandable reference such
   as `~nosuchuser/f.txt`, and the file tools catch only `ValueError` — so
   agent-supplied input could crash a handler, and `read_file` would raise
   `RuntimeError` where it documents `FileNotFoundError`. The new
   `expand_user_path()` falls back to the literal path; such a path then stays
   inside the sandbox and is simply reported as missing.

### Tests

- Tilde read succeeds; a missing tilde path still raises `FileNotFoundError`.
- `safe_resolve_path` rejects `~/...` and still allows in-base paths.
- `ReadFileTool` tilde escape is blocked — a companion to the existing
  GHSA-fg23-3346-88f5 traversal tests.
- Unexpandable `~user` paths do not raise `RuntimeError` out of `read_file`,
  `safe_resolve_path`, or the `ReadFileTool` handler.

Each new test was counter-verified RED with its corresponding source fix
reverted, and all 33 tests in `test_system_utils.py` + `test_file_tools.py`
pass on this branch. `black --check`, `ruff check`, and `mypy` on the changed
source are clean.

### DIFF
--- a/langroid/utils/system.py
+++ b/langroid/utils/system.py
@@ -192,9 +192,10 @@ def safe_resolve_path(base_dir: str | Path, user_path: str | Path) -> Path:
     Resolve ``user_path`` relative to ``base_dir`` and ensure the result stays
     within ``base_dir`` (a path-traversal guard for file tools).
 
-    A ``user_path`` containing ``..`` segments, an absolute path, or a symlink
-    pointing outside ``base_dir`` is rejected. Symlinks are resolved via
-    :meth:`pathlib.Path.resolve`, so symlink-based escapes are caught as well.
+    A ``user_path`` containing ``..`` segments, an absolute path, a ``~`` home
+    reference, or a symlink pointing outside ``base_dir`` is rejected. Symlinks
+    are resolved via :meth:`pathlib.Path.resolve`, so symlink-based escapes are
+    caught as well.
 
     Args:
         base_dir (str|Path): Directory that file operations must stay within.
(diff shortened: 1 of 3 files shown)


pchalasani:
Thanks — the P1 is valid and is fixed in 8b0ac871.

You're right that the guard and the operation had diverged: `WriteFileTool.handle` and `ListDirTool.handle` validate with `safe_resolve_path` but then operate on the raw path, and `create_file`/`list_dir` don't expand `~`. So with `curr_dir == $HOME`, `~/x` validated as the in-base `$HOME/x` while the literal `$HOME/~/x` could follow a `~` symlink out of the sandbox.

Rather than changing the three handlers, `safe_resolve_path` now checks **both** readings — expanded and literal — and rejects if either escapes. That's strictly stricter and holds regardless of whether a given caller expands. Added a regression test for exactly your scenario ($HOME as curr_dir, `~` symlinked outside), counter-verified RED without the fix.

@codex review

PR #1145, opened 2026-09-19
fix: preserve document identity in fuzzy search results

dingpuyu:
Fuzzy search can return the wrong document when one cleaned chunk contains another. For example, searching for `quantum entanglement` in `["quantum mechanics", "quantum"]` ranks the second chunk first, but the substring lookup maps both matches to the first document. Reciprocal rank fusion then loses the second document when collecting results by ID. Identical cleaned chunks from different sources have the same problem.

Keep the document index returned by `process.extract` and use it to retrieve the original document, so each score stays attached to the correct content and metadata.

Regression tests cover overlapping and identical cleaned text, top-k selection, and the Agent retrieval path. Four new cases fail before the fix; all 54 tests in the two affected test modules pass afterward. `make check` also passes, including Black, Ruff, and mypy. The provider integration suite was not run.


dingpuyu:
Thanks for the `sweep-approved` label. The Lint workflow is still waiting for approval; could a maintainer please allow it to run when convenient? The string-search and retrieval-threshold test suites pass locally (54 tests), including the overlapping and identical-text regression cases. Happy to make any adjustments needed.

pchalasani:
@codex review

PR #1146, opened 2026-09-20
fix: merge complete groups of overlapping context windows

YaoxinHuang:
When a later window connects two existing groups, `components` only relabels its immediate neighbors. This leaves already-connected windows separate, so `VectorStore.remove_overlaps` can return duplicate chunks. Relabel every member of the joined groups and add graph and context-window regressions.

Verified the regression fails before the fix and passes after it; Black, Ruff, mypy, and the pinned pre-commit hook pass. The tests use no model or database service; the full integration suite was not run.


pchalasani:
@codex review

